PR #3, opened 2016-03-02
Add a schema for Fennec "core" pings.

Dexterp37:
The "meta" section is not being validated in this schema. Should the meta bits be added server-side?


mreid-moz:
This looks good.  We don't need the "meta" part (at least, not yet), since it is added by the server and not included in submissions from clients.

Thanks!

Issue #5, opened 2016-03-03
Setup Travis CI to automatically validate sample pings against the schema


mreid-moz:
Could use something like [is-my-json-valid](https://github.com/mafintosh/is-my-json-valid) for simplicity.

trink:
we should use a rapidjson tester (since the patterns are slightly different)

mreid-moz:
Yes, I'm in favour of using the exact same code we'll use to validate on ingestion.

gregglind:
+10000  on this.    what's blocking?

trink:
Integration tests for both the JSON and Parquet schemas were added here https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/55

Natim:
I looked into it, right now the readme is only for CentOS while Travis is running on Ubuntu.
I am unable to run the tests on Ubuntu so far because I need to find out what are the packages for the following:

- luasandbox → lua-sandbox-extensions?
- hindsight → hindsight
- luasandbox-lfs →
- luasandbox-lpeg →
- luasandbox-rjson → 
- luasandbox-cjson →
- luasandbox-parquet

The other solution would be to build a Dockerfile able to run the tests and [run the docker image on Travis](https://twindb.com/building-rpm-on-travis-ci-in-docker-containers/)

Natim:
I just realized that most of this missing package are actually Mozilla projects, should I be building the DEB packages from source then?

trink:
You could but the packages from each sprint are also available here: https://hsadmin.trink.com/packages/

acmiyaguchi:
I'm interesting in setting up this service. I have a few ideas that should make the continuous integration process a useful tool in both development and review of schemas. 

I've filed a bug for tracking [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1450289).

acmiyaguchi:
Issue #165 has a schema from a batch comparison tool. TravisCI would generate this data to use when commenting on a PR via a hook. This can be used with sampled ingestion data to monitor the effect of changes from `dev` to `master`.

PR #7, opened 2016-04-29
Add graphic fields

bgirard:
This is an improvement. I'm not familiar with this so you might want to review it carefully.

I left opengl/webgl/d2d/d3d11 as object but they will have an optional sub field 'failureId' that I'd really like to get into the dataset. But they also have a lot of other datafields. I think that's what the ["string", "null"] type might be but I wasn't sure.

I'm also not sure how to test this change?

### DIFF
--- a/telemetry/main.schema.json
+++ b/telemetry/main.schema.json
@@ -387,13 +387,25 @@
                   ]
                 },
                 "adapters": {
-                  "type": "array"
+                  "type": "array",
+                  "items": {
+                    "$ref": "#/definitions/adapter"
+                  }
                 },
                 "monitors": {
                   "type": [
                     "array",
                     "object"
-                  ]
+                  ],
+                  "items": {
+                    "$ref": "#/definitions/monitor"
+                  }
+                },
+                "features": {
+                  "type": "object",
+                  "items": {
+                    "$ref": "#/definitions/features"
+                  }
                 }
               }
             },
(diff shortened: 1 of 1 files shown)


SamPenrose:
Thanks for the PR! To test, "python validation/telemetry/validate_main.py" will catch simple issues. Real testing requires grabbing a cross-section of pings via the telemetry infrastructure, seeing how many pings fail various checks, and deciding whether the pings or the schema should be blamed. @Dexterp37 @mreid-moz any thoughts on process here?
---- Fixes to the PR below here ----

diff --git a/telemetry/main.schema.json b/telemetry/main.schema.json
index bc3dd4c..4b84265 100644
--- a/telemetry/main.schema.json
+++ b/telemetry/main.schema.json
@@ -668,7 +668,7 @@
       "type": "object",
       "properties": {
         "description": {
-          "type": "string"
+          "type": ["string", "null"]
         },
         "vendorID": {
           "type": "string"
@@ -677,22 +677,22 @@
           "type": "string"
         },
         "subsysID": {
-          "type": "string"
+          "type": ["string", "null"]
         },
         "RAM": {
-          "type": "integer"
+          "type": ["integer", "null"]
         },
         "driver": {
-          "type": "string"
+          "type": ["string", "null"]
         },
         "driverVersion": {
-          "type": "string"
+          "type": ["string", "null"]
         },
         "driverDate": {
-          "type": "string"
+          "type": ["string", "null"]
         },
         "GPUActive": {
-          "type": "bool"
+          "type": "boolean"
         }
       }
     },
@@ -786,7 +786,7 @@
           "type": "number"
         },
         "pseudoDisplay": {
-          "type": "false"
+          "type": "boolean"
         },
         "scale": {
           "type": "number"

Dexterp37:
Running the validations script as a spark job with one week's worth of FF data from version covered by the schema might do the trick. Maybe we should take one week for each month over the past N months.

We could then have the script report the failures and their count, along with an anonymised sample for each failure.

I think @georgf might have some useful insights as well.

mreid-moz:
I filed #5 a while ago to do a quick sanity check when a PR is submitted.

bgirard:
I've made the required changed and properly implemented the optional feature schema.

telemetry/main.schema.json
```diff
@@ -1,6 +1,3 @@
-{
```
mreid-moz: These first 3 lines seem to have gotten removed and should be re-added.

bgirard:
Ok sorry about the delay. All fixed now.

mreid-moz:
Looks good, thanks!

PR #10, opened 2016-08-02
Support core ping version 1 or higher

mreid-moz:
In reality, core pings are already up to version 7.


whd:
Incorporated into #13.

PR #13, opened 2016-10-03
Fix up schemas for first prod deploy

whd:
This incorporates #10. These are the changes I have determined empirically cause the new infrastructure to most closely resemble the current infrastructure from a validation perspective.

@mreid-moz or @trink r?

### DIFF
--- a/telemetry/main.schema.json
+++ b/telemetry/main.schema.json
@@ -14,15 +14,7 @@
           "pattern": "^[0-9]{10}"
         },
         "channel": {
-          "type": "string",
-          "enum": [
-            "nightly",
-            "aurora",
-            "beta",
-            "release",
-            "esr",
-            "default"
-          ]
+          "type": "string"
         },
         "name": {
           "type": "string"
@@ -299,15 +291,7 @@
                   "type": "boolean"
                 },
                 "channel": {
-                  "type": "string",
-                  "enum": [
-                    "nightly",
-                    "aurora",
-                    "beta",
-                    "release",
-                    "esr",
-                    "default"
-                  ]
+                  "type": "string"
                 },
                 "enabled": {
                   "type": "boolean"
--- a/telemetry/main.schema.json
+++ b/telemetry/main.schema.json
@@ -19,7 +19,9 @@
             "nightly",
             "aurora",
             "beta",
-            "release"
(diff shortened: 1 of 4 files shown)


telemetry/main.schema.json
```diff
@@ -19,7 +19,9 @@
             "nightly",
             "aurora",
             "beta",
-            "release"
+            "release",
+            "esr",
+            "default"
```
mreid-moz: `channel` should just be a string, I don't think we want to specifically enumerate the list of possible values (there are a lot of variations in the long tail, such as `release-cck-yahoo`).

telemetry/main.schema.json
```diff
@@ -302,7 +304,9 @@
                     "nightly",
                     "aurora",
                     "beta",
-                    "release"
+                    "release",
+                    "esr",
+                    "default"
```
mreid-moz: Same here.

telemetry/main.schema.json
```diff
@@ -478,7 +482,7 @@
                   "type": "string"
                 },
                 "version": {
-                  "type": "string"
+                  "type": ["string", "integer"]
```
mreid-moz: The [Telemetry Environment docs](https://gecko.readthedocs.io/en/latest/toolkit/components/telemetry/telemetry/data/environment.html) specify this as a string field, are you seeing cases where it's an integer?

telemetry/main.schema.json
whd: Why were we enumerating them in the first place then? I have updated the PR.

telemetry/main.schema.json
whd: Likewise updated.

telemetry/main.schema.json
whd: Yes, I made these schema changes by observing the "reason" fields for the most frequent decode errors and changed the schema until the error rates were sufficiently low (< 1%).

For this case specifically, I actually did a bit of looking around and I think the reason it is problematic is that this version is reported via some os-specific lookup that on Linux may return a number instead of a string. This might be a bug in Firefox (or the docs) but in any case, this is the easiest workaround.

mreid-moz:
Looks good

PR #16, opened 2016-10-26
bug 1288045 - update schema to account for changes from bug 1218576

chutten:
r? @mreid-moz

### DIFF
--- a/telemetry/main.schema.json
+++ b/telemetry/main.schema.json
@@ -561,6 +561,25 @@
               "required": [
                 "scalars"
               ]
+            },
+            "content": {
+              "type": "object",
+              "properties": {
+                "histograms": {
+                  "type": "object",
+                  "additionalProperties": {
+                    "$ref": "#/definitions/histogram"
+                  }
+                },
+                "keyedHistograms": {
+                  "type": "object",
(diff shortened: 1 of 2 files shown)


telemetry/main.schema.json
```diff
@@ -561,10 +561,30 @@
...
+                    }
+                  }
+                }
+              }
             }
           },
           "required": [
-            "parent"
+            "parent",
+            "content"
```
mreid-moz: Will all pings that have the `processes` section in the payload have both of these keys?

chutten: Yup. See [TelemetrySession around line 1312](https://dxr.mozilla.org/mozilla-central/rev/3f4c3a3cabaf94958834d3a8935adfb4a887942d/toolkit/components/telemetry/TelemetrySession.jsm#1312).

There _may_ have been a short period of time where it could have one but not the other... but I don't think it straddled a merge day.

mreid-moz: Ok, that should be fine. As long as we can safely discard data that has one but not the other (as not validating), we're good.

mreid-moz:
Looks good, thanks!

PR #21, opened 2017-01-23
Schemas from gregglind/shield-study-schemas, updated

gregglind:
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/mozilla-pipeline-schemas/21)

### DIFF
--- a/telemetry/shield-study-addon.schema.json
+++ b/telemetry/shield-study-addon.schema.json
@@ -1,58 +1,80 @@
 {
-  "$schema": "http://json-schema.org/draft-04/schema#",
-  "type": "object",
-  "title": "shield-study-addon doc_type schema",
-  "description": "Shield-Study-addon addon-specific data, sent as Map(s,s).  You may have to cast back at re:dash.",
-  "properties": {
-    "version": {
-      "type": "integer",
-      "title": "Version schema.  Will be 3",
-      "enum": [
-        3
-      ]
-    },
-    "study_name": {
-      "type": "string",
-      "minLength": 1,
-      "description": "Name of a particular study.  Usually the addon_id."
-    },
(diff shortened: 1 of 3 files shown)


mreid-moz:
Looks like the `description` fields were better before - they used to contain logical descriptions of what the fields should contain, now they seem to just describe the validation pattern.

Also, please move the `packet` field up to the very top level and change it's name to `type`, in keeping with the [`main`](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/telemetry/main.schema.json#L669-L675) and [`crash`](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/telemetry/crash.schema.json#L558-L563) schemas.

mreid-moz:
I don't think it's a good idea to include a `testing` field in the payload - that effectively means every query against this data will need to include `WHERE testing = false AND ...`. I would prefer to handle test data in a cleaner way, but I'm not sure what's the best option.

@gregglind how do you foresee `testing` data being used?

gregglind:
1.  WILLFIX.  Descriptions were auto-generated by de-referencing the pattern field.  I will see if there is a nice way of undoing that.  
2.  packet.  Happy to rename it.  BUT, moving it up to the top level doesn't help with the jsonschema "oneOf" "figure out which packet it is" validation issue.   I can solve this other ways.

3.  TESTING.  I want it to be easy for devs to dev, and not publish real data until the final second.  An alternative would be to send it... "somewhere else".  I think having "and testing == false" in sql isn't that weird.  I think it's "in total" less weird than having to point the data source at a different place (and keep that place existing), but I am open to ideas.

mreid-moz:
Looks good to me! Thanks.

PR #23, opened 2017-01-26
Don't enumerate specific channels for crash pings.

mreid-moz:
There are many other valid values for `channel`. Relax this requirement, similar to [this commit](https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/1149cb7a7f356f1a9888c664a805f2434823b7ef) for `main` pings.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/mozilla-pipeline-schemas/23)

### DIFF
--- a/telemetry/crash.schema.json
+++ b/telemetry/crash.schema.json
@@ -13,13 +13,7 @@
           "$ref": "#/definitions/buildId"
         },
         "channel": {
-          "type": "string",
-          "enum": [
-            "nightly",
-            "aurora",
-            "beta",
-            "release"
-          ]
+          "type": "string"
         },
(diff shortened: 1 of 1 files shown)


telemetry/crash.schema.json
```diff
@@ -13,13 +13,7 @@
           "$ref": "#/definitions/buildId"
         },
         "channel": {
-          "type": "string",
-          "enum": [
-            "nightly",
-            "aurora",
-            "beta",
-            "release"
-          ]
```
trink: 'Other' is missing from the enumeration, that should fix the issue

PR #26, opened 2017-01-30
Updated Shield Schemas, as seen at the collector AND PARQUET versions.

gregglind:
New folder:  "parquet" to show what goes into parquet, for re:dash.

Schemas now reflect aspects of the 'envelope' well.

These are derived from gregglind/shield-studies-schemas 0.7.0.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/mozilla-pipeline-schemas/26)


telemetry/shield-study-addon.schema.json
```diff
@@ -1,81 +1,266 @@
...
                 "shield-study-addon"
             ]
+        },
+        "version": {
+            "type": "number",
+            "minimum": 4,
+            "maximum": 4
+        },
+        "payload": {
+            "$schema": "http://json-schema.org/draft-04/schema#",
```
mreid-moz: This `$schema` field needs to go back to the top-level of the schema [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/26/files#diff-cf02c0cb76e53d1738f859bba28b6f02L2)

telemetry/shield-study-addon.schema.json
```diff
@@ -1,81 +1,266 @@
...
+            "type": "object",
+            "title": "shield-study-addon",
+            "description": "Shield-Study-addon addon-specific data, sent as Map(s,s).  You may have to cast back at re:dash.",
```
mreid-moz: This description seems wrong. It should be something like "Payload as sent by the shield study addon via telemetry"

gregglind:
@trink can you R+ this (or F+) for correctness parquet schemas?

robhudson:
I think it would be a good idea to have a README in the parquet/ folder saying that those files are generated and shouldn't be edited by hand.

gregglind:
I will do this.  F+

parquet/shield-study-addon.parquetmr.txt
```diff
@@ -0,0 +1,73 @@
...
+        required binary architecture (UTF8);
+        required binary buildId (UTF8);
+        required binary channel (UTF8);
+        required binary name (UTF8);
+        required binary platformVersion (UTF8);
+        required binary version (UTF8);
+    }
+    required binary clientId (UTF8);
+    required binary creationDate (UTF8);
+    required group environment {
```
trink: Although valid do we want required groups with all optional fields?

gregglind: It came derived from the json schema.  I imagine it's that we expect env to always exist (because adding a key is safe), but we aren't sure it can actually populate the stuff (because the code for generating the stuff can fail for weird reasons).

trink: It is only safe if you add another optional key.  In which case the parent being optional is still correct.

trink:
Other than that the schemas are valid parquet-mr but without some actually data I cannot make any claims about their correctness.

telemetry/shield-study-addon.schema.json
```diff
@@ -1,81 +1,265 @@
...
+                        },
+                        "update": {
+                            "type": "object",
+                            "properties": {
+                                "autoDownload": {
+                                    "type": "boolean"
+                                },
+                                "channel": {
+                                    "type": "string",
+                                    "enum": [
```
mreid-moz: Please delete the `enum` for the `channel` field in all places it appears - it should just allow any string value. The enum is overly restrictive and will cause lots of legitimate records to fail. My bad for having this in the refactored schemas I gave you earlier.

telemetry/shield-study-addon.parquetmr.txt
```diff
@@ -0,0 +1,73 @@
...
+        required binary buildId (UTF8);
+        required binary channel (UTF8);
+        required binary name (UTF8);
+        required binary platformVersion (UTF8);
+        required binary version (UTF8);
+    }
+    required binary clientId (UTF8);
+    required binary creationDate (UTF8);
+    optional group environment {
+        required group system {
```
trink: many other groups still need to be updated

telemetry/shield-study-error.schema.json
```diff
@@ -1,124 +1,308 @@
...
+                "architecture": {
+                    "type": "string"
+                },
+                "buildId": {
+                    "type": "string",
+                    "pattern": "^[0-9]{10}"
+                },
+                "channel": {
+                    "type": "string",
+                    "enum": [
```
mreid-moz: ditto

telemetry/shield-study-error.schema.json
```diff
@@ -1,124 +1,308 @@
...
+                        },
+                        "update": {
+                            "type": "object",
+                            "properties": {
+                                "autoDownload": {
+                                    "type": "boolean"
+                                },
+                                "channel": {
+                                    "type": "string",
+                                    "enum": [
```
mreid-moz: ditto

telemetry/shield-study.schema.json
```diff
@@ -1,100 +1,284 @@
...
+                "architecture": {
+                    "type": "string"
+                },
+                "buildId": {
+                    "type": "string",
+                    "pattern": "^[0-9]{10}"
+                },
+                "channel": {
+                    "type": "string",
+                    "enum": [
```
mreid-moz: and again

telemetry/shield-study.schema.json
```diff
@@ -1,100 +1,284 @@
...
+                        },
+                        "update": {
+                            "type": "object",
+                            "properties": {
+                                "autoDownload": {
+                                    "type": "boolean"
+                                },
+                                "channel": {
+                                    "type": "string",
+                                    "enum": [
```
mreid-moz: here too

telemetry/shield-study-addon.schema.json
```diff
@@ -1,81 +1,265 @@
...
+                "architecture": {
+                    "type": "string"
+                },
+                "buildId": {
+                    "type": "string",
+                    "pattern": "^[0-9]{10}"
+                },
+                "channel": {
+                    "type": "string",
+                    "enum": [
```
mreid-moz: remove enum

telemetry/shield-study-addon.schema.json
```diff
@@ -7,6 +7,208 @@
...
+                                        "type": "integer",
+                                        "minimum": 0
+                                    },
+                                    "signedState": {
+                                        "type": "integer"
+                                    }
+                                }
+                            }
+                        },
+                        "active Experiment": {
```
mreid-moz: s/active Experiment/activeExperiment/

mreid-moz:
@trink can you take another look?

trink:
Looks good

PR #27, opened 2017-01-30
Allow null for defaultSearch

whd:
...or acknowledge that we want to discard these.

@mreid-moz r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/mozilla-pipeline-schemas/27)


mreid-moz:
This field is [allowed to be null](http://gecko.readthedocs.io/en/latest/toolkit/components/telemetry/telemetry/data/core-ping.html#defaultsearch), so r+.

PR #30, opened 2017-02-14
shield-schemas-v3, 0.9.0 series

gregglind:
- jsonschema 'number' => double

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/mozilla-pipeline-schemas/30)


whd:
I'm r+ing this since it won't make things worse (per https://bugzilla.mozilla.org/show_bug.cgi?id=1337927) and we're using it in prod per https://bugzilla.mozilla.org/show_bug.cgi?id=1339611.

PR #34, opened 2017-03-24
Add regex validation to Telemetry events

fedepad:
This PR adds validation to Telemetry events by adding the regex pattern
to be matched for the following event fields: category, method, object.
The pattern to be matched was defined in Bug 1337022 at the link
https://bugzilla.mozilla.org/show_bug.cgi?id=1337022

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/mozilla-pipeline-schemas/34)

### DIFF
--- a/telemetry/main.schema.json
+++ b/telemetry/main.schema.json
@@ -846,13 +846,16 @@
           "minimum": 0
         },
         {
-          "type": "string"
+          "type": "string",
+          "pattern": "^[a-zA-Z][a-zA-Z0-9_.]*[a-zA-Z0-9]$"
         },
         {
-          "type": "string"
+          "type": "string",
+          "pattern": "^[a-zA-Z][a-zA-Z0-9_.]*[a-zA-Z0-9]$"
         },
         {
-          "type": "string"
+          "type": "string",
+          "pattern": "^[a-zA-Z][a-zA-Z0-9_.]*[a-zA-Z0-9]$"
         },
         {
           "type": ["string", "null"]
--- a/validation/telemetry/sample_v4_ping.json
+++ b/validation/telemetry/sample_v4_ping.json
@@ -10824,7 +10824,15 @@
             "key1": false,
             "key2": true
           }
-        }
+        },
+        "events": [
+          [2147, "ui", "click", "back_button"],
+          [2213, "ui", "search", "search_bar", "google"],
+          [2892, "ui", "completion", "search_bar", "yahoo",
+            {"querylen": "7", "results": "23"}],
+          [5434, "dom", "load", "frame", null,
+            {"prot": "https", "src": "script"}]
+        ]
       },
       "content": {
         "histograms": {


fedepad:
@Dexterp37 Here it is! Not 100% sure though!

Dexterp37:
Thank you @fedepad , this looks good. To make sure it works correctly:

- temporarily add the example events from the [doc](https://gecko.readthedocs.io/en/latest/toolkit/components/telemetry/telemetry/collection/events.html) to the `sample_v4_ping.json` file;
- use the `validate_main.py` script which lives in this repository.

You should see no error and the validation should pass.

fedepad:
@Dexterp37 Done, validation pass, it says ok! Do you want me to update `sample_v4_ping.json` with the events from doc that I added and push it in this PR? Inserted the events in:

`
    "fileIOReports": {},
    "events": [
        [2147, "ui", "click", "back_button"],
        [2213, "ui", "search", "search_bar", "google"],
        [2892, "ui", "completion", "search_bar", "yahoo",
                {"querylen": "7", "results": "23"}],
        [5434, "dom", "load", "frame", null,
                {"prot": "https", "src": "script"}]
    ],
    "histograms": {
      "A11Y_IATABLE_USAGE_FLAG": {
        "bucket_count": 3,
        "histogram_type": 3,
        "range": [
          1,
          2
        ],`

Dexterp37:
Yes, good idea. Let's update the sample JSON in this PR.

@fedepad , does the validation still pass if you use an invalid string longer than 2 characters?

fedepad:
@Dexterp37 I tried with quite few options, and the validation passes actually even with one letter invalid string, for example if I pass "!" instead of "ui".
As far as I understand should not pass, but it does pass validation.
For example even
`"events": [
        [-2, "!e~", "click", "back_button"]
  ]`
passes the validation, which it shouldn't for a couple of reasons:

1. timestamp should be positive
2. first character in "category" or the others defined as identifier should be an alphanumeric as far as I understand, instead it passes with non alphanumeric too...

Hints on what's happening?

Dexterp37:
Ah! Now I figured it out :-) I think you're adding the `events` in the wrong part of the `sample_v4_ping.json` file. You should add the events in the *parent* section of the *processes*, roughly [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/validation/telemetry/sample_v4_ping.json#L10827).

fedepad:
Perfect, good eye! 
I can confirm that validation of the sample in doc passes and also errors are shown if one voluntarily try to trigger them, for example, I tried the followings:

1. Negative event number
2. Non alphanumeric at the beginning of the "identifiers"
3. Non alphanumeric at the end of the "identifiers"
4. Single character identifier
5. Empty identifier
and all trigger errors as they should according to the schema.
I'm gonna push `sample_v4_ping.json` to this branch with the events from the doc that I added.

validation/telemetry/sample_v4_ping.json
```diff
@@ -10824,7 +10824,15 @@
             "key1": false,
             "key2": true
           }
-        }
+        },
+        "events": [
+          [2147, "ui", "click", "back_button"],
+	  [2213, "ui", "search", "search_bar", "google"],
```
fedepad: I don't know why it's not indented as the line above...I see it's indented in my local editor... 
Do you have an idea why? @Dexterp37

Dexterp37: @fedepad you're mixing a tab with whitespaces on that line. Just use whitespaces and you should be fine! :-)

fedepad: Thanks! Should be ready now!

Dexterp37:
@fedepad , one last request before merging this: would you mind squashing your commits into one and pushing again? You can keep the commit message from the first commit.

fedepad:
@Dexterp37 Done!

Dexterp37:
Great job, r+!

I'll take this for a test-drive on our real-time CEP (Hindsight) before merging it.

Dexterp37:
@trink , any chance you could take this for a test drive? I'm having troubles accessing the CEP box I used before.

telemetry/main.schema.json
```diff
@@ -846,13 +846,16 @@
           "minimum": 0
         },
         {
-          "type": "string"
+          "type": "string",
+          "pattern": "^[a-zA-Z][a-zA-Z0-9_.]*[a-zA-Z0-9]$"
```
trink: It seems like this should be factored out and accessed via a reference instead of being repeated 3 times.

Dexterp37: @fedepad , could you follow-up on this?

fedepad: @Dexterp37 Yes, I will.

fedepad: @trink I have no idea how to do that and what is the syntax to accomplish what you suggest...
If you could guide me I will do it and push again!

Dexterp37: @fedepad , this slightly changed meanwhile :) We're now using templatized schemas that get build off the schema sections in templates/include/telemetry. We don't need to change multiple schemas anymore, but making this change has become slightly more involved. If you're still interested in working on it, at the very least you should:

-  rebase the patch and move your changes to the schema to the [template file](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/include/telemetry/processData.1.schema.json#L50);
- create a new file called _pattern_event_string.1.schema.json_ with a content similar to [this one](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/include/common/pattern_uuid.1.schema.json), but with your regex;
- reference the pattern in your schema change as done [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/include/telemetry/clientId.1.schema.json#L3)

@trink do we need to build and commit the output schemas too (e.g. schemas/telemetry/*)?

trink: The build process updates the schemas/telemetry directory (with a beautified, sorted output) and yes those files should be included as part of the PR.

trink:
Closing, please correct the PR as instructed above and submit a new PR against the dev branch (as per the readme)

PR #44, opened 2017-05-02
Bug 1337927 - Add version support to schema filenames

mreid-moz:
Change the repository layout to allow multiple versions of each document's schema. 

See also [related decoder changes](https://github.com/mozilla-services/lua_sandbox_extensions/pull/126) in the ingestion code.

I populated `core` schema version 1 using a slightly modified version of the initial commit (with fields ordered to match the current version for easier `diff`ing), and all others up to v8 as copies of the current schema (with version pinned to a specific value).

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/mozilla-pipeline-schemas/44)


mreid-moz:
Note that I did not add a version to the "vacuous" schema as that doesn't seem to make much sense. Do we even need that schema or should I just drop it altogether?

trink:
rm vacuous.schema.json

mreid-moz:
Done

schemas/telemetry/core/core.1.schema.json
```diff
@@ -0,0 +1,41 @@
+{
```
trink: core.1.schema.json	failed to parse offset:745 Missing a name for object member.

mreid-moz: Fixed.

PR #45, opened 2017-05-09
Restrict event extra values to strings

georgf:
This is per https://bugzilla.mozilla.org/show_bug.cgi?id=1339195#c3

Extra objects for events are restricted to string keys and string
values.


georgf:
I think this is fine. Prior validation showed no issues at all and we only receive the navigation.search event so far.

PR #46, opened 2017-05-10
Generate schemas dynamically (WIP)

wlach:
This is a straw-man prototype of using the m4 macro preprocessor + make
to generate the schemas dynamically. If the idea looks ok, I would also:

* Replace many more instances of redundant json with macros (we could also
  consider programmatic ones for schemas like "main" where the only difference
  is a number)
* Expand the validation tests and run them automatically on the generated
  output via travis


wlach:
@mreid-moz do you want to take a quick look at this?

trink:
fyi: we have been standardizing more or cmake it is a little sad to see us slipping backwards

wlach:
Are there practical advantages of using cmake here? I am not wedded to make, I just used it because I know it and it seemed like the simplest tool that would get the job done. I would not propose using GNU make for building real software in this day and age.

mreid-moz:
This looks quite promising to me!

Per IRC discussion, it would be great to have the `.json.in` files in a `src` directory or similar, and then put the built artifacts at their current location under `/schemas` for compatibility with the validation code.

More generally, I will try to put together some non-hand-wavy requirements for what we'd like to be able to do with the dynamically-generated schemas. Otherwise it's quite hard to evaluate different approaches in a meaningful way.

wlach:
Updated PR to build the artifacts under `schemas/`, do the environment substitution for more pings, and consolidate the dependant definitions (e.g. `monitor`).

I'll defer future work until we decide what approach we want to take. I think most of this could be reused if we wanted to use e.g. cmake to do the building/template substitution.

wlach:
Closing in favor of  #55

PR #47, opened 2017-05-11
Reference other schemas using the $ref property

Dexterp37:
This additionally cleans up the validate_pings.py file.
More importantly, this allows to reference other schemas
living in other files: this way we don't have to define
common sections (e.g. environment) in all the ping definitons
using them.


Dexterp37:
@mreid-moz this might be especially interesting for you, so flagging you here :-D

I'm wondering if this way of referencing the schemas will work on when the schemas are used on the pipeline side. Any clue there?

Dexterp37:
Hah, I just saw #46 :-D That's interesting timing!

mreid-moz:
@Dexterp37 can you take a look at #55 and see if it will suit your needs?

Dexterp37:
It does!

PR #48, opened 2017-05-12
Add a schema for the 'new-profile' ping

Dexterp37:
This just checks in the schema using the "old" process, bringing in a copy of the environment in the schema. The 'new-profile' ping is being deployed, so I'm not sure we should block on #47 or #46 .

### DIFF
--- a/schemas/telemetry/new-profile/new-profile.4.schema.json
+++ b/schemas/telemetry/new-profile/new-profile.4.schema.json
@@ -1,0 +1,777 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "new-profile",
+  "properties": {
+    "application": {
+      "type": "object",
+      "properties": {
+        "architecture": {
+          "type": "string"
+        },
+        "buildId": {
+          "type": "string",
+          "pattern": "^[0-9]{10}"
+        },
+        "channel": {
+          "type": "string"
+        },
+        "name": {
+          "type": "string"
+        },
+        "platformVersion": {
+          "type": "string",
+          "pattern": "^[0-9]{2,3}\\."
+        },
+        "version": {
+          "type": "string",
+          "pattern": "^[0-9]{2,3}\\."
+        },
+        "displayVersion": {
+          "type": "string",
+          "pattern": "^[0-9]{2,3}\\."
+        },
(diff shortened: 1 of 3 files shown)


Dexterp37:
@mreid-moz r?

schemas/telemetry/new-profile/new-profile.1.schema.json
```diff
@@ -0,0 +1,777 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "name": "new-profile",
```
mreid-moz: I believe this should be `title` rather than `name`.

validation/telemetry/new-profile_v1_ping.json
```diff
@@ -0,0 +1,334 @@
+{
+	"type" : "new-profile",
```
mreid-moz: Why tabs in this file instead of spaces?

Dexterp37: Whoops. :)

schemas/telemetry/new-profile/new-profile.1.schema.json
Dexterp37: Gosh, you're right... and most of the schemas in this repo are wrong :(

Dexterp37:
@mreid-moz should be good now! :)

schemas/telemetry/new-profile/new-profile.1.schema.json
```diff
@@ -0,0 +1,777 @@
...
+    },
+    "type": {
+      "type": "string",
+      "enum": [
+        "new-profile"
+      ]
+    },
+    "version": {
+      "type": "number",
+      "minimum": 4,
```
mreid-moz: Woops, I was a bit too quick approving there - looks like the filename should be `new-profile.4.schema.json` rather than `new-profile.1.schema.json`

mreid-moz:
Ok, this looks good now. Has it been tested on the hindsight test box?

Dexterp37:
There are no 'new-profile' pings coming in just yet. This schema was used by QA to do client-side validation in bug 1120370.

mreid-moz:
Ok. Let's keep an eye on the failure rate once they start coming in.

PR #49, opened 2017-05-15
Bug 1357043 - Disallow negative subsessionLength

whd:
Can `subsessionLength` be 0 or null?


mreid-moz:
This looks good to me, pending the outcome of the test @trink is running

PR #50, opened 2017-06-05
Remove the "meta" object from the ping schemas

Dexterp37:
The metadata for each ping is not added on the client, so
we don't need to have a "meta" object in these schemas.


Dexterp37:
@trink , can you review this?

PR #52, opened 2017-06-15
Make environment, payload and clientId mandatory

Dexterp37:
This is the final demonstration that copy/pasting does not solve all the problems in the world.


Dexterp37:
@trink r?

PR #54, opened 2017-06-19
Bug 1369038 - Update the telemetry schemas for the new sandbox settings key

alex:


schemas/telemetry/crash/crash.4.schema.json
```diff
@@ -316,6 +316,9 @@
             },
             "userPrefs": {
               "type": "object"
+            },
+            "sandbox": {
+              "type": "object"
```
Dexterp37: Since we know which properties will live in the "sandbox" section, we should probably validate them as well. For example, this could become:

```
"sandbox": {
  "type": "object",
  "properties": {
    "effectiveContentProcessLevel": {
      "type": ["integer", "null"],
      "minimum": 0
    }
  }
}
```

This tells the schema that _sandbox_ can contain a property called _effectiveContentProcessLevel_ which can be either null or an integer with a minimum value of 0 (assuming that's the correct minimum).

The same change must be done in the other files as well.

Dexterp37:
Also, thanks @alex for your patch :)

alex:
Done. I did not include the `minimum` key because hypothetically on nightly builds someone could send a lower value, and I think I'd like those to be included.

Dexterp37:
Looks good to me, thanks @alex ! Let me flag the real repo-owners now ;-)

r? @mreid-moz  :D

trink:
r-
I will include this in the new shared environment schema so in the future these changes only need to occur in one place.

alex:
Sorry, not sure I follow. Is there something I need to do now, or just wait for the shared environment schema bits to land?

trink:
It will land next Monday with this change added

alex:
👍 thanks

trink:
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/include/telemetry/environment.1.schema.json#L359

PR #55, opened 2017-06-22
Add packaging, schema templates/generation/beautification and integra…

trink:
…tion tests


trink:
replaces PR #46

README.md
```diff
@@ -1,8 +1,53 @@
...
+[Parquet Schema Examples](https://mozilla-services.github.io/lua_sandbox_extensions/parquet/io_modules/lpeg/parquet.html)
+is a limited resource as the full specification is only documented in code.
+
+## Build
+
+### Note
+
+All schemas are generated from the 'templates' directory and written into the
+'schemas' directory (i.e., the artifacts are generated/saved back into the
+repository as per mreid's request).
```
wlach: I think we should specify the rationale for this (even if not everyone agrees with it?), rather than just making it sound like we're doing this based on one person's arbitrary decision. :)

README.md
```diff
@@ -1,8 +1,53 @@
...
+
+The [Parquet-MR](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md)
+schemas are used for direct to parquet output.
+
+To learn more about Parquet-MR Schemas.
+[Parquet Schema Examples](https://mozilla-services.github.io/lua_sandbox_extensions/parquet/io_modules/lpeg/parquet.html)
+is a limited resource as the full specification is only documented in code.
+
+## Build
+
```
wlach: I think this section is a mistake?

README.md
trink: It is one person's arbitrary decision. @mreid care to write it up?

trink:
The formatting changes are part of the artifact generation which also alters the content since they now use standardize templates e.g, old schemas with a subset of the environment will now have a full environment. definitions go away etc.). The main alteration was some metadata corrections to replace 'name' with 'title', the main ping type enumeration also changed (removing saved-session) but that is about it.  Unfortunately factoring out those changes doesn't make the diff any less cumbersome.

The fact the schemas are now tested and production data has been run through all the new versions it is more rigorous then anything in the past.  The main question is does the overall process fit your needs?

README.md
trink: I will reword this when mreid provides the artifact clarification.

The [Parquet-MR](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md) schemas are used for direct to parquet output; some examples of Parquet-MR schemas can be found here: [Parquet Schema Examples](https://mozilla-services.github.io/lua_sandbox_extensions/parquet/io_modules/lpeg/parquet.html)

README.md
mreid-moz: Will do

mreid-moz: The general requirements for the templated approach, as I see them:

- Ability to factor out common schema components (telemetry "environment" section, clientId / docId spec, etc).
- One source schema can be used to generate multiple output schemas. Example: `main` schema should be used for both `main` and `saved-session` doctypes.
- Generate other schema formats from canonical JSONSchema representation (namely parquet-mr schema representation, but possibly others in the future).
- Produce standalone single-document JSON Schemas. No external `$ref`s. This is to avoid the need to implement custom ref resolvers in multiple languages/platforms.
- Simplify the workflow for adding a new schema, particularly if it shares common components.

Regarding storing generated artifacts in the repository, the rationale is twofold. 

First, it lets us easily see and refer to complete schemas as they are actually used. This means that the schemas can be referenced directly in bugs and such, as well as being fetched directly from the repo for testing other schema consumers (*test* being important here, as any production use should be using the installable packages). 

Second, it gives us a changelog for each schema, rather than having to reason about changes to templated external pieces and when/how that impacted a given doctype's schema over time. This means that it should be easy to look back in time for the provenance of different parts of the schema for each doc type.

trink:
Changes complete.  If you want me to reformat the current schemas, commit them to master and rebase this PR onto the new master let me know (for the sake of reducing the artifact diffs)

wlach:
@trink If you could push a reformatting commit to master and rebase this on that I think it would be helpful for review/auditing purposes (not just in this PR, but in the future too). I admit it's rather unlikely that this will reveal any problems but it seems worth the small amount of effort required given how critical this stuff is.

schemas/telemetry/core/core.2.schema.json
```diff
@@ -32,6 +32,12 @@
     "arch": {
       "type": "string"
     },
+    "experiments": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
```
wlach: Is there something we could do (maybe an option passed to jq) to get the object properties of the schema in sorted order?

trink: Now sorted to make future artifact compares easier... not going to update the master and reconstruct this PR again since it only impacts only a few small diffs.

PR #59, opened 2017-07-19
Bug 1382223 - Allow null values in focus-event ping

fbertsch:
This allows null values in 3 places:
1. As values in the settings map
2. As values in the Events list
3. As values in the extra field of the Events list

The schema validates against the example. @trink can you take a peek at the parquet schema and make sure it seems correct?

### DIFF
--- a/schemas/telemetry/focus-event/focus-event.1.schema.json
+++ b/schemas/telemetry/focus-event/focus-event.1.schema.json
@@ -24,13 +24,13 @@
             "type": "string"
           },
           {
-            "type": "string"
+            "type": [ "string", "null" ]
           },
           {
             "additionalProperties": {
-              "type": "string"
+              "type": [ "string", "null" ]
             },
-            "type": "object"
+            "type": [ "object", "null" ]
           }
         ],
         "maxItems": 6,
@@ -54,7 +54,7 @@
     },
     "settings": {
       "additionalProperties": {
-        "type": "string"
+        "type": [ "string", "null" ]
       },
       "type": "object"
     },
--- a/schemas/telemetry/focus-event/focus-event.1.parquetmr.txt
+++ b/schemas/telemetry/focus-event/focus-event.1.parquetmr.txt
@@ -19,7 +19,7 @@ message focus_event {
   required group settings (MAP) {
     repeated group key_value {
       required binary key (UTF8);
-      required binary value (UTF8);
+      optional binary value (UTF8);
     }
   }
   required group events (LIST) {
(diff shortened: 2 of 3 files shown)


schemas/telemetry/focus-event/focus-event.1.schema.json
```diff
@@ -54,7 +54,7 @@
     },
     "settings": {
       "additionalProperties": {
-        "type": "string"
+        "type": [ "string", "null" ]
```
trink: unnecessary in an object

schemas/telemetry/focus-event/focus-event.1.schema.json
```diff
@@ -24,11 +24,11 @@
             "type": "string"
           },
           {
-            "type": "string"
+            "type": [ "string", "null" ]
           },
           {
             "additionalProperties": {
-              "type": "string"
+              "type": [ "string", "null" ]
```
trink: you cannot have a NULL key in an object

schemas/telemetry/focus-event/focus-event.1.schema.json
```diff
@@ -24,11 +24,11 @@
             "type": "string"
           },
           {
-            "type": "string"
+            "type": [ "string", "null" ]
```
trink: can only this one array member be NULL?

fbertsch: Yes, see the spec here: http://gecko.readthedocs.io/en/latest/toolkit/components/telemetry/telemetry/collection/events.html#eventtelemetry

schemas/telemetry/focus-event/focus-event.1.schema.json
fbertsch: The validation breaks on the example ping if I remove this. I believe it's indicating the value is NULL, not the key.

schemas/telemetry/focus-event/focus-event.1.schema.json
fbertsch: Actually, the object below it can be null as well. I'll make it `"type": [ "object", "null" ]`

schemas/telemetry/focus-event/focus-event.1.schema.json
trink: yup, ignore

fbertsch:
@trink the parquet schema validated against some sample error pings. The event format here now allows `null` for `extra` field.

trink:
Yes, looks good

PR #61, opened 2017-07-21
Enable schema validation for environment.experiments

Dexterp37:
This PR also contains the update schemas and an update main ping sample. The generated schema validates correctly against the example.

I'm not sure why the generation step also changed `focus-event`. Looks like this is reverting the hotfix from 89a43b8? @fbertsch , any clue?

### DIFF
--- a/validation/telemetry/main.4.sample.json
+++ b/validation/telemetry/main.4.sample.json
@@ -304,6 +304,11 @@
       "version": "45.0",
       "xpcomAbi": "x86_64-gcc3"
     },
+    "experiments": {
+      "test_id123": {
+        "branch": "brunch"
+      }
+    },
     "partner": {
       "distributionId": null,
       "distributionVersion": null,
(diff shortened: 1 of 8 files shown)


trink:
Ha, yeah the generated output was modified instead of the template input.  All modification should happen in https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/master/templates

fbertsch:
@Dexterp37 looks like I updated the wrong files. I put in a fix, maybe we should merge that before merging this.

Dexterp37:
Thanks @fbertsch! @trink I think this is ready for another look (assuming you should be the one reviewing it). I cannot set the reviewer nor assign this to myself :-\

validation/telemetry/main.4.sample.json
```diff
@@ -304,6 +304,11 @@
       "version": "45.0",
       "xpcomAbi": "x86_64-gcc3"
     },
+    "experiments": {
```
trink: In general I like seeing real world examples but over all it looks good to me.
```json
"e10sCohort":{"branch":"disqualified-test"}}"
```

PR #67, opened 2017-07-28
Introduce the "update" ping schema

Dexterp37:
This is not to check in. For some reason, the `cmake ..` step fails with with the following error:

> CMake Error at CMakeLists.txt:48 (message):
>   invalid JSON telemetry/update/update.4.schema.json - No such file or
>   directory
> 

The "No such file or directory" part is there because I've printed the `JQ_RV` result variable in `CMakeLists.txt`.

### DIFF
--- a/schemas/telemetry/update/update.4.schema.json
+++ b/schemas/telemetry/update/update.4.schema.json
@@ -1,0 +1,872 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "properties": {
+    "application": {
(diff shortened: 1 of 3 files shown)


trink:
- Missing the parquet output schema
- At some point we should break `version` out as a template in the JSON schema

trink:
The reason for the push back to create all the pieces up front (in/out schema and monitoring/output cfg) is because historically building it out as we go has caused two data outages already (test pilot and focus) and we continue not to address the issue.

PR #70, opened 2017-08-01
Move the puppet-config templates here to allow for self-service

trink:
Now all the changes to bring a new/modified ping type on-line can
be made in a single repository.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 cmake_minimum_required(VERSION 3.0)
-project(mozilla-pipeline-schemas VERSION 0.0.1 LANGUAGES NONE)
+project(mozilla-pipeline-schemas VERSION 0.0.3 LANGUAGES NONE)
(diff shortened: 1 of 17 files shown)


whd:
I've tested this style of configuration and it'll mostly work. We need to change the puppet structure slightly to differentiate between CEP and DWL configurations, since the alerts will go on the CEP but the parquet outputs will go on the DWL. I think having `/etc/puppet/modules/pipeline/templates/hindsight/{cep,dwl}` would suffice.

As an aside, these aren't actually schemas so it's a bit misleading to put them in this RPM, but I still think it's better than having to manage them entirely in puppet.

whd:
Another piece that would be nice to manage outside of puppet is the s3 schemas used by the outputs, which currently live [here](https://github.com/mozilla-services/puppet-config/tree/master/pipeline/modules/pipeline/files/schema). We could e.g. auto-generate the list of valid doctypes from the schemas in this repo (as part of the make step to keep with the checked-in artifacts theme).

trink:
Closing will let @mreid-moz team drive this

PR #75, opened 2017-08-08
simple schemas

SamPenrose:
Add payload budget schemas per bug 1370529.


trink:
The telemetry directory is for Firefox client data schemas; this in not a supported ping type.  Also in the future you will want to run cmake to create the artifacts which are also committed.

SamPenrose:
I'll run cmake. Which directory would you like it in?

trink:
What is the data source?

SamPenrose:
https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/moz_telemetry/sandboxes/heka/analysis/moz_telemetry_payload_size.lua#L42

mreid-moz:
Can we call these "payload-size" rather than "budget"?

trink:
It should go here https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/master/templates/heka/telemetry (no JSON/input schema as it is a Heka message)

SamPenrose:
RFAL.

fbertsch:
Closed in favor of #103

PR #77, opened 2017-08-21
Remove pioneer-heatmap schema, replace with pioneer custom ping.

hannosch:
The pioneer-heatmap study won't be launched at all. In order to make sending the pioneer data easier, this defines one custom ping compatible schema to send data from all pioneer studies. For the time being, this also leaves the data encrypted with JOSE/JWE at rest.

This is WIP, to be discussed with @georgf in context of https://bugzilla.mozilla.org/show_bug.cgi?id=1369737


georgf:
We talked through this and this approach is sound to me.

The ping example in "pioneer-study.1.sample.json" looks sensible.
This has a shorter time-frame (Fx57), so going with the standard client send mechanism and the generic end-points is a feasible solution.

Medium-/longer-term, we should talk about how to better support custom end-points from Firefox Telemetry (whether that is just sending the blob elsewhere or custom submission mechanics per endpoint).

Please make sure that data review is covered and ping documentation available in a public place.

hannosch:
@trink, @mreid-moz thoughs on this PR?

This will require changes on the backend DWL instance to read and parse the data. I'm not sure if there's already code for that.

The goal here is to make the client side logic simpler for version 1, so we can re-use the normal telemetry upload mechanism. This leaves the question of when and who decrypts the data and validates the data inside the encrypted blob open. Each study addon will have a different schema for what's inside this blob. But given the new urgency of getting something up and running, I want to punt on those questions for version 2.

schemas/telemetry/pioneer-study/pioneer-study.1.parquetmr.txt
```diff
@@ -0,0 +1,20 @@
+message pioneer-study {
+  required binary creationDate (UTF8);
+  required binary id (UTF8);
+  required binary type (UTF8);
+  required double version;
+  required group payload {
+    required binary encryptedData (UTF8);
```
trink: I thought the original point of this exercise was to not make each consumer decrypt the data?

hannosch: Yes, but right now we don't know the schema of what's going to be inside the encrypted blob for the first study. heatmap as such isn't going to launch at all. lonnen still wants to get the pipeline as such up and running quickly, so for a version 1 I decided to punt on the encryption part and let the first study consumer deal with it on its own. Using the common ping format also means we can re-use the client side telemetry code, which is another big win for a version 1 of this.

trink:
This was already tested end to end (edge->dwl->parquet data lake almost 3 months ago) according to the original specification. The entire payload was encrypted and the submission URL contained the necessary information for decryption and validation after which the decrypted document is written as parquet to a restricted S3 bucket. The current proposal is the polar opposite and should go through another design and security review as it does not meet the original data consumption requirements.

trink:
If you want to go that path it should work like telemetry.where json and parquet schemas are provided for each study name/version.  The common message will be parsed and validated the payload will be decrypted, parsed, and validated and a combination of the metadata and decrypted payload will be written to parquet.

mreid-moz:
@hannosch If we go with the approach herein (where data remains encrypted to the data lake), how will we handle the secrets required to read and decrypt the data at analysis time?

hannosch:
@mreid-moz That would be up to each individual study.

Another approach is to only do the decryption part on a custom DWL in hindsight, without any schema validation of the study specific data. In the original proposal anything in the pioneer namespace was routed to a custom DWL and hindsight running on it assumed the entire payload it got was an encrypted blob. In the new format, the pioneer-study ping could cause the data to routed to such a custom DWL. Instead of assuming the entire payload is an encrypted blob, it could assume the structure of the ping stays constant and look for the `payload/encryptedData` and `payload/encryptionKeyId` fields inside the data it gets and apply JWE decryption on the `payload/encryptedData` blob, using the identifier provided in `payload/encryptionKeyId` as the key id. In the parquet output the `payload/encryptedData` would than hold the actually decrypted data or one could rename the field in the output to make its content clearer.

All of this leaves the technical details of how to read the data and the way each individual study validates its data out of scope. The timeline I got from lonnen was to try to get something ready, so the first study could start sending data to prod by end of September, meaning the study client side needs it done some 2 weeks before that in order to do some actual integration testing. Given that timeline I don't think we can figure out a good and final solution to how any of this should work. This is just a version 1 proposal to get us started.

trink:
There is nothing custom about the DWL we are just using a separate instance due to the sensitive nature of the data.  Whether the encrypted data is the entire message or a field is insignificant from a decoding standpoint.   The important part is the data validation as I don't want to see a repeat of what happened with shield 'just put something in and fix later' which doesn't work because the clients have such a long tail and we have to support bad decisions for years. The fact this is now a fire-drill (again) even though it has been available for production deployment for three months is a bit frustrating..  Clearly it hasn't been high priority and it should slip until a viable end to end process is defined.

templates/telemetry/pioneer-study/pioneer-study.1.schema.json
```diff
@@ -0,0 +1,75 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "properties": {
+    "creationDate": {
```
trink: Include the common definition    
@TELEMETRY_CREATIONDATE_1_JSON@,

templates/telemetry/pioneer-study/pioneer-study.1.schema.json
```diff
@@ -0,0 +1,75 @@
...
+      "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]{3}Z$",
+      "type": "string"
+    },
+    "id": {
```
trink: @TELEMETRY_ID_1_JSON@,

templates/telemetry/pioneer-study/pioneer-study.1.schema.json
```diff
@@ -0,0 +1,75 @@
...
+    },
+    "type": {
+      "description": "doc_type, restated",
+      "enum": [
+        "pioneer-study"
+      ],
+      "type": "string"
+    },
+    "version": {
+      "maximum": 1,
```
trink: Incorrect version telemetry is currently at 4

templates/telemetry/pioneer-study/pioneer-study.1.parquetmr.txt
```diff
@@ -0,0 +1,20 @@
+message pioneer-study {
```
trink: The parquet output schemas should be study specific (the generic version isn't necessary at all)
i.e. templates/pioneer-study/foobar.1.parquetmr.txt

templates/telemetry/pioneer-study/pioneer-study.1.schema.json
```diff
@@ -0,0 +1,75 @@
+{
```
trink: There also needs to be study specify schemas for validation after decryption.
i.e. templates/pioneer-study/foobar.1.schema.json

trink: This should be renamed to pioneer-study.4.schema.json

hannosch: fixed

templates/telemetry/pioneer-study/pioneer-study.1.schema.json
hannosch: fixed

templates/telemetry/pioneer-study/pioneer-study.1.schema.json
hannosch: fixed

templates/telemetry/pioneer-study/pioneer-study.1.schema.json
hannosch: fixed, thx. didn't know this specified the telemetry version, rather than the version of the individual ping schema

hannosch:
@trink Regarding the review comments: There currently isn't a schema for a first study yet. The goal is to get the mechanism as such up and running, so the launching the first study won't be blocked on the getting the infrastructure in place.

> This new design incurs two JSON parses/validations due to nested design plus the decryption step (this nested encoding is what really want to avoid).

I think the original proposal did have merit. But given the suddenly tight deadline, I don't think it's doable on the client end, as it would require rewriting the gather, upload and retry logic. We get that logic for free from using the telemetry format. I think georg agreed that we could make changes to the telemetry client code going forward, but those will need to ride the trains and as such are a few months out.

> What is the expected submission size (compressed/uncompressed) and what is the expected number of submissions per day?

We don't have a concrete study proposal yet. I've seen two things being mentioned, one is a study by Rebecca needing a cohort size of 40-50 thousand users and a collection period of a couple months. I don't know anything about the study contents. The other is a proposal to compare the Telemetry and Pioneer cohorts to see how representative the Pioneer cohort is. This requires a one-time data blob per user, so will amount to at most one kilobyte per user.

Since we are growing the Pioneer cohort slowly, it will be thousands to tens of thousands of users this year and maybe hundreds of thousands next year. It might grow to low millions in the very end.

Based on that, I'm assuming it's ok to do a non-optimal version 1 of the pipeline now, while it still has to deal with little data and optimize it for a version 2 in the next coming months.

schemas/telemetry/pioneer-study/pioneer-study.4.schema.json
```diff
@@ -0,0 +1,75 @@
...
+          "type": "string"
+        },
+        "studyName": {
+          "description": "Name of a particular study. Usually the addon_id.",
+          "maxLength": 100,
+          "minLength": 1,
+          "pattern": "^\\S+$",
+          "type": "string"
+        },
+        "studyVersion": {
```
trink: I missed one.  If this is the version we are using for the nested validation schema it should be an integer.

mreid-moz:
I'm OK with this proposal for round one. I highly recommend deploying a very simple "hello world" pioneer study to ensure that we can do a simple analysis of the encrypted data. Is that feasible on this timeline?

schemas/telemetry/pioneer-study/pioneer-study.4.schema.json
hannosch: Changed to integer version for the schema. I copied the semantic version approach from the shield study schema.

hannosch:
Last commit adds a probably incorrect and incomplete Pioneer example study specific schema and the corresponding Parquet schema. I've assumed the study specific part ends up in a new "data" group inside the payload of the envelope schema (corresponding to the contents of the `encryptedData` field). And the `encryptionKeyId` field is not forwarded to the Parquet schema, as it is no longer relevant after decryption.

schemas/telemetry/pioneer-study/pioneer-study.4.schema.json
trink: Yeah that is the exact problem: you are using the last shoe-horned 'need it now' hack as the basis for new development.  A few months from now someone will say we are just doing what pioneer does and try to further propagate the technical debt and processing overhead.

templates/pioneer-study/example.1.parquetmr.txt
```diff
@@ -0,0 +1,21 @@
...
+    required double studyVersion;
+  }
+  required group metadata {
+    required int64  Timestamp;
+    required binary documentId (UTF8);
+    optional binary geoCity (UTF8);
+    optional binary geoCountry (UTF8);
+    optional binary normalizedChannel (UTF8);
+  }
+}
```
trink: I was thinking something more like this (the schema of the study plus a metadata object): 
```
message pioneer-study-example {
  required group metadata {
    required int64  Timestamp;
    required binary documentId (UTF8);
    required binary pioneerId (UTF8);
    required binary studyName (UTF8);
    required int64 studyVersion;
    optional binary geoCity (UTF8);
    optional binary geoCountry (UTF8);
    optional binary normalizedChannel (UTF8);
  }        
  required binary exampleString (UTF8);
}
```

templates/telemetry/pioneer-study/pioneer-study.4.schema.json
```diff
@@ -0,0 +1,67 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "properties": {
+    @TELEMETRY_CREATIONDATE_1_JSON@,
+    @TELEMETRY_ID_1_JSON@,
```
trink: You will want to pull in the required application section here     @TELEMETRY_APPLICATION_1_JSON@,  Also update the validation input

validation/telemetry/pioneer-study.1.sample.json
```diff
@@ -0,0 +1,13 @@
+{
```
trink: Rename the file with a version number of 4

templates/pioneer-study/example.1.parquetmr.txt
```diff
@@ -0,0 +1,21 @@
+message pioneer-study {
```
trink: the path should be templates/pioneer-study/example/example.1.parquetmr.txt

templates/pioneer-study/example.1.schema.json
```diff
@@ -0,0 +1,16 @@
+{
```
trink: the path should be  templates/pioneer-study/example/example.1.schema.json

You will also want a validation/pioneer-study/example.1.sample.json containing
```json
{ "exampleString" : "foobar" }
```

templates/pioneer-study/example.1.parquetmr.txt
hannosch: If that's doable, that's indeed much better - no more `payload/data/what-I-really-want` nesting :)

hannosch:
@trink I think I've covered all your comments. I've rebased and squashed the earlier two commits, but left the example schema and my review changes in separate ones. Feel free to squash-merge this if you want to. Otherwise I'll get to it on Saturday.

One last thing: There's currently both a top level schema and templates directory with much the same contents (except schema is after expanding the template bits). I've manually created the files in the schema directory. Are those still required or is the template expansion part of the build process now? Could the schema directory be removed from Git?

trink:
The schema directory is for generated artifacts see: https://github.com/mozilla-services/mozilla-pipeline-schemas#notes. To build the artifacts see: https://github.com/mozilla-services/mozilla-pipeline-schemas#cmake-build-instructions

schemas/telemetry/pioneer-study/pioneer-study.4.schema.json
```diff
@@ -0,0 +1,73 @@
...
+    "payload": {
+      "description": "",
+      "properties": {
+        "encryptedData": {
+          "description": "JOSE/JWE encrypted payload.",
+          "type": "string"
+        },
+        "encryptionKeyId": {
+          "description": "JOSE/JWK key id, e.g. pioneer-20170520.",
+          "type": "string"
```
trink: do you want a pattern on this?

hannosch: Not right now. I'll leave it up to ops to decide the naming scheme and don't want to presume anything.

trink:
The integration tests will fail due to pioneerId not being part of the stubbed out metadata. I will fix this when I bump the package version.

PR #78, opened 2017-08-21
Add a Dockerfile to run tests with.

hannosch:
This is based on the latest hindsight docker container, which already installs most tools and libraries.

The Makefile is a convinience, so one can run `make` or `make test` without having to remember the full docker invocations.

Currently tests fail with two different errors, one has to do with the treat json null as parquet null change, the other failure is in the validation telemetry.focus-event.1 test.

@trink thoughts on the dockerfile and makefile?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+**/.DS_Store
+.dockerignore
+.git
+.gitignore
+Dockerfile
+Makefile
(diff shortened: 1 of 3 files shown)


trink:
The official plan is Travis-CI integration, with artifact verification, integration test, RPM package generation on tagging with an upload to the github release tag.

As for pioneer specifically you are better off running the full integration tests here
https://github.com/mozilla-services/lua_sandbox_extensions/tree/master/moz_pioneer/tests/hindsight which now needs to leverage the schema package (came into existence after the pioneer module) so we can get rid of the duplication.

PR #79, opened 2017-08-28
Add support for the "update" ping with reason "success"

Dexterp37:
This additionally adds a ParquetMR schema for serializing the ping using direct 2 parquet.


Dexterp37:
Good point, I've just added it to the repo, thanks for noticing!

trink:
And name it properly so it is actually run ;)
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/tests/hindsight/run/input/test_json_schemas.lua#L72

PR #84, opened 2017-09-18
Bug 1400939 Add parquet output for URI errors

trink:
r? @mreid-moz


mreid-moz:
Can you split the pioneer-study stuff into a separate commit or PR?

trink:
I could but is there a point?  It is just the generated artifacts that someone forgot to check-in but are already part of the currently deployed package.

mreid-moz:
Not really, just that they don't have anything to do with the commit message here.

trink:
done

PR #88, opened 2017-10-04
Add  displayVersion to the update_parquet schema

Natim:
Refs https://bugzilla.mozilla.org/show_bug.cgi?id=1405614
f? @mreid-moz

### DIFF
--- a/schemas/telemetry/update/update.4.parquetmr.txt
+++ b/schemas/telemetry/update/update.4.parquetmr.txt
@@ -2,15 +2,7 @@ message update {
   required binary id (UTF8);
   required binary clientId (UTF8);
   required group application {
-    required binary architecture (UTF8);
-    required binary buildId (UTF8);
-    required binary channel (UTF8);
     required binary displayVersion (UTF8);
-    required binary name (UTF8);
-    required binary platformVersion (UTF8);
-    required binary vendor (UTF8);
-    required binary version (UTF8);
-    required binary xpcomAbi (UTF8);
   }
   required group metadata {
     required int64  Timestamp;
--- a/schemas/telemetry/update/update.4.parquetmr.txt
+++ b/schemas/telemetry/update/update.4.parquetmr.txt
@@ -1,6 +1,17 @@
 message update {
   required binary id (UTF8);
   required binary clientId (UTF8);
+  required group application {
+    required binary architecture (UTF8);
+    required binary buildId (UTF8);
+    required binary channel (UTF8);
+    required binary displayVersion (UTF8);
+    required binary name (UTF8);
+    required binary platformVersion (UTF8);
+    required binary vendor (UTF8);
+    required binary version (UTF8);
+    required binary xpcomAbi (UTF8);
+  }
   required group metadata {
     required int64  Timestamp;
     optional binary Date (UTF8);
--- a/templates/telemetry/update/update.4.parquetmr.txt
+++ b/templates/telemetry/update/update.4.parquetmr.txt
@@ -1,6 +1,9 @@
 message update {
   required binary id (UTF8);
   required binary clientId (UTF8);
+  required group application {
+    required binary displayVersion (UTF8);
+  }
   required group metadata {
     required int64  Timestamp;
     optional binary Date (UTF8);
@@ -16,7 +19,6 @@ message update {
       required binary architecture (UTF8);
       required binary version (UTF8);
       required binary buildId (UTF8);
-      required binary displayVersion (UTF8);
       required binary vendor (UTF8);
       optional binary hotfixVersion (UTF8);
     }
@@ -57,10 +59,8 @@ message update {
     optional binary targetChannel (UTF8);
     optional binary targetVersion (UTF8);
     optional binary targetBuildId (UTF8);
-    optional binary targetDisplayVersion (UTF8);
     optional binary previousChannel (UTF8);
     optional binary previousVersion (UTF8);
     optional binary previousBuildId (UTF8);
-    optional binary previousDisplayVersion (UTF8);
   }
 }
--- a/templates/telemetry/update/update.4.parquetmr.txt
+++ b/templates/telemetry/update/update.4.parquetmr.txt
@@ -2,15 +2,7 @@ message update {
   required binary id (UTF8);
   required binary clientId (UTF8);
   required group application {
-    required binary architecture (UTF8);
-    required binary buildId (UTF8);
-    required binary channel (UTF8);
     required binary displayVersion (UTF8);
-    required binary name (UTF8);
-    required binary platformVersion (UTF8);
-    required binary vendor (UTF8);
-    required binary version (UTF8);
-    required binary xpcomAbi (UTF8);
   }
   required group metadata {
     required int64  Timestamp;
--- a/templates/telemetry/update/update.4.parquetmr.txt
+++ b/templates/telemetry/update/update.4.parquetmr.txt
@@ -2,7 +2,15 @@ message update {
   required binary id (UTF8);
   required binary clientId (UTF8);
   required group application {
+    required binary architecture (UTF8);
+    required binary buildId (UTF8);
+    required binary channel (UTF8);
     required binary displayVersion (UTF8);
+    required binary name (UTF8);
+    required binary platformVersion (UTF8);
+    required binary vendor (UTF8);
+    required binary version (UTF8);
+    required binary xpcomAbi (UTF8);
   }
   required group metadata {
     required int64  Timestamp;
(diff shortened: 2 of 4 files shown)


templates/telemetry/update/update.4.parquetmr.txt
```diff
@@ -16,6 +16,7 @@ message update {
       required binary architecture (UTF8);
       required binary version (UTF8);
       required binary buildId (UTF8);
+      required binary displayVersion (UTF8);
```
mreid-moz: You need to add a new `required group application` for this field - it comes from the application part of the submission [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/telemetry/update/update.4.schema.json#L4-L38).

That group should live at the same level as `metadata` or `environment`.

It should look something like
```
required group application {
  required binary displayVersion (UTF8);
}
```

templates/telemetry/update/update.4.parquetmr.txt
```diff
@@ -56,8 +57,10 @@ message update {
     optional binary targetChannel (UTF8);
     optional binary targetVersion (UTF8);
     optional binary targetBuildId (UTF8);
+    optional binary targetDisplayVersion (UTF8);
```
mreid-moz: This field is not available in the incoming `update`  pings, as far as I can tell. A client-side code change would be required to add this field.

templates/telemetry/update/update.4.parquetmr.txt
```diff
@@ -56,8 +57,10 @@ message update {
...
     optional binary previousChannel (UTF8);
     optional binary previousVersion (UTF8);
     optional binary previousBuildId (UTF8);
+    optional binary previousDisplayVersion (UTF8);
```
mreid-moz: This field is not available either, I don't think.

Dexterp37:
@mreid-moz yes, I think you're right! Let's go and add the application section as a whole. @Natim , lplease also update the documentation after that (mozilla/firefox-data-docs).

trink:
- Run cmake to generate the artifact and add it to the PR
- Update the test data
- Verify it works

mreid-moz:
@Natim please update per @trink suggestions

Natim:
Is it possible to add what @trink said as a Travis-CI configuration?

I tried locally but I can't really make sense of the output:

```
$ ctest -V -C hindsight
UpdateCTestConfiguration  from :~/mozilla/mozilla-pipeline-schemas/release/DartConfiguration.tcl
Parse Config file:~/mozilla/mozilla-pipeline-schemas/release/DartConfiguration.tcl
UpdateCTestConfiguration  from :~/mozilla/mozilla-pipeline-schemas/release/DartConfiguration.tcl
Parse Config file:~/mozilla/mozilla-pipeline-schemas/release/DartConfiguration.tcl
Test project ~/mozilla/mozilla-pipeline-schemas/release
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 1
    Start 1: mozilla-pipeline-schemas-hindsight

1: Test command: ~/mozilla/mozilla-pipeline-schemas/release/hindsight/run.sh
1: Test timeout computed to be: 1500
1: 1507280477197922884 [info] hindsight starting
1: 1507280477204638708 [error] input.test_json_schemas run/input/test_json_schemas.lua:10: module 'lfs' not found:
1: 	no file '/usr/lib/luasandbox/modules/lfs.lua'
1: 	no file '/usr/lib/luasandbox/io_modules/lfs.lua'
1: 	no file '/usr/lib/luasandbox/modules/lfs.so'
1: 	no file '/usr/lib/luasandbox/io_modules/lfs.so'
1: 1507280477204677895 [error] input_plugins input.test_json_schemas lsb_heka_create_input failed
1: 1507280477204682272 [error] input_plugins test_json_schemas.cfg create_inputs_plugin failed
1: 1507280477205126717 [info] input.test_heka_messages starting
1: 1507280477205192926 [info] input.test_heka_messages detaching received: 0 msg: 
1: 1507280477207323406 [info] analysis_plugins starting thread: 0
1: 1507280477207352854 [debug] input_reader analysis0 opened file: output/input/0.log
1: 1507280477207818863 [error] output.test_parquet_schemas run/output/test_parquet_schemas.lua:9: module 'cjson' not found:
1: 	no file '/usr/lib/luasandbox/modules/cjson.lua'
1: 	no file '/usr/lib/luasandbox/io_modules/cjson.lua'
1: 	no file '/usr/lib/luasandbox/modules/cjson.so'
1: 	no file '/usr/lib/luasandbox/io_modules/
1: 1507280477207853436 [error] output_plugins output.test_parquet_schemas lsb_heka_create_output failed
1: 1507280477207857173 [error] output_plugins output.test_parquet_schemas create_output_plugin failed
1: 1507280478208929789 [info] hindsight input plugins have exited; cascading shutdown initiated
1: 1507280478209023962 [info] hindsight stop signal received
```

schemas/telemetry/update/update.4.parquetmr.txt
```diff
@@ -1,6 +1,17 @@
 message update {
   required binary id (UTF8);
   required binary clientId (UTF8);
+  required group application {
+    required binary architecture (UTF8);
+    required binary buildId (UTF8);
```
Natim: I seems to me that a lot of fields are redundant with the one in `environment.build` is that fine?

Dexterp37: Correct - I wonder if it makes sense to just keep the application.displayVersion at this point. @mreid-moz ?

mreid-moz: Hmm... it's been a while since I've looked into this, but last time I did, I thought that the values in [`application` ](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/telemetry/TelemetryController.jsm#355-365) were available more often / more reliably than the ones in [`environment.build`](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/telemetry/TelemetryEnvironment.jsm#1286-1296). If that's so, it's preferable to keep the entire `application` group and remove the dupes from `environment.build`. If not, please let me know so I can update my understanding :)  At the least, the `application` section should be available for all pings whether they include the `environment` or not, so it should be more "standard".

Natim:
I merged the travis config so that we can see if tests are passing on this PR.

templates/telemetry/update/update.4.parquetmr.txt
```diff
@@ -1,6 +1,17 @@
 message update {
   required binary id (UTF8);
   required binary clientId (UTF8);
+  required group application {
```
trink: Common sections should be updated here https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/templates/include/telemetry/ and included.  If displayVersion is truly required you will need to create a new version of the application template.

schemas/telemetry/update/update.4.parquetmr.txt
Dexterp37: Mh, weird: both sections should be there and have identical values for matching fields. Unless there's a bug (in the client), of course :-D There should be no mismatches for pings that report the environment.

With that said, let's just add the full application and remove the dupes from the environment as @mreid-moz said. Mark, I imagine this change would require backfilling, right?

mreid-moz: Ok, as long as both sets of fields should be the same, and are equally reliable, I am fine with the original approach of just adding `application.displayVersion`. That should avoid the need for backfill and minimize impact on any existing queries / analyses.  Sorry for the back-and-forth on this :-/

templates/telemetry/update/update.4.parquetmr.txt
mreid-moz: @trink what do you think about just including the one field that's needed here per [this comment](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/88/files#r143238502) ? It seems preferable to repeating a whole bunch of redundant fields, or having to backfill to move things around.

Natim:
I am going to hand over this PR to @Dexterp37 because I don't understand all the conversation that are going on here.

Dexterp37:
@Natim , I don't have the time to work on this right now I'm afraid. However, all you need to do is drop the changes that were made to `application.2.parquetmr.txt`: they are not being used anymore.

templates/telemetry/update/update.4.parquetmr.txt
trink: Won't this be a problem with Athena if you want to add another field later?

mreid-moz: Everything else from the `application` section is already present in the `environment`.

PR #93, opened 2017-10-20
Add the targetDisplayVersion to the update ping

Dexterp37:
This optional field is only added to the ping with
reason = ready and may not have a format that strictly
matches the one in application.displayVersion.


Dexterp37:
Mh, the test sems to fail with "hindsight_cli: command not found". @Natim , any clue on why this could be failing?

Natim:
> may not have a format that strictly matches the one in application.displayVersion.

Why?

validation/telemetry/update.4.ready.json
```diff
@@ -18,7 +18,8 @@
     "reason" : "ready",
     "targetChannel" : "default",
     "targetVersion" : "56.0a1",
-    "targetBuildId" : "20080811053724"
+    "targetBuildId" : "20080811053724",
+    "targetDisplayVersion" : "56.0 Beta 1"
```
Natim: I think what we are aiming at is `56.0b1` rather than `56.0 Beta 1`

@sylvestre can you confirm that it is the format of displayVersion?

Natim:
> the test sems to fail with "hindsight_cli: command not found". @Natim , any clue on why this could be failing?

https://travis-ci.org/mozilla-services/mozilla-pipeline-schemas/builds/290333716#L1103

I guess the version number changed :)

validation/telemetry/update.4.ready.json
Dexterp37: As mentioned [in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1405614#c5), that's the information reported by the update MAR. Telemetry just takes it and reports it back.

If you want/need that to change, you should reach out to the Update team and file a bug to change it.

PR #102, opened 2017-11-29
WIP: mozza events

fbertsch:
Considerations:
- The partitioning column will be `property`, e.g. "which web property is sending this data?" but I'm not attached to that name. Alternatives could be simply partitioning by base URL.
- We could add more dimensional information; a good one might be some sort of user or session id (for eventual funnel analyses)
- Moved to a common event JSON schema
- The parquet schema will be written when we finalize what this data will look like

### DIFF
--- a/schemas/mozza/event/event.1.schema.json
+++ b/schemas/mozza/event/event.1.schema.json
@@ -1,0 +1,73 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "properties": {
+    "browser": {
+      "type": "string"
+    },
+    "browser_version": {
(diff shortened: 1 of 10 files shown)


fbertsch:
@mreid-moz RFAL: tests pass, and the enum will limit the number of partitions. This means if new properties want to send us events, they will need to register first.

fbertsch:
@mreid-moz looking over what ga does, I split it into `domain` and `request_uri`. `domain` should then be very small on disk as well, since it usually won't change much for a single `property`. See here [0].

[0] https://support.google.com/analytics/answer/1012243?hl=en

PR #104, opened 2017-11-30
Add mozza event parquet output

fbertsch:
:whd, just want to make sure that d2p for moz_ingest will work similarly. It seems all of those metadata fields are available. Feel free to reassign reviewers if you're busy.

### DIFF
--- a/schemas/telemetry/focus-event/focus-event.1.parquetmr.txt
+++ b/schemas/telemetry/focus-event/focus-event.1.parquetmr.txt
@@ -27,18 +27,19 @@ message focus_event {
   required group events (LIST) {
     repeated group list {
       required group element (TUPLE) {
-        required int64 timestamp;
-        required binary category (UTF8);
-        required binary method (UTF8);
-        required binary object (UTF8);
-        optional binary value (UTF8);
-        optional group extra (MAP) {
-          repeated group key_value {
-            required binary key (UTF8);
-            optional binary value (UTF8);
-          }
-        }
-      }
+  required int64 timestamp;
+  required binary category (UTF8);
+  required binary method (UTF8);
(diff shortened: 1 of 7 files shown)


schemas/mozza/event/event.1.parquet.txt
```diff
@@ -0,0 +1,29 @@
+message event {
+  required group metadata {
+    required binary documentId (UTF8);
+    required int64  Timestamp;
+    required binary Date (UTF8);
```
whd: This is a header specified by the client so will only be available if the client sends it.

fbertsch:
@whd I think that's okay for our purposes, we control the client so we should ensure it's available. RFAL

templates/include/common/event.1.parquetmr.txt
```diff
@@ -0,0 +1,13 @@
+(TUPLE) {
```
trink: Not liking the partial object template unfortunately the other options aren't much better with this simple templating.

I would get rid of the extra indent spacing (just go with the standard 2) as it adds little value and won't align properly in the two different contexts where the template is used.  Added #105 to address formatting in general.

schemas/telemetry/focus-event/focus-event.1.parquetmr.txt
```diff
@@ -39,6 +39,7 @@ message focus_event {
           }
         }
       }
+
```
trink: We discussed cleaning this up with `string(STRIP "${${VARNAME}}" ${VARNAME})` but I would rather have it addressed by #105

fbertsch:
@trink I'll leave the pretty-printing for later. Changed the indentation, RFAL

PR #107, opened 2017-12-01
Bug 1422349 - Change size to double

fbertsch:


tests/hindsight/run/input/test_heka_messages.lua
```diff
@@ -55,7 +55,7 @@ local messages = {
             build = "20170518000419",
             channel = "release",
             docType = "main",
-            size = 4242
+            size = 4242.1
```
whd: Isn't this already a double?

mreid-moz:
Can we cast this to an int on the lua side? Since it's a size in bytes, a double doesn't make too much sense.

whd:
We can.

tests/hindsight/run/input/test_heka_messages.lua
fbertsch: I suppose it is, how did the tests pass?

trink: It doesn't fail because it doesn't run (the schema name is incorrect in the input) it should be `EnvVersion = "heka.telemetry_payload_size.1"`

trink: Looking at the original code:
1. why can't we just fix the message to store size as an int64 (does it negatively impact down stream consumers?)
1. even better: the analysis plugin generating these messages is a pretty bad hack requiring a custom cfg to override sensible defaults.  With a small change to the direct to parquet code we can get rid if it all together and stop doubling the message volume going through Hindsight.

whd: 1. is what mreid proposed earlier. I don't think it will impact downstream consumers since the only ones I am aware of are CEP filters.

templates/heka/telemetry/telemetry_payload_size.1.parquetmr.txt
```diff
@@ -4,6 +4,6 @@ message telemetry_payload_size {
     required binary build (UTF8);
     required binary channel (UTF8);
     required binary docType (UTF8);
-    required int64 size;
+    required double size;
   }
 }
```
trink: 1. Proposed schema
```
message telemetry_payload_size {
  required int64 Timestamp;
  required int64 size; -- << add support for d2p output of size
  required group Fields {
    required binary build (UTF8);
    required binary channel (UTF8);
    required binary docType (UTF8);
  }
}
```
2. switch the budget dashboard to use this parquet output (the outputs can run in parallel during the migration)
3. remove moz_telemetry_payload_size.lua and its S3 protobuf stream output

trink: https://github.com/mozilla-services/lua_sandbox_extensions/commit/018f20dae29b776df22a663e4da138a8b2105213

mreid-moz: f+, this plan sounds good to me.

fbertsch: Also sounds good to me. I don't think we need to worry about the switch over, since the dashboard isn't using this data yet anyways.

mreid-moz: For the record, the "dashboard" piece is [Bug 1370529](https://bugzilla.mozilla.org/show_bug.cgi?id=1370529), and we can follow up on that separately. Once we have this data in parquet form, we might be able to migrate that dashboard over to re:dash.

fbertsch:
@trink RFAL with the new schema

PR #98, opened 2017-11-22
Added schema for esper-pioneer-shield-study

motin:
Add-on repo: https://github.com/motin/esper-pioneer-shield-study

PR #108, opened 2017-12-05
Revert "Added schema for esper-pioneer-shield-study"

trink:
Reverts mozilla-services/mozilla-pipeline-schemas#98


trink:
There is no good way to clean this up short of a hard reset, the changes will be applied back to dev

PR #112, opened 2017-12-05
Add 'processStartTimestamp' as a required field for mobile-event and an optional field for focus-event.

justindarc:
This can be used in reporting to convert event timestamps from being relative to the process start time to being standard absolute timestamps.

### DIFF
--- a/templates/telemetry/mobile-event/mobile-event.1.schema.json
+++ b/templates/telemetry/mobile-event/mobile-event.1.schema.json
@@ -2,11 +2,6 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
   "title": "mobile-event",
-  "properties": {
-    @TELEMETRY_MOBILEEVENT_1_JSON@,
-    "processStartTimestamp": {
-      "type": "integer"
-    }
-  },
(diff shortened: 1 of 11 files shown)


justindarc:
R? @fbertsch

templates/telemetry/mobile-event/mobile-event.1.schema.json
```diff
@@ -2,5 +2,11 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
   "title": "mobile-event",
-  @TELEMETRY_MOBILEEVENT_1_JSON@
+  "properties": {
+    @TELEMETRY_MOBILEEVENT_1_JSON@,
+    "processStartTimestamp": {
```
fbertsch: I would instead move `processStartTime` to the `mobile-event` template (so it would be included with `focus-event` as well), include `properties` in that template, and have the only difference between `mobile-event` and `focus-event` JSON schemas be the `required` section (where `mobile-event` should have `processStartTimestamp`.

justindarc: Done.

PR #113, opened 2017-12-19
Bug 1421787 - Add deletion ping schema

relud:
based on health ping with no payload

r? @fbertsch


relud:
thank you, did those and filed new PR

PR #123, opened 2018-02-14
Add the schema for the ICQ v1 SHIELD study

Dexterp37:
Let's review but not merge: it's pending data-review!

### DIFF
--- a/schemas/telemetry/shield-study-icq/shield-study-icq.4.schema.json
+++ b/schemas/telemetry/shield-study-icq/shield-study-icq.4.schema.json
@@ -1,0 +1,933 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "properties": {
+    "application": {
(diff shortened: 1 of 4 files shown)


georgf:
@mreid-moz, a quick question:
In case we would see any unexpected ping formats that don't fit the schema here, what would happen to those pings?
Would they go to the error stream and could we recover them later?

mreid-moz:
Yes, validation failures will go to the error stream. We can theoretically examine / recover them, though not without significant effort [[Citation Needed]]. @whd can you comment on the ability to recover error stream data?

whd:
Recovering from the error stream is possible, but not automated. It's "fairly easy" in that the error stream pings can be passed as-is into the original telemetry decoder to produce the backfilled result, and I have ad-hoc scripts for doing this. It's also fast since as long as the pings are bucketed correctly, which going forward is guaranteed by having schemas.

PR #130, opened 2018-03-08
Bug 1308040 - Add "gc" schema for main pings

mreid-moz:
Note that per [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1308040#c40), the schema may be overly restrictive.

@trink can we test this against the real data to see if it's causing any significant validation failures?


templates/include/telemetry/gc.1.schema.json
```diff
@@ -0,0 +1,16 @@
...
+      "items": @TELEMETRY_GCITEM_1_JSON@
+    },
+    "random": {
+      "type": "array",
+      "maxItems": 2,
+      "items": @TELEMETRY_GCITEM_1_JSON@
+    }
+  },
+  "additionalProperties": false
+}
```
trink: Add newlines to both files

acmiyaguchi:
[Here's a notebook](https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/8325) taking advantage of some code I wrote to figure out where all the invalid shield pings were coming from.

This should work since the current schema is currently permissive w.r.t. the gc metrics.

trink:
I just tested against 3MM+ entries from the 3/12/2018 landfill and there were no new errors generated by the gc schema in that data set (and yes, the new schema is active)

templates/include/telemetry/gc.1.schema.json
mreid-moz: done

mreid-moz:
Thanks @acmiyaguchi and @trink, sounds like the existing data should be OK.

@trink I added the missing newlines, does this look right now?

PR #132, opened 2018-03-15
Bug 1442699 - Specify valid histogram values

acmiyaguchi:
I set this so keys and values must be in the shape of integers. From my cursory scan of all of our histograms, this is the defacto way to serialize them. I validated this against nightly pings from a day in this month.


templates/include/telemetry/histogram.1.schema.json
```diff
@@ -33,7 +33,13 @@
       "minimum": 0
     },
     "values": {
-      "type": "object"
+      "type": "object",
+      "patternProperties": {
+        "^[0-9].*$": {
```
mreid-moz: I believe this would allow keys like "3foo". For all-digits (and disallowing empty keys), you'd need something like `"^[0-9]+$"`.

templates/include/telemetry/histogram.1.schema.json
```diff
@@ -33,7 +33,13 @@
...
+          "type": "string",
```
mreid-moz: The value-type should be `"integer"` here.

acmiyaguchi: Ah, this passed validation because I didn't set "additionalProperties" to false.

acmiyaguchi:
@mreid-moz RFAL

mreid-moz:
It was concerning to me that the tests passed despite setting the expected "value" type to `"string"` rather than `"integer"`, so I dug in a bit more and it looks like we are not enforcing the histogram schema at all right now :-/

The histogram template has a key `"histogram"` which is preventing this schema from being applied to all keys within the `"histograms"` section.  We should remove that key, which I've attempted over in #134.

acmiyaguchi:
@trink Can you help me diagnose the unit-tests here?

The schema is failing to validate against the sample ping, but looking at the json specification for `patternProperties` and the ping, I don't understand why this case is failing.

```
1: 1521743911631329860 [debug] input.test_json_schemas loading	../../validation/telemetry/main.4.sample.pass.json
1: 1521743911634328816 [info] input.test_json_schemas detaching received: 1 msg: process_message() run/input/test_json_schemas.lua:89: SchemaURI: #/properties/payload/properties/histograms/additionalProperties/properties/values Keyword: type DocumentURI: #/payload/histograms/A11Y_IATABLE_USAGE_FLAG/values/0
```

I wrote a small python script to test the python jsonschema behavior.

```
import json
import jsonschema

schema = json.load(open("schemas/telemetry/main/main.4.schema.json"))
test_pass = json.load(open("validation/telemetry/main.4.sample.pass.json"))
test_fail= json.load(open("validation/telemetry/main.4.histogram.fail.json"))

validator = jsonschema.Draft4Validator(schema)

print("testing passing case")
for e in validator.iter_errors(test_pass):
    print(e.validator, e.schema_path, e.path)

print("testing failing case")
for e in validator.iter_errors(test_fail):
    print(e.validator, e.schema_path, e.path)
```

This gives the expected behavior.
```
testing passing case
testing failing case
(u'additionalProperties', deque([u'properties', u'payload', u'properties', u'histograms', u'additionalProperties', u'properties', u'values', u'additionalProperties']), deque([u'payload', u'histograms', u'INVALID_BUCKET_TYPE', u'values']))
```

Is this an issue with the underlying rjson library?


=====

Separately, I've validated data against dev and this branch and it has shown no change.

https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/8754/command/8755

trink:
@acmiyaguchi it looks like this issue: https://github.com/Tencent/rapidjson/issues/825

acmiyaguchi:
@mreid-moz RFAL

acmiyaguchi:
I trimmed the failing test case by 90% (mostly removing addons and histogram fields). I'm also going to run the validation script from 20180314-20180321 on a 5% sample of all incoming main pings.

acmiyaguchi:
```
20180314 - 0.00% error
20180315 - 0.00% error
20180316 - 0.01% error
20180317 - 0.00% error
20180318 - 0.01% error
20180319 - 0.00% error
20180320 - 0.00% error
20180321 - 0.01% error
```

There were a couple of histograms that had values below the minimum, including `MAIN_THREAD_RUNNABLE_MS/AsyncEventDispatcher` and `PRCONNECTCONTINUE_BLOCKING_TIME_NORMAL`. I think this looks normal though.

PR #134, opened 2018-03-19
Ensure that sections of the schema are being used

mreid-moz:
There were a few sub-sections that had extra key names and such that resulted in them not properly being applied (so any "object" data would pass validation).

This PR ensures that `events`, `histograms`, and `scalars` sections are validated.

There is risk with this change, as we are effectively making the validation significantly more strict.

### DIFF
--- a/templates/include/telemetry/mainPayload.1.schema.json
+++ b/templates/include/telemetry/mainPayload.1.schema.json
@@ -53,7 +53,7 @@
       "type": "object",
       "additionalProperties": {
         "additionalProperties": {
-          @TELEMETRY_PROCESSDATA_1_JSON@
+          @TELEMETRY_HISTOGRAM_1_JSON@
         }
       }
     },
--- a/templates/include/telemetry/mainPayload.1.schema.json
+++ b/templates/include/telemetry/mainPayload.1.schema.json
@@ -30,10 +30,8 @@
       "type": "object",
       "properties": {
         "parent": {
-          "allOf": [
-          { @TELEMETRY_PROCESSDATA_1_JSON@ },
-          { "required": [ "scalars" ] }
-            ]
+          @TELEMETRY_PROCESSDATA_1_JSON@ ,
+          "required": [ "scalars" ]
         },
         "content": { @TELEMETRY_PROCESSDATA_1_JSON@ },
         "gpu": { @TELEMETRY_PROCESSDATA_1_JSON@ }
--- a/templates/include/telemetry/processData.1.schema.json
+++ b/templates/include/telemetry/processData.1.schema.json
@@ -1,24 +1,17 @@
-"processData": {
   "type": "object",
   "properties": {
     "scalars": {
       "type": "object",
       "additionalProperties": {
-        "scalar": {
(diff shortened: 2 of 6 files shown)


mreid-moz:
Note that if you add "?w=1" to the diff URL, it will hide all the indentation changes and be easier to compare.

templates/include/telemetry/processData.1.schema.json
```diff
@@ -1,24 +1,17 @@
-"processData": {
   "type": "object",
   "properties": {
     "scalars": {
       "type": "object",
       "additionalProperties": {
-        "scalar": {
-          "type": [ "integer", "string", "boolean" ]
-        }
+        "type": [ "integer", "string", "boolean" ]
```
trink: Should scalars be a template since it is repeated in multiple places?

trink:
Looking at the history; the original definitions we imported directly as the new templates and the definition name became the extra level of nesting..

mreid-moz:
re: nested `additionalProperties`, I believe this is correct for both `keyedHistograms` and `keyedScalars`.

`keyedHistograms` are of the form:
```json
"keyedHistograms": {
  "HISTOGRAM1_NAME": {...},
  "HISTOGRAM2_NAME": {
    "key1": {
      "bucket_count": 16,
      "histogram_type": 1,
      "range": [1, 15],
      "sum": 258,
      "values": {
        "0": 0,
        "1": 234,
        "6": 4,
        "7": 0
      }
    },
    "key2": {
      "bucket_count": 16,
      "histogram_type": 1,
      "range": [1, 15],
      "sum": 100,
      "values": {
        "0": 0,
        "1": 100,
        "7": 0
      }
    }
  }
}
```

So within `keyedHistograms`, we have `additionalProperties` of histogram names (for example `HISTOGRAM2_NAME`), and within each histogram, we have `additionalProperties` of key names (`"key1"` and `"key2"`).  

Keyed scalars are similar in that they are of the form:
```json
"keyedScalars": {
  "scalar1_name": {
    "key1": 1,
    "key2": 37
  },
  "scalar2_name": {
    "key3": false,
    "key4": true
  }
}
```

I made a simplified example [here](https://jsonschemalint.com/#/version/draft-06/markup/json?gist=6edf437f43389081ab64118d2b0339cb) that you can look at for comparison.

templates/include/telemetry/mainPayload.1.schema.json
```diff
@@ -55,7 +53,7 @@
       "type": "object",
       "additionalProperties": {
         "additionalProperties": {
-          @TELEMETRY_PROCESSDATA_1_JSON@
+          @TELEMETRY_HISTOGRAM_1_JSON@
```
mreid-moz: @trink I noticed this while looking at the nested `additionalProperties` too - looks like this should be including the histogram definition rather than the "processData" one as it was previously. Does that sound right?

templates/include/telemetry/processData.1.schema.json
mreid-moz: Yes, I filed #135 to follow up on this.

PR #136, opened 2018-03-19
Small fix to the gc schema

mreid-moz:
The array field is "slices_list" rather than "slices". Also rename
the gc sample data so it is actually used in the test.


mreid-moz:
The `oneOf` approach to handling the various `gc` schemas wasn't working, so this makes the earlier iteration more permissive (allowing the `slices` field to be either a `number` or an `array`).

I went through the commit history for [GCTelemetry.jsm](https://searchfox.org/mozilla-central/source/toolkit/components/telemetry/GCTelemetry.jsm#112-115) to find the "high water mark" for each of the constant max lengths, so those should now allow historic versions as well.

mreid-moz:
@trink this is ready to test against real data

acmiyaguchi:
As another sanity check, I've done more modifications on a moztelemetry based validation script to make schema changes easy to interpret. 

https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/8325/command/8328

mreid-moz:
I'm not sure what happened to the travis-ci checks on this PR... I ran the tests locally and all seems well.

PR #139, opened 2018-03-21
Issue 138 - Fix the properties in the telemetry environment template

trink:
- Add validation of the schemas on the test load and removed the python
  code


trink:
We try to go by the spec and yes there have been a bunch of things not spec'ed as NULL that are but I don't think we should preemptively allow it for everything.  We should also take a hard look at the schemas and only validate things we would want to fail an entire submission. e.g. I don't think we would want to fail a submission for some addon logging a string instead of an int if it was the only error in the document and me don't actually use it in a warehouse table.

acmiyaguchi:
The lack of this pattern in some other schema's explain a large number of otherwise benign validation errors. For example, `isDefaultBrowser` was causing a large number of pings in the pioneer shield studies to fail validation because it could be a null value.

Taking a hard look at the entire schema seems requires combing over most of the higher level parts of the schema carefully. I'm trying to put the right level of effort into making sure that validating the histograms properly won't cause any serious issues. But maybe each of the major fields in the environment should be looked at piece by piece? Is it best to do a comparison of the error validation rates before and after schema changes to determine if it'll make a significant impact on the number of pings that flow into analysis.

trink:
The Heka message breaks the JSON into pieces so you can just work with the part you care about.   Independent validation of the various sections would be interesting but it does complicate error reporting and consumption of the data.

And yes we do error rate comparisons before and after a schema change as a pre-production sign-off.

trink:
After 67MM docs tested less than 100 were impacted by this change.

templates/include/telemetry/environment.1.schema.json
```diff
@@ -5,7 +5,7 @@
       "properties": {
         "activeAddons": {
           "type": "object",
-          "additionalProperties": {
+          "properties": {
```
mreid-moz: I don't think this is right - this should be an `additionalProperties` field, but we should remove the "addon" piece. It should just be something like:
```
"additionalProperties": {
  "type": "object",
  "properties": { ... all the properties of an addon ...}
}
```

templates/include/telemetry/environment.1.schema.json
```diff
@@ -230,12 +228,12 @@
         ]
     },
     "experiments": {
+      "properties": {
```
mreid-moz: I believe this section was right before - we expect the structure:
```
"experiments": {
  "some_experiment_id": { "branch": "A" }
}
```
where `"some_experiment_id"` would be the additionalProperties, each with a structure with one property called `"branch"`.

tests/hindsight/jsonschema.4.json
```diff
@@ -0,0 +1,149 @@
+{
```
mreid-moz: Please add a note to the README (or somewhere) with the provenance of this schema.

templates/include/telemetry/environment.1.schema.json
```diff
@@ -451,90 +447,81 @@
...
+                    "type": "number"
                   }
                 }
               }
             },
             "features": {
               "type": "object",
-              "items": {
-                "feature": {
-                  "type": "object",
```
trink: Sigh, this section never matched reality https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/environment.html

trink:
@mreid-moz RFAL

templates/include/telemetry/environment.1.schema.json
```diff
@@ -420,17 +414,19 @@
...
-          },
-          "manufacturer": {
-            "type": "string"
-          },
-          "hardware": {
-            "type": "string"
-          },
-          "isTablet": {
-            "type": "boolean"
+          "properties": {
```
mreid-moz: The `"device"` section needs a `"type": "object"` declaration

PR #144, opened 2018-04-06
Remove obsole "childPayloads" from schema

badboy:
With bug 1443599 we stopped reporting "childPayloads" in pings.
https://bugzilla.mozilla.org/show_bug.cgi?id=1443599
Bug 1443600 removes all recording code.
https://bugzilla.mozilla.org/show_bug.cgi?id=1443600


badboy:
(Rebased on dev)

mreid-moz:
We can't safely remove these from the schema, since old Firefoxes will continue to send documents with a `childPayloads` section.

badboy:
@mreid-moz Ok, given that these are optional anyway, it will not cause any problems, correct?

mreid-moz:
Hm, it depends. If we're truly never using this data anymore, then it shouldn't cause any problems. If anything is still using the data from old versions, then it might matter - we are currently validating that `childPayloads` is an array, and if we remove it from the schema then we no longer check its data type.

badboy:
We already [fully removed the `childPayload` object from the ping](https://bugzilla.mozilla.org/show_bug.cgi?id=1443599) (not just sending an empty array). This already landed 3 weeks ago on Nightly.
I don't know where in the pipeline to look if anything depends on a valid array there.

mreid-moz:
I guess my point is that the schema is used to validate submissions from not only the latest version of Firefox, but also old versions that continue to send data. As such, the schema should be as complete as possible for historic versions as well. 

In other words, things become obsolete in terms of _validation_ in a different way than the client code producing the data - we should maintain backwards compatibility where possible.

So I think we should leave this section in the schema.

fbertsch:
This kind of change is the entire reason we have versioned schemas. Really we should be able to support historical pings with the old `childPayloads` and new pings without them, without cluttering the most recent schema.

@mreid-moz It seems to me we should consider supporting this kind of change better, for e.g. specific releases, some schema to use.

mreid-moz:
I filed [Bug 1453377](https://bugzilla.mozilla.org/show_bug.cgi?id=1453377) to review the versioning approach used by Fx desktop.

badboy:
So as I understand it we keep the schema as is for now as it still validates incoming pings from old browser versions. We have a couple of other objects in the ping that we stopped reporting (some report an empty object/array for now). I will check the schema for each of them and make sure they are at least optional.

This can be closed then.

mreid-moz:
@badboy Yup, that sounds right to me. Thanks!

PR #145, opened 2018-04-11
Add optional displayVersion field to environment schema

wlach:
This information is redundant, but we're adding the field for now to support
https://bugzilla.mozilla.org/show_bug.cgi?id=1452745. This just adds a little
bit of extra validation on the server side to make sure that the field
is properly specified if it is present (obviously older versions will not have
the field).


mreid-moz:
@wlach Just to be clear, this is expected to duplicate the `displayVersion` field in the [`application` section](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/templates/include/telemetry/application.1.schema.json#L25)?

wlach:
@mreid-moz yup

mreid-moz:
Ok, per IRC conversation, this is a workaround for pingsender, which can't see the `"application"` info, and will be addressed in [Bug 1453344](https://bugzilla.mozilla.org/show_bug.cgi?id=1453344).

mreid-moz:
I think `shield-icq-v1.4.schema.json` also needs to be updated.

For some reason the docker validation is failing for me locally against this code as well as the `dev` branch... not sure what's up with that.

mreid-moz:
Ahh, I had to run `docker build --no-cache -t mps` to pick up the latest hindsight packages.

validation/telemetry/main.4.displayversion.fail.json
```diff
@@ -0,0 +1,21 @@
...
+    "buildId": "20151103030248",
+    "channel": "beta",
+    "name": "Firefox",
+    "platformVersion": "45.0",
+    "vendor": "Mozilla",
+    "version": "45.0",
+    "xpcomAbi": "x86_64-gcc3"
+  },
+  "creationDate": "2015-11-05T01:25:43.312Z",
+  "environment": {
```
mreid-moz: There is a bit more "environment" stuff that's needed to otherwise-pass, should be something like:
```
  "environment": {
    "build": {
      "applicationId": "{ec8030f7-c20a-464f-9b0e-13a3a9e97384}",
      "applicationName": "Firefox",
      "architecture": "x86-64",
      "buildId": "20151103030248",
      "displayVersion": "baddisplayversion",
      "platformVersion": "45.0",
      "vendor": "Mozilla",
      "version": "45.0",
      "xpcomAbi": "x86_64-gcc3"
    },
    "partner": {},
    "settings": {},
    "system": {}
  },
```

wlach:
Thanks for the helpful feedback @mreid-moz, does it look ok now?

PR #146, opened 2018-04-17
Update crash ping schema

akkomar:
Add `metadata.StartupCrash` field, specify required fields.

### DIFF
--- a/schemas/telemetry/crash/crash.4.schema.json
+++ b/schemas/telemetry/crash/crash.4.schema.json
@@ -925,6 +925,9 @@
             "SecondsSinceLastCrash": {
               "type": "string"
             },
+            "StartupCrash": {
+              "type": "string"
+            },
             "SystemMemoryUsePercentage": {
               "type": "string"
             },
@@ -949,6 +952,10 @@
           "type": "integer"
         }
       },
+      "required": [
+        "crashDate",
+        "metadata"
+      ],
       "type": "object"
     },
     "type": {
@@ -963,6 +970,14 @@
       "type": "number"
     }
   },
+  "required": [
+    "type",
+    "id",
+    "creationDate",
+    "version",
+    "application",
+    "payload"
+  ],
   "title": "crash",
   "type": "object"
 }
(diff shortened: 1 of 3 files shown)


mreid-moz:
@acmiyaguchi are you able to run a test of this change?

acmiyaguchi:
@mreid-moz I'll run a test of this today, the work I've done on the schema-validator should be able to handle this case.

Here's the steps I'll run through once I have a machine up:

```
$ git clone https://github.com/acmiyaguchi/schema-validator
$ cd validator

$ pip install pipenv
$ pipenv run python setup.py bdist_egg

$ spark-submit \
    --master yarn \
    --deploy-mode client \
    --py-files dist/*.egg \
    --files ${SCHEMA}.json
    --protocol s3
    driver.py SCHEMA INPUT_PATH OUTPUT_PATH
```

acmiyaguchi:
Here's a fairly lengthy script that runs the validation: https://gist.github.com/acmiyaguchi/16bd0bb8c53804afcaedacd677221fec

In short, on a 1% sample of crash pings for a day, the errors are the same for master and this PR.

akkomar:
Thanks @acmiyaguchi , that's really useful!
However, how is it possible that there are parse errors for master? Isn't data you read via `Dataset` api already validated?

mreid-moz:
Processing data from `Dataset` works, but only when making the schema more strict.

mreid-moz:
@fbertsch not yet - at least not in this example. Reading from landfill (pre-validation) data is up next.

acmiyaguchi:
@akkomar I ran the job for 20180401, so there could have been changes to the schema in the meanwhile. Here's the single non-null validation output (under the `/output/validation` directory as parquet).[1]

```
path: environment/addons/activeAddons/jetpack-extension@dashlane.com/version
schema_path: properties/environment/properties/addons/properties/activeAddons/additionalProperties/properties/version/type
```

@fbertsch There are a couple changes that have to be made to `python_moztelemetry` before we can read from landfill. See bug 1447851 for more details.[2]  I have administrative access to launch a databricks cluster with access to landfill that can be used with a patched version of moztelemetry, I haven't figured out the best way to lay out the data yet though.

[1] https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/11057
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1447851

PR #147, opened 2018-04-18
Bug 1414352 - Add mobile metrics ping

fbertsch:
* Add JSON Schema for mobile metrics ping
* Add createdDate and createdTimestamp for mobile events pings
* Add experiments for mobile events pings


fbertsch:
@mreid-moz note that the new fields added to existing pings (`mobile-event` and `focus-event`) are a bit tighter bounds than what we currently accept, but none of them are required so the effect should be largely nil on validation.

templates/include/telemetry/metricsData.1.schema.json
```diff
@@ -0,0 +1,21 @@
+"scalars": @TELEMETRY_SCALARS_1_JSON@,
+"keyedScalars": {
+  "type": "object",
+  "additionalProperties": {
+    "keyedScalar": @TELEMETRY_SCALARS_1_JSON@
```
mreid-moz: This section should not have a name, and should just look like `"additionalProperties":  @TELEMETRY_SCALARS_1_JSON@`

fbertsch:
@mreid-moz good catch, that changes was left in after I rebased from `master` to `dev`. RFAL

PR #148, opened 2018-04-25
Bug 1455385 - Move mobile-event 'os' to partitions

relud:
co-dependent on mozilla-services/puppet-config#2725

### DIFF
--- a/schemas/telemetry/mobile-event/mobile-event.1.parquetmr.txt
+++ b/schemas/telemetry/mobile-event/mobile-event.1.parquetmr.txt
@@ -15,7 +15,6 @@ message mobile_event {
   required binary clientId (UTF8);
   required int64 seq;
   required binary locale (UTF8);
-  required binary os (UTF8);
   required binary osversion (UTF8);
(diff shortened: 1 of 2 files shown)


relud:
deploying this should wait until I have finished creating a backfill, to minimized downtime of the data set

templates/include/telemetry/mobileCommon.1.schema.json
```diff
@@ -11,7 +11,8 @@
   "type": "string"
 },
 "os": {
-  "type": "string"
+  "type": "string",
+  "enum": [ "Android", "iOS", "iPhone OS" ]
```
relud: we need to validate os if we are using it as a partition, this is a temporary solution until Bug 1455381 is implemented

cc @fbertsch this change effects focus-event and mobile-metrics. I was able to confirm that this is all `os` values found in `focus_event_longitudinal`, but I couldn't find a mobile-metrics dataset to check.

relud: nevermind, doing this differently

PR #149, opened 2018-04-25
Bug 1455383 - Change core ping partitions

relud:
co-dependent on mozilla-services/puppet-config#2726

### DIFF
--- a/schemas/telemetry/core/core.9.parquetmr.txt
+++ b/schemas/telemetry/core/core.9.parquetmr.txt
@@ -6,6 +6,7 @@ message core {
     optional binary geoCountry (UTF8);
     optional binary geoCity (UTF8);
     optional binary appBuildId (UTF8);
+    optional binary normalizedChannel (UTF8);
   }
   optional group application {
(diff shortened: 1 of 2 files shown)


relud:
deploying this should wait until I have finished creating a backfill, to minimized downtime of the data set

templates/include/telemetry/coreCommon.1.schema.json
```diff
@@ -6,7 +6,8 @@
   "type": "string"
 },
 "os": {
-  "type": "string"
+  "type": "string",
+  "enum": [ "Android", "iOS", "iPhone OS" ]
```
relud: we need to validate `os` if we are using it as a partition, this is a temporary solution until [Bug 1455381](https://bugzilla.mozilla.org/show_bug.cgi?id=1455381) is implemented

whd: Instead of dropping these entirely when os isn't an expected value I think we should probably just change the parquet output message matcher to match only the specific os values we care about (for now).

PR #151, opened 2018-04-27
Add processType to crash ping schema

akkomar:


templates/telemetry/crash/crash.4.schema.json
```diff
@@ -70,6 +70,10 @@
             }
           }
         },
+        "processType": {
+          "type": "string",
+          "enum": [ "main", "content", "gpu" ]
```
mreid-moz: Do we want the ping to fail validation if we introduce a new `processType`?  If there's no strong reason to make this an `enum`, I would somewhat prefer to leave this as a free-form string value.

akkomar: Good point, I'll leave it as free-form string.

georgf: +1, a free-form string value would be more flexible in the future.
Is there any potential down-side to allowing free-form values, e.g. would it affect analysis in some way if unexpected values show up?

mreid-moz: I checked some of the existing data and it is well-behaved (one of these three values or null), so I'm more worried about intentionally-introduced new values (maybe `dynamic`?) than a bunch of unexpected values.

If we start to observe unexpected values here it seems like we should address it at that time.

akkomar: But how would you know you're getting unexpected values if it's not validated?
On the other hand I totally agree with the flexibility point.

mreid-moz: You'd start to see them appear in datasets / analyses downstream.

fbertsch: I'm kind of the opinion that we should have just included all current known process types [0]. But as per usual I'm late to the party :)

[0] https://searchfox.org/mozilla-central/source/toolkit/components/crashes/nsICrashService.idl#24

mreid-moz: I'm not confident enough in our alerting on validation failures yet, so I'm inclined to be permissive / future-proof where possible.

fbertsch: @mreid-moz that is concerning, can you make a bug to track that?

fbertsch: I think we should prioritize that bug highly so we don't have to be concerned with missing data, and ending up with too lenient schemas.

PR #154, opened 2018-05-09
Fix hgpush schema path

whd:
We're supposed to support flattened schemas but seemingly don't. For now we should move them to conform with the infra's expectations.


trink:
We only support them for pioneer atm, the CI tests were relaxed to allow it but were not restricted specifically to pioneer so they allowed this case to pass.  +1 on allowing both forms directory/flattened everywhere (i.e. if you have many versions you can opt to create a sub directory)

PR #155, opened 2018-05-11
Bug 1454739 - Add geo_subdivision to telemetry_new_profile_parquet

relud:


relud:
needs a version bump and a backfill, but it won't break any queries to automatically switch from v1 to v2

PR #156, opened 2018-05-16
Add optional displayVersion field to core ping

wlach:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1461772


wlach:
Done, also added a test which should fail.

PR #158, opened 2018-05-17
Update compiled new-profile parquet schema

relud:
#155 only updated the template, and not the compiled version


relud:
also worth noting that while I don't understand _how_, this did not prevent the new dataset from getting the new fields.

trink:
The schema directory is just for reference in the repo.  When a new schema package is created it is automatically generated, so that is why it works.  https://github.com/mozilla-services/mozilla-pipeline-schemas#notes explains why we check in the artifacts.

PR #160, opened 2018-05-23
Add downgrade ping schemas and validation test.

Mossop:
Adds the downgrade ping for bug 1463198. This will be submitted by the pingsender.

A couple of things I'm not sure of. The ping filename suggests version 1, should I have used 4 like most of the other pings? The version in the ping schema is 4, should I have matched the version from the filename?

The test is generated by my device from the code in bug 1463198.


mreid-moz:
> The version in the ping schema is 4, should I have matched the version from the filename?

Yes, the filenames should be changed to use a version of `4` here to match the `"version"` field inside the document. For Firefox telemetry pings, that field is currently used by the [ingestion code](https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/moz_telemetry/io_modules/decoders/moz_ingest/telemetry.lua#L202) to determine which schema version to apply.

templates/telemetry/downgrade/downgrade.4.parquetmr.txt
```diff
@@ -0,0 +1,14 @@
+message downgrade {
+  required binary id (UTF8);
+  required binary creationDate (UTF8);
+  @TELEMETRY_APPLICATION_1_TXT@
+  required binary clientId (UTF8);
+  required double version;
+  required group payload {
+    required binary lastVersion (UTF8);
```
mreid-moz: We generally recommend making fields `optional` unless the record is completely useless without them - that way if they are removed in later versions it won't cause trouble with the data storage.

mreid-moz:
cc @whd for a deploy (thanks!)

PR #167, opened 2018-06-05
Make shield-study Date header metadata optional

whd:
This is optional everywhere else and we see occasionally clients sending data without this header.


whd:
I checked and we're all over the board with those fields. We should probably standardize, but because those fields are added by the ingestion infrastructure they are guaranteed to be there, so having them be optional or required is fine.

mreid-moz:
Sounds good. Let's look at standardizing them as we migrate to using top-level metadata fields (versus nested).

PR #169, opened 2018-06-13
Bug 1468498 - add landingSystem to hgpush events

mars-f:
Add an optional 'landingSystem' field to hgpush events so that we can
track commits that came through Lando, our automated code landing system.


mars-f:
The data review request for this change is in progress: [Bug 1468339](https://bugzilla.mozilla.org/show_bug.cgi?id=1468339)

mars-f:
@mreid-moz I've removed the v2 schema and applied the changes to v1.  The data collection review is complete, too, so it's OK to deploy after landing.

PR #175, opened 2018-06-26
Add study_variation, frecency_scores and four new weights to frecency-update

florian:
As discussed with Sunah / Anthony, this PR adds four new weights and a new field for treatment / control, frecency scores to the frecency-update ping.


schemas/telemetry/frecency-update/frecency-update.1.schema.json
```diff
@@ -25,12 +25,15 @@
...
+          "type": "string"
+        },
         "update": {
           "items": {
             "type": "number"
           },
-          "maxItems": 18,
-          "minItems": 18,
+          "maxItems": 22,
+          "minItems": 22,
```
mreid-moz: If there are any messages still being sent with the current iteration of the schema, this change (and the one making the new `study_variation` field required) will cause them to fail validation - is that the desired effect?

schemas/telemetry/frecency-update/frecency-update.1.schema.parquetmr.txt
```diff
@@ -7,6 +7,7 @@ message frecency-update {
     required int32 num_chars_typed;
     required int32 rank_selected;
     required int32 num_suggestions_displayed;
+    required binary study_variation;
```
mreid-moz: This field should be added as `optional`, since there will be historic data where it is missing.

schemas/telemetry/frecency-update/frecency-update.1.schema.json
florian: All the messages sent so far were just from me debugging and testing my addon, so we should probably delete those messages anyways. All the new pings I'm sending will follow the new format.

schemas/telemetry/frecency-update/frecency-update.1.schema.parquetmr.txt
florian: See the other comment regarding this.

florian:
I've updated the file names. Just out of interest, what exactly does `version` refer to here? Is it the JSON schema version?
The README doesn't mention explicitly mention what version is about, so I just assumed that it was the version of the respective ping.

mreid-moz:
It is the version of the ping, however for pings sent by Firefox, it is currently based on the version of the ["common" ping format](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/common-ping.html), which is `4` for everything :-/

florian:
@mreid-moz Okay, that's good to know. I updated the PR. Can you take another look at it?

PR #176, opened 2018-06-28
Add a couple more histogram validation examples.

mreid-moz:
Add a minimal passing histogram, a failure due to a negative count, and simplify the failure case for an invalid bucket key pattern.


mreid-moz:
Note this came out of investigating [Bug 1471680](https://bugzilla.mozilla.org/show_bug.cgi?id=1471680)

PR #179, opened 2018-07-10
Automatically remove the test container when it exits

badboy:
The created container is only used for a single test run and never reused. It however persists on the user's machine until its deleted. Let's just remove it immediately after running the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ To run the tests:
     docker build -t mps .
 
     # run the tests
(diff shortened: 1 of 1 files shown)


acmiyaguchi:
This looks good to me. I'd been using `docker ps` to keep track of running container state, I didn't realize that the final container state is stored on disk for reference. I used `docker container list --all` to double check.

@badboy Thanks!

PR #182, opened 2018-07-19
Add Timestamp to frecency-update

florian:
This PR adds the `Timestamp` field to the Parquet schema, for easier analysis after the study ends.

### DIFF
--- a/schemas/telemetry/frecency-update/frecency-update.4.schema.parquetmr.txt
+++ b/schemas/telemetry/frecency-update/frecency-update.4.schema.parquetmr.txt
@@ -2,6 +2,7 @@ message frecency-update {
   required binary clientId (UTF8);
   required binary type (UTF8);
   required group payload {
+    required long Timestamp;
     required group frecency_scores (LIST) {
       repeated group list {
         required float element;
@@ -13,7 +14,6 @@ message frecency-update {
     required int32 rank_selected;
     required int32 num_suggestions_displayed;
     required binary study_variation;
-    required long Timestamp;
     required group update (LIST) {
       repeated group list {
         required float element;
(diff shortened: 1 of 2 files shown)


schemas/telemetry/frecency-update/frecency-update.4.schema.parquetmr.txt
```diff
@@ -13,6 +13,7 @@ message frecency-update {
     required int32 rank_selected;
     required int32 num_suggestions_displayed;
     required binary study_variation;
+    required int32 Timestamp;
```
sunahsuh: This is a long and needs to be int64. I'd also put it as a the first field, before clientId

florian: I changed it to a long now. In the last review, Anthony told me to order the fields alphabetical. Is Timestamp an exception to this or should I leave it completely alphabetically sorted?

sunahsuh: For parquet? I suspect he meant for the json schema, since a json object isn't ordered, but the parquet schema will determine the actual column order that's output

florian: Oh, I wasn't aware of that. `Timestamp` is the first element in `payload` now.

schemas/telemetry/frecency-update/frecency-update.4.schema.parquetmr.txt
```diff
@@ -2,6 +2,7 @@ message frecency-update {
   required binary clientId (UTF8);
   required binary type (UTF8);
   required group payload {
+    required long Timestamp;
```
sunahsuh: This is still in the payload nested field

This should be at the top level, and after chatting with whd, because this is heka metadata we need to append the field name with "metadata" (i.e. metadata_Timestamp)

florian: Like this?

florian:
Thanks for looking over this.
That took more iterations than expected :)

PR #183, opened 2018-07-19
Use int64 for frecency-update Timestamp field

sunahsuh:


whd:
This has been deployed.

PR #186, opened 2018-08-02
Update environment

mreid-moz:
In [Bug 1476630](https://bugzilla.mozilla.org/show_bug.cgi?id=1476630) I took a look at the Telemetry environment to see if we're missing anything in the schema and found a few things that were missing.

For review purposes, it may be easiest to look at the commits individually.

One thing I am not sure about is whether we can receive `null` values for any of the newly added fields.

### DIFF
--- a/schemas/telemetry/new-profile/new-profile.4.schema.json
+++ b/schemas/telemetry/new-profile/new-profile.4.schema.json
@@ -295,6 +295,9 @@
               "pattern": "^[0-9]{2,3}\\.",
               "type": "string"
             },
+            "updaterAvailable": {
+              "type": "boolean"
+            },
             "vendor": {
               "type": "string"
             },
@@ -383,6 +386,9 @@
         },
         "settings": {
           "properties": {
+            "addonCompatibilityCheckEnabled": {
+              "type": "boolean"
+            },
             "attribution": {
               "properties": {
                 "campaign": {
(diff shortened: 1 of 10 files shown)


schemas/telemetry/new-profile/new-profile.4.schema.json
```diff
@@ -875,6 +893,12 @@
                 }
               },
               "type": "object"
+            },
+            "virtualMaxMB": {
```
mreid-moz: This field does need to allow nulls, at least based on the `validation/telemetry/new-profile.4.sample.pass.json` example.

schemas/telemetry/new-profile/new-profile.4.schema.json
```diff
@@ -477,6 +486,12 @@
         },
         "system": {
           "properties": {
+            "appleModelId": {
```
mreid-moz: This field says it can be null in [the docs](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/environment.html)

templates/include/telemetry/environment.1.schema.json
```diff
@@ -336,6 +342,9 @@
             }
           }
         },
+        "searchCohort": {
+          "type": "string"
```
acmiyaguchi: From the data, it looks like this value should be nullable.

mreid-moz:
Looks like `architecturesInBinary` was removed [here](https://hg.mozilla.org/mozilla-central/rev/1045a32c6e18). That's quite recent (June 7th), so we should probably keep this field in the schema.

mreid-moz:
Thanks for the review! That notebook for checking the presence of nulls is super helpful!

PR #187, opened 2018-08-02
block-autoplay JSON schema and validation tests.

alastor0325:
We would send the custom telemetry ping from block-autoplay shield-study extension [1] which is used to analysis the user behavior when they interact with the block-autoplay prompt and setting.

This custom ping also pass the data review [2].

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1475099
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1479589

### DIFF
--- a/schemas/telemetry/block-autoplay/block-autoplay.1.schema.json
+++ b/schemas/telemetry/block-autoplay/block-autoplay.1.schema.json
@@ -1,0 +1,280 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "properties": {
+    "application": {
+      "additionalProperties": false,
+      "properties": {
+        "architecture": {
+          "type": "string"
+        },
+        "buildId": {
+          "pattern": "^[0-9]{10}",
+          "type": "string"
+        },
+        "channel": {
+          "type": "string"
+        },
+        "displayVersion": {
+          "pattern": "^[0-9]{2,3}\\.",
+          "type": "string"
+        },
+        "name": {
+          "type": "string"
+        },
+        "platformVersion": {
+          "pattern": "^[0-9]{2,3}\\.",
+          "type": "string"
+        },
+        "vendor": {
+          "type": "string"
+        },
+        "version": {
+          "pattern": "^[0-9]{2,3}\\.",
+          "type": "string"
+        },
+        "xpcomAbi": {
+          "type": "string"
+        }
+      },
+      "required": [
+        "architecture",
+        "buildId",
+        "channel",
+        "name",
+        "platformVersion",
+        "version",
+        "vendor",
+        "xpcomAbi"
+      ],
+      "type": "object"
+    },
+    "clientId": {
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
+      "type": "string"
+    },
+    "creationDate": {
+      "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]{3}Z$",
+      "type": "string"
+    },
+    "id": {
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
+      "type": "string"
+    },
+    "payload": {
+      "additionalProperties": false,
+      "properties": {
+        "branch": {
+          "enum": [
+            "control",
+            "allow-and-notRemember",
+            "deny-and-notRemember",
+            "allow-and-remember",
+            "deny-and-remember"
+          ],
+          "type": "string"
+        },
+        "id": {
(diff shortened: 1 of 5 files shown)


alastor0325:
Hi, @trink,
Could you help me review my patches? 
Thank you!

templates/telemetry/block-autoplay/block-autoplay.1.schema.json
```diff
@@ -0,0 +1,140 @@
...
+			"type": "string",
+			"enum": [
+        "control",
+        "allow-and-notRemember",
+        "deny-and-notRemember",
+        "allow-and-remember",
+        "deny-and-remember"
+      ]
+		},
+		"type": {
```
trink: Since the following objects are conditional on type and mutually exclusive the schema should reflect that.  e.g. currently a type of counts isn't required to have a counters object (see oneOf https://json-schema.org/understanding-json-schema/reference/combining.html#oneof)

templates/telemetry/block-autoplay/block-autoplay.1.schema.json
```diff
@@ -0,0 +1,140 @@
...
+		}
+	},
+	"additionalProperties": false,
+	"required": [
+		"id",
+		"clientId",
+		"branch",
+		"type"
+	]
+}
```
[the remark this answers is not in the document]
alastor0325: thanks!

templates/telemetry/block-autoplay/block-autoplay.1.schema.json
alastor0325: OK, will do.

templates/telemetry/block-autoplay/block-autoplay.1.schema.json
```diff
@@ -0,0 +1,163 @@
...
+			"type": "string",
+			"enum": [
+        "control",
+        "allow-and-notRemember",
+        "deny-and-notRemember",
+        "allow-and-remember",
+        "deny-and-remember"
+      ]
+		},
+		"type": {
```
trink: This still allows a type of "counts" with an array of "settingsChanged".

trink: > I still can't understand how to use "oneOf" to make the types mapping to their correspond objects

Yeah, it is very difficult to create a 100% accurate schema for this JSON.  If the object was always named the same i.e. 'payload' then oneOf is easier to apply (in that case type would be in the payload).  Without restructuring the data you would have to allow additional properties in which case any of the objects could exist; you could verify that if type == 'counts' a counters object exists but nothing would prevent you from also having a settings object and passing validation.

alastor0325:
Based on the offline dicussion with trink, we decide to add new attribute `payload` which could carry different contents but one at a time, such as `counters`, `settingsChanged` or `promptRespond`.

jtg567:
do these pings conform to the schema for `shield_study_addon` docType or will the pipeline file these under docType "OTHER"?

alastor0325:
No, we create a new format which doctype is `block-autoplay`.

schemas/telemetry/block-autoplay/block-autoplay.1.schema.json
```diff
@@ -0,0 +1,201 @@
...
+                }
+              },
+              "required": [
+                "totalPages",
+                "totalPagesAM",
+                "totalBlockedAudibleMedia"
+              ],
+              "type": "object"
+            },
+            "type": {
```
trink: As we discussed `type` is unnecessary/redundant but was requested to remain part of the schema.

mreid-moz:
It looks from [here](https://github.com/alastor0325/autoplay-shield-study/pull/3/files#diff-0ef0b65fe461b9218696c92f6fd68e72R128) like the pings will be submitted using `TelemetryController.submitExternalPing`, which wraps the given payload in the [Common Ping format](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/common-ping.html#common-ping-format).

That implies a couple of changes:
- `clientId` can be removed in the payload - it's part of the Telemetry environment, which is included [here](https://github.com/alastor0325/autoplay-shield-study/pull/3/files#diff-0ef0b65fe461b9218696c92f6fd68e72R127).
- the schema should be structured more like the [crash ping](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/templates/telemetry/crash/crash.4.schema.json), where the common ping format is used for the outer part and the `payload` field contains your custom ping structure.

You should be able to confirm the overall document structure by looking in `about:telemetry`, finding one of these pings locally, then clicking the "Raw JSON" link in the bottom left.

alastor0325:
@mreid-moz Hi, Mark, I've modified my patches, could you help me review them again? Thank you!

trink:
Technically it is fine.  However, it is quite a large ping now.  Why is the environment being include? It adds a lot of bloat and clearly wasn't a requirement in the first version of the ping.  If there are plans to use it r+ if not r- as it can easily be disabled.

trink:
It also needs to be disabled here https://github.com/alastor0325/autoplay-shield-study/pull/3/files#diff-0ef0b65fe461b9218696c92f6fd68e72R127

alastor0325:
Thank you for the reminding, I've already modified that.

PR #193, opened 2018-08-31
Bug 1458734 - Test schemas against real sampled data using edge-validator

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1458734#c1)

This PR tests schemas against real, sampled data. A report is generated between the `HEAD` of the PR and `mozilla:dev`. The sampled data (`sanitized-landfill-sample`) is generated by a scheduled batch job under a stricter IAM role with access to landfill data. The CI takes advantage of a script included in the `mozilla/edge-validator` docker image to generate integration reports and diffs between two revisions of mozilla-pipeline-schemas. If there is a change in the report (e.g. a decrease in validation errors on a doctype), a diff is generated and added as a test artifact that can be reviewed.

This moves from TravisCI to CircleCI to support more complex workflows. I wrote some test cases in [this PR on my local repository](https://github.com/acmiyaguchi/mozilla-pipeline-schemas/pull/1). I've copied and pasted the diffs into a comment. They can be retrieved clicking on the CI checkbox > integration` details` link > `Artifacts` tab > `app/test-reports` tree. 
image

The `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` need to be set in CircleCI in order to fetch the samples from S3. The samples do not contain any sensitive information and are limited to 1000 documents per doctype a day. The resulting report data only contains information about the namespace, doctype, version, and validation exceptions. I've been able to run CircleCI images with ssh enabled to debug, which means that the scope of the keys should probably be limited in scope. I think CircleCI limits access to the ssh functionality through Github teams though. 

The report takes between 3-6 minutes to run a report depending on the day. I would like to decrease the time by generating a gzip'ed archive for each submission date, which are improvements that need to be made against the landfill sampler and edge-validator image.

I've also included this mermaid graph for a high level overview of the feedback loop I want to introduce for schema development. 
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,20 @@
+version: 2
+jobs:
+  test:
+    docker:
+      - image: docker:18.02.0-ce-git
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Build Docker image
+          command: docker build -t mps .
+      - run:
+          name: Test Code
+          command: docker run mps
+
+workflows:
+    version: 2
+    build:
+        jobs:
+            - test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 version: 2
 jobs:
   test:
     docker:
-      - image: docker:18.02.0-ce-git
+      - image: docker:stable-git
     steps:
       - checkout
       - setup_remote_docker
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,56 @@
...
+      - setup_remote_docker
+      - run:
+          name: Build Docker image
+          command: docker build -t mps .
+      - run:
+          name: Test Code
+          command: docker run mps
+  integrate:
+    docker:
+      - image: mozilla/edge-validator:latest
```
whd: Should we use a stable tag?

acmiyaguchi: Not a bad idea, considering some of the potential changes.

acmiyaguchi:
It looks like Circle won't report on the PR until the config is set in the `dev`/`master` branch.

Thanks @whd for reviewing many of the pieces needed to make this happen.

PR #195, opened 2018-09-05
Mobile experiments

fbertsch:


fbertsch:
@trink can you to review the CMakeLists.txt change? @akkomar these are some changes to support the GeckoView rollout experiments (and future experiments - as discussed in data pod today).

schemas/telemetry/focus-event/focus-event.1.parquetmr.txt
```diff
@@ -8,26 +8,26 @@ message focus_event {
...
-  required int64 created;
-  optional int64 processStartTimestamp;
-  optional int64 tz;
-  required group settings (MAP) {
-    repeated group key_value {
-      required binary key (UTF8);
-      optional binary value (UTF8);
-    }
+  required int64 v;
+required binary clientId (UTF8);
```
akkomar: Indentation got broken.

schemas/telemetry/mobile-event/mobile-event.1.parquetmr.txt
```diff
@@ -12,24 +12,24 @@ message mobile_event {
...
-  optional binary arch (UTF8);
-  required int64 created;
-  required int64 processStartTimestamp;
-  optional int64 tz;
-  required group settings (MAP) {
-    repeated group key_value {
-      required binary key (UTF8);
-      optional binary value (UTF8);
-    }
+required binary clientId (UTF8);
```
akkomar: Indentation got broken.

CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ endforeach()
 
 file(GLOB_RECURSE SCHEMAS LIST_DIRECTORIES FALSE RELATIVE "${CMAKE_SOURCE_DIR}/templates" "${CMAKE_SOURCE_DIR}/templates/*")
 foreach(ITEM IN ITEMS ${SCHEMAS})
-  if(NOT ITEM MATCHES "^include/" )
+  if(NOT ITEM MATCHES "^(include/|.*/\\.[^/])" )
```
whd: Why is this change necessary? There shouldn't be any hidden files. Is this for local development where there might be editor temp files or something?

fbertsch: Exactly - for example I usually run tests while editing the files in another tab, which fails without this.

whd: Wouldn't you want to have saved your edited file *before* running the test? That aside, this does not seem to me to be an unreasonable change to make, so LGTM.

fbertsch: Well yeah, but I can save while still having them open :) saves me an extra `nvim` cmd

whd: I'll chalk this up as an emacs vs. vim thing, since the scenario you describe makes no sense to me ;)

PR #196, opened 2018-09-05
Use `[0-9]` instead of `\d` in histograms

acmiyaguchi:
This uses the more strict subset of regex that's recommended in [Understanding JSON Schema](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html) tutorial.

The python wrapper for rapidjson doesn't include the `std::regex` at compile time to support `\d`, but square brackets suffice. This should make the schemas more widely compatible and will fix the validation errors in the `edge-validator` server.

### DIFF
--- a/schemas/telemetry/main/main.4.schema.json
+++ b/schemas/telemetry/main/main.4.schema.json
@@ -986,7 +986,7 @@
               "values": {
                 "additionalProperties": false,
                 "patternProperties": {
-                  "^\\d+$": {
+                  "^[0-9]+$": {
                     "minimum": 0,
                     "type": "integer"
                   }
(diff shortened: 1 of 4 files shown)


acmiyaguchi:
To save you some clicks, here's the link to the diff: https://7-51375003-gh.circle-artifacts.com/0/app/test-results/4b86be2-ad6a1e1.diff

```diff
--- 
+++ 
@@ -54,7 +54,7 @@
         "error_rate": 0.0
     },
     "telemetry.main.4": {
-        "error_rate": 100.0
+        "error_rate": 0.0
     },
     "telemetry.malware-addon-states.4": {
         "error_rate": 100.0
@@ -78,7 +78,7 @@
         "error_rate": 100.0
     },
     "telemetry.saved-session.4": {
-        "error_rate": 100.0
+        "error_rate": 0.2
     },
     "telemetry.searchvol.4": {
         "error_rate": 100.0
```

whd:
I assume this doesn't have any material performance implication on the ingestion side. @trink can you verify that? Otherwise LGTM.

trink:
Technically this change should be faster as it matches a smaller set but I have never profiled it.

PR #197, opened 2018-09-06
Remove focus-event trailing comma

whd:
https://bugzilla.mozilla.org/show_bug.cgi?id=1383111

The tests don't fail because they regenerate the schemas from templates before running.


fbertsch:
Ah, I see that's what the bug is about. Thanks for fixing this!

PR #201, opened 2018-09-12
Package as jar

jklukas:
First piece of https://github.com/mozilla/gcp-ingestion/issues/56

Publishing won't work correctly until we get AWS creds uploaded to CircleCI that allow publishing an artifact to S3 from this project.


jklukas:
I'm beginning to have some second thoughts here. We currently only need this for one project, and it would likely be less code and more straight forward to provide a shell script in that project that pulls down this repo and copies the contents directly into `src/main/resources`.

PR #207, opened 2018-10-29
Bug 1502439: Add bug_1501329_affected to core ping

fbertsch:


schemas/coverage/coverage/coverage.1.schema.json
```diff
@@ -16,8 +16,11 @@
       "type": "string"
     },
     "telemetryEnabled": {
-      "type": "integer",
-      "enum": [0, 1]
+      "enum": [
```
fbertsch: Did our formatting change? I can  remove this if needed.

mreid-moz: Either the pretty-printing changed a bit or just didn't get checked in last time. There's [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1383111) on file to enforce consistency here.

Let's keep this fix, thanks!

templates/include/telemetry/coreCommon.2.schema.json
```diff
@@ -47,4 +47,7 @@
   "items": {
     "type": "string"
   }
+},
+"bug_1501329_affected": {
```
mreid-moz: Do we want to add this to all versions of the core ping? Or just the latest?

fbertsch: We'll do all

PR #208, opened 2018-10-29
Update the location of sampled data

acmiyaguchi:


whd:
@acmiyaguchi yes. I have updated the permissions to point at the new bucket.

PR #213, opened 2018-11-05
Store events as we do on desktop

mdboom:
This changes how events are stored in the glean ping to follow [how they are on desktop](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html#serialization-format)

### DIFF
--- a/schemas/glean/baseline/baseline.1.schema.json
+++ b/schemas/glean/baseline/baseline.1.schema.json
@@ -96,41 +96,45 @@
       ]
     },
     "events": {
-      "allOf": [
-        {
-          "$ref": "#/definitions/base"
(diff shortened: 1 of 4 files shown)


fbertsch:
@mdboom any way we can use the existing json template, with possible revisions? If this is backwards compatible with the desktop definition, that is. https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/templates/include/common/event.1.schema.json

mdboom:
@fbertsch: This now reuses the event template used by everything else.  It's a bit less strict than what was here, but tightening it could be done as a follow-on if necessary.

PR #215, opened 2018-11-07
Bug 1435827: Adding JSON and Parquet schemas

CarlCorcoran:
This adds the necessary validation and ingestion schemas for the new "Untrusted Modules" ping introduced in Bug 1435827, along with basic test payloads.

The schema includes the telemetry ping environment block, which is essentially a port of /templates/include/telemetry/environment.1.schema.json, minus the following structures because of the difficulty in representing in parquet:
  - userPrefs
  - activeAddons/userDisabled
  - activeAddons/foreignInstall
  - activeGMPlugins/applyBackgroundUpdates
  - theme/foreignInstall
  - activeAddons/version
  - system/os/version
  - experiments

### DIFF
--- a/schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
+++ b/schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
@@ -174,10 +174,10 @@ message untrusted-modules {
         }
       }
     }
-    required group addons {
+    optional group addons {
       optional binary persona (UTF8);
-      required group theme {
-        required binary id (UTF8);
+      optional group theme {
+        optional binary id (UTF8);
         optional boolean blocklisted;
         optional binary description (UTF8);
         optional binary name (UTF8);
@@ -189,44 +189,44 @@ message untrusted-modules {
         optional int64 installDay (UINT_64);
         optional int64 updateDay (UINT_64);
       }
-      required group activeAddons (MAP) {
+      optional group activeAddons (MAP) {
         repeated group key_value {
           required binary key (UTF8);
           required group value {
-            required boolean blocklisted;
-            required binary name (UTF8);
+            optional boolean blocklisted;
+            optional binary name (UTF8);
             optional binary description (UTF8);
-            required boolean appDisabled;
-            required int64 scope;
-            required binary type (UTF8);
-            required boolean hasBinaryComponents;
-            required int64 installDay (UINT_64);
-            required int64 updateDay (UINT_64);
+            optional boolean appDisabled;
+            optional int64 scope;
+            optional binary type (UTF8);
+            optional boolean hasBinaryComponents;
+            optional int64 installDay (UINT_64);
+            optional int64 updateDay (UINT_64);
             optional int64 signedState;
-            required boolean isSystem;
+            optional boolean isSystem;
           }
         }
       }
-      required group activeGMPlugins (MAP) {
+      optional group activeGMPlugins (MAP) {
         repeated group key_value {
           required binary key (UTF8);
           required group value {
-            required binary version (UTF8);
-            required boolean userDisabled;
+            optional binary version (UTF8);
+            optional boolean userDisabled;
           }
         }
       }
-      required group activePlugins (LIST) {
+      optional group activePlugins (LIST) {
         repeated group list {
           required group element {
-            required binary name (UTF8);
-            required binary version (UTF8);
-            required binary description (UTF8);
-            required boolean blocklisted;
-            required boolean disabled;
-            required boolean clicktoplay;
-            required int64 updateDay;
-            required group mimeTypes (LIST) {
+            optional binary name (UTF8);
+            optional binary version (UTF8);
+            optional binary description (UTF8);
+            optional boolean blocklisted;
+            optional boolean disabled;
+            optional boolean clicktoplay;
+            optional int64 updateDay;
+            optional group mimeTypes (LIST) {
               repeated group list {
                 required binary element (UTF8);
               }
--- a/schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
+++ b/schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
@@ -1,0 +1,313 @@
+message untrusted-modules {
+  optional group environment {
+    optional group profile {
+      optional int64 creationDate;
+      optional int64 resetDate;
+    }
+    required group settings {
+      optional boolean isDefaultBrowser;
+      optional boolean addonCompatibilityCheckEnabled;
+      optional binary locale (UTF8);
+      optional boolean e10sEnabled;
+      optional boolean blocklistEnabled;
+      optional binary defaultSearchEngine (UTF8);
+      optional boolean telemetryEnabled;
+      optional binary e10sCohort (UTF8);
+      optional binary searchCohort (UTF8);
+      optional int64 e10sMultiProcesses;
+      optional group defaultSearchEngineData {
+        optional binary name (UTF8);
+        optional binary loadPath (UTF8);
+        optional binary submissionURL (UTF8);
+        optional binary origin (UTF8);
+      }
+      optional group update {
+        optional boolean autoDownload;
+        optional binary channel (UTF8);
+        optional boolean enabled;
+      }
+      optional group sandbox {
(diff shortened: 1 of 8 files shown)


templates/telemetry/untrusted-modules/untrusted-modules.4.schema.json
```diff
@@ -0,0 +1,154 @@
...
+          "properties": {
+            "memoryMap": {
+              "type": "array",
+              "items": {
+                "type": "array",
+                "minItems": 2,
+                "maxItems": 2,
+                "items": [
+                  {
+                    "$comment": "// PDB filename, e.g. mozglue.pdb",
```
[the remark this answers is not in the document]
CarlCorcoran: Will do! Quick question: the `examples` keyword seems to be a draft 6 addition, though we're working with the draft 4 spec. Do you think it's safe to use `examples` for a draft 4 schema? Our tests pass so it seems safe.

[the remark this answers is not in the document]
mreid-moz: Yup! `examples` is backwards compatible per [the draft 6 release notes](http://json-schema.org/draft-06/json-schema-release-notes.html#additions-and-backwards-compatible-changes).

mreid-moz:
@fbertsch This is ready for a closer look with the comment updates. Thanks @CarlCorcoran!

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
```diff
@@ -0,0 +1,313 @@
...
+            optional binary subsysID (UTF8);
+            optional int64 RAM (UINT_64);
+            optional binary driver (UTF8);
+            optional binary driverVersion (UTF8);
+            optional binary driverDate (UTF8);
+            optional boolean GPUActive;
+          }
+        }
+      }
+      optional group monitors (LIST) {
```
fbertsch: On Linux, monitors is a single object; see [0/1]. That will cause those pings to be discarded.

[0] https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/templates/include/telemetry/environment.1.schema.json#L509
[1] https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/environment.html

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
```diff
@@ -0,0 +1,313 @@
...
+        optional int64 scope;
+        optional boolean hasBinaryComponents;
+        optional int64 installDay (UINT_64);
+        optional int64 updateDay (UINT_64);
+      }
+      required group activeAddons (MAP) {
+        repeated group key_value {
+          required binary key (UTF8);
+          required group value {
+            required boolean blocklisted;
```
fbertsch: It doens't look like any of these are required by the JSON schema

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
```diff
@@ -0,0 +1,313 @@
...
+            optional int64 signedState;
+            required boolean isSystem;
+          }
+        }
+      }
+      required group activeGMPlugins (MAP) {
+        repeated group key_value {
+          required binary key (UTF8);
+          required group value {
+            required binary version (UTF8);
```
fbertsch: It also doesn't look like these are required

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
```diff
@@ -0,0 +1,313 @@
...
+            required boolean userDisabled;
+          }
+        }
+      }
+      required group activePlugins (LIST) {
```
fbertsch: Seems this can be either a list or an object [0], but the docs don't indicate when it's not a list. Any ideas? If it ever is returned an object the ping will be discarded.

[0] https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/templates/include/telemetry/environment.1.schema.json#L93

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
```diff
@@ -0,0 +1,313 @@
...
+        repeated group list {
+          required group element {
+            required binary name (UTF8);
+            required binary version (UTF8);
+            required binary description (UTF8);
+            required boolean blocklisted;
+            required boolean disabled;
+            required boolean clicktoplay;
+            required int64 updateDay;
+            required group mimeTypes (LIST) {
```
fbertsch: Another one that is both list and object.

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
```diff
@@ -0,0 +1,313 @@
...
+          required group value {
+            required binary version (UTF8);
+            required boolean userDisabled;
+          }
+        }
+      }
+      required group activePlugins (LIST) {
+        repeated group list {
+          required group element {
+            required binary name (UTF8);
```
fbertsch: None of these properties are required by the JSON schema, is that acceptable?

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
CarlCorcoran: The Untrusted Modules ping is Windows-only, but I'd prefer to be correct in any case. The code that generates this structure suggests it's always an array:

https://searchfox.org/mozilla-central/rev/4e094f66ced333d69b24cd49273789e3a1173dfc/widget/nsIGfxInfo.idl#69
https://searchfox.org/mozilla-central/rev/4e094f66ced333d69b24cd49273789e3a1173dfc/widget/GfxInfoBase.cpp#1199

I wonder if maybe it was an object (not an array) in previous versions? In that case I think it should be OK as-is.

The comment about Linux I think just means even if you have multiple monitors on Linux, only 1 monitor object will be returned representing the whole virtual screen.

Thoughts?

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
CarlCorcoran: Thanks for catching this; I corrected the required / optional for these and a few other properties based off the JSON schema.

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
CarlCorcoran: As above, I have corrected this.

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
CarlCorcoran: As above, I have corrected this.

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
CarlCorcoran: Hmm, looking at where we generate this, it is always returned as an array.

https://searchfox.org/mozilla-central/rev/4e094f66ced333d69b24cd49273789e3a1173dfc/toolkit/components/telemetry/app/TelemetryEnvironment.jsm#764

But as above I would prefer to be perfectly correct here, especially as fields in the environment block are really a nice-to-have for us. If there's a risk that we're filtering out good data because of this, I'd prefer to go safe than sorry.

What do you think?

schemas/telemetry/untrusted-modules/untrusted-modules.4.parquetmr.txt
CarlCorcoran: Seems this is also always an array in our current source code, but as above, what would you suggest?

https://searchfox.org/mozilla-central/rev/4e094f66ced333d69b24cd49273789e3a1173dfc/dom/plugins/base/nsIPluginTag.idl#61
https://searchfox.org/mozilla-central/rev/4e094f66ced333d69b24cd49273789e3a1173dfc/dom/plugins/base/nsPluginTags.cpp#672

PR #216, opened 2018-11-07
Add WebPageTest ping schema

akkomar:
This adds schema for WebPageTest results (see https://github.com/mozilla/wpt-api/issues/91)

/cc @stephendonner @davehunt


schemas/webpagetest/webpagetest-run/webpagetest-run.1.parquetmr.txt
```diff
@@ -0,0 +1,19 @@
+message webpagetest-run {
+  required binary appName (UTF8);
+  required binary channel (UTF8);
```
davehunt: Would it make sense to include the version and build ID of the application? I'm not sure if we can get this information from WebPageTest, but we at least have the user agent string from the request headers.

schemas/webpagetest/webpagetest-run/webpagetest-run.1.parquetmr.txt
```diff
@@ -0,0 +1,19 @@
...
+  required binary connection (UTF8);
+  required binary domain (UTF8);
```
davehunt: At the moment we're only testing landing pages, but I've some some suggestions that we may want to test content pages as they're more representative of what users are visiting. Perhaps a full URL option or path?

schemas/webpagetest/webpagetest-run/webpagetest-run.1.parquetmr.txt
```diff
@@ -0,0 +1,19 @@
...
+  required binary platform (UTF8);
+  required binary runner (UTF8);
```
davehunt: I suspect we'd put the `${JENKINS_URL}` value here.

schemas/webpagetest/webpagetest-run/webpagetest-run.1.parquetmr.txt
```diff
@@ -0,0 +1,19 @@
...
+  required int32 runId;
```
davehunt: Could this be a string? I'm thinking we'd use `${BUILD_TAG}` from Jenkins, which equates to `jenkins-${JOB_NAME}-${BUILD_NUMBER}`, for example: "jenkins-wpt/google-123".

schemas/webpagetest/webpagetest-run/webpagetest-run.1.parquetmr.txt
akkomar: Absolutely. Version (at least) and buildId (if you can get it) should be there.

schemas/webpagetest/webpagetest-run/webpagetest-run.1.parquetmr.txt
akkomar: Good point. We can just rename this to url, we're not handling millions of them so we should be fine without separate domain field?

schemas/webpagetest/webpagetest-run/webpagetest-run.1.parquetmr.txt
akkomar: Yes. We can also drop `runner` since example value you gave seems to be containing some Jenkins instance identifier `jenkins-wpt`.

davehunt: `runner` might still be useful as we _could_ have multple Jenkins instances. The "jenkins" in "jenkins-wpt/google-123" is hardcoded and does not distinuish the instance.

akkomar:
@mreid-moz thanks. Re: schema - yeah, looks like required fields don't bring much value here, re: tests - agree, especially single statistic case.

stephendonner:
This looks really good to me, @akkomar; I'm terribly sorry for the delay here, and I've resumed work from my side, in earnest, over in https://github.com/mozilla/wpt-api/pull/187.

I've only just started, though, and have few crucial pieces to line up, so feel free and encouraged to comment on/submit PRs, etc. :-)

akkomar:
@mreid-moz final r+? We're now convinced this schema fits WPT results.

mreid-moz:
cc @whd for a deploy

stephendonner:
For those following along, and for cross-reference/posterity, here's the "Deploy WebPageTest Schema" bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1527906

PR #217, opened 2018-11-09
[glean] Add place to store information about active experiments

mdboom:
https://bugzilla.mozilla.org/show_bug.cgi?id=1497807

### DIFF
--- a/templates/include/glean/glean.1.schema.json
+++ b/templates/include/glean/glean.1.schema.json
@@ -387,23 +387,24 @@
         "$ref": "#/definitions/short_id"
       },
       "additionalProperties": {
-        "type": "array",
-        "minItems": 1,
-        "maxItems": 2,
-        "items": [
-          {
+        "type": "object",
+        "properties": {
+          "branch": {
             "$ref": "#/definitions/short_id"
           },
-          {
+          "extra": {
             "type": "object",
-            "propertyNames": {
-              "type": "string"
-            },
-            "additionalProperties": {
-              "type": "string"
+            "properties": {
+              "type": {
+                "type": "string"
+              }
             }
           }
-        ]
+        },
+        "required": [
+          "branch"
+        ],
+        "additionalProperties": false
       }
     }
   },
--- a/templates/include/glean/glean.1.schema.json
+++ b/templates/include/glean/glean.1.schema.json
@@ -381,6 +381,30 @@
           }
         }
       ]
+    },
+    "experiments": {
+      "propertyNames": {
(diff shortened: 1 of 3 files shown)


templates/include/glean/glean.1.schema.json
```diff
@@ -381,6 +381,30 @@
...
         }
       ]
+    },
+    "experiments": {
+      "propertyNames": {
+        "$ref": "#/definitions/short_id"
+      },
+      "additionalProperties": {
+        "type": "array",
+        "minItems": 1,
```
fbertsch: Now this is interesting, can you elaborate on what the extra field is for? This does seem like a non-intuitive structure for this layout. It might be better to support it as e.g.:
```
[
  {
    "experiment_id": " experiment2",
    "experiment_branch": "branch_b",
    "extras: {
       "extra_key": "extra_val"
    }
]
```

But it really depends on what the `extra` part is for.

mdboom: I'm not at all opposed to making each of these entries an object.

"extras" is the same as "options" on desktop (renamed due to confusion).  It's basically a place for experiment authors to put extra metadata about the experiment settings, I suppose if a branch name alone isn't enough.

fbertsch: I'm not familiar with "options" on desktop, can you point me to that? All I see are `experiment_id` and `experiment_branch` [0].

[0] http://gecko-docs.mozilla.org.s3.amazonaws.com/toolkit/components/telemetry/telemetry/data/environment.html

mdboom: Ah, I think this might be based on my misunderstanding.  I was looking at this:

https://searchfox.org/mozilla-central/rev/efc0d9172cb6a5849c6c4fc0f19d7fd5a2da9643/toolkit/components/telemetry/app/TelemetryEnvironment.jsm#85-115

But if you drill down into the implementation, it looks like all it stores to send to the backend is `options.type`, if it exists.

https://searchfox.org/mozilla-central/source/toolkit/components/telemetry/app/TelemetryEnvironment.jsm#1018

I'm leaning to just taking that all out, but I'll bring it to a discussion first.

mdboom:
@frank: Updated this based to store each experiment as an object rather than an array, and retained `extras` as discussed.

mdboom:
@fbertsch: Would you mind merging this?  Cheers.

fbertsch:
@mdboom we should probably get you commit access to this repo

PR #218, opened 2018-11-14
Add proper typing for keyedHistograms

acmiyaguchi:
Keyed histograms are represented as doubly-nested maps. The first level of nested is missing the `"type": "object"` field, which was breaking a script.


acmiyaguchi:
It's a bit hard considering the duplication of measurements at the top-level and process-level. 

And yes, except the script was throwing an error instead of ignoring the field.

Issue #105, opened 2017-12-01
Beautify the parquet schema output on artifact generation


trink:

mreid-moz:
@trink this is done, right?

PR #219, opened 2018-11-19
Bug 1505874 - Add parquet glean schema

fbertsch:


schemas/glean/baseline/baseline.1.parquetmr.txt
```diff
@@ -0,0 +1,163 @@
...
+          required binary time_unit (UTF8);
+        }
+      }
+    }
+    optional group timing_distribution (MAP) {
+      repeated group key_value {
+        required binary key (UTF8);
+        optional group value {
+          optional group range (TUPLE) {
+  required double lo;
```
fbertsch: known formatting issue, see https://github.com/mozilla-services/mozilla-pipeline-schemas/issues/105

schemas/glean/baseline/baseline.1.parquetmr.txt
```diff
@@ -0,0 +1,163 @@
...
+    required binary client_id (UTF8);
+    required int32 seq;
+    required binary start_time (UTF8);
+    required binary end_time
+  }
+  optional group metrics {
+    optional group boolean (MAP) {
+      repeated group key_value {
+        required binary key (UTF8);
+        optional boolean value;
```
fbertsch: These values are all nullable (which it doesn't look like the JSON schema specifies). We can change this to required if we would like.

schemas/glean/baseline/baseline.1.parquetmr.txt
```diff
@@ -0,0 +1,163 @@
...
+    optional group string (MAP) {
+      repeated group key_value {
+        required binary key (UTF8);
+        optional binary value (UTF8);
+      }
+    }
+    optional group number (MAP) {
+      repeated group key_value {
+        required binary key (UTF8);
+        optional double value;
```
fbertsch: 64-bit floating point -- is this appropriate for the `number` type in the glean SDK?

schemas/glean/baseline/baseline.1.parquetmr.txt
```diff
@@ -0,0 +1,163 @@
...
+      }
+    }
+    optional group string_list (MAP) {
+      repeated group key_value {
+        required binary key (UTF8);
+        optional group value (LIST) {
+          repeated group list {
+            optional binary element (UTF8);
```
fbertsch: nullable string list, nullable elements in the list as well.

templates/include/glean/histogram.1.parquetmr.txt
```diff
@@ -0,0 +1,14 @@
+optional group range (TUPLE) {
```
fbertsch: Should the histogram type require any of these?

templates/include/glean/histogram.1.parquetmr.txt
```diff
@@ -0,0 +1,14 @@
...
+  required double lo;
```
fbertsch: These won't be named in the output (see the event tuple)

templates/include/glean/histogram.1.parquetmr.txt
```diff
@@ -0,0 +1,14 @@
...
+  required double hi;
+}
+optional inte64 bucket_count;
+optional binary histogram_type (UTF8);
+optional group values (MAP) {
+  repeated group key_value {
+    required binary key (UTF8);
+    optional int64 value;
```
fbertsch: What is the size here on the client? We do see lots of very large histogram values on desktop.

schemas/glean/baseline/baseline.1.parquetmr.txt
mdboom: The JSON schema specifies that the values are not nullable, and given the actual implementation of what generates these things, I don't think null values are possible.  That doesn't mean that's right...  What's the use case for sending a metric with a null value?  If there is one, we should probably be explicit about that and bubble that up to the spec document and implementation.

schemas/glean/baseline/baseline.1.parquetmr.txt
mdboom: Yes, I think so.  Other "numeric-like" types, like count, will be integers, but I think this one is sort of free-form/general purpose, so 64-bit float is probably the right choice.  (But we haven't implemented this one yet).

schemas/glean/baseline/baseline.1.parquetmr.txt
```diff
@@ -0,0 +1,163 @@
...
+
+          required binary time_unit (UTF8);
+        }
+      }
+    }
+    optional group datetime (MAP) {
+      repeated group key_value {
+        required binary key (UTF8);
+        optional group value {
+          optional binary value (UTF8)
```
mdboom: I see Parquet has specific [date/time formats](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#datetime-types).  Do we want to use those?  In JSON it's an ISO8601 string, but only because JSON doesn't have a datetime type.

schemas/glean/baseline/baseline.1.parquetmr.txt
```diff
@@ -0,0 +1,163 @@
...
+        optional group extra (MAP) {
+          repeated group key_value {
+            required binary key (UTF8);
+            optional binary value (UTF8);
+          }
+        }
+      }
+    }
+  }
+
```
mdboom: We'll need to update this for experiments from the incoming #217.

templates/include/glean/histogram.1.parquetmr.txt
mdboom: The JSON schema doesn't currently require any of these, but I feel like that's just an oversight on my part.  Definitely `range`, `bucket_count`, and `values` should be required.  `histogram_type` has an obvious default (linear) so could be optional, and `underflow` and `overflow` could default to zero.  Or I'd be fine with everything required, too.

templates/include/glean/histogram.1.parquetmr.txt
mdboom: We haven't implemented histograms yet, but I think it's reasonable to assume 64 bit like this.

templates/include/glean/histogram.1.parquetmr.txt
fbertsch: Great, I'll update that schema as well.

schemas/glean/baseline/baseline.1.parquetmr.txt
fbertsch: I'm not sure there is one. I'll make these required and match the json spec

schemas/glean/baseline/baseline.1.parquetmr.txt
fbertsch: That would be fantastic, but those store the values as ints. Let's leave it as-is and handle it at query time, and on BQ we can use this string type as an actual DATETIME type [0].

[0] https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#datetime-type

fbertsch:
@mdboom I've updated this with experiments, non-nulled columns, required elements for histograms, and events (per https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/213).

PR #220, opened 2018-11-20
1507788: Support dots in category part of dotted notation

mdboom:
Given [1507788](https://bugzilla.mozilla.org/show_bug.cgi?id=1507788), we need to support dots within the metric category name.  Therefore the `category.name` notation used to identify metrics can now have multiple dots.  This broadens the regex used to validate the metric name to support that.

### DIFF
--- a/templates/include/glean/glean.1.schema.json
+++ b/templates/include/glean/glean.1.schema.json
@@ -26,7 +26,7 @@
     },
     "dot_separated_short_id": {
       "type": "string",
-      "pattern": "^[a-z_][a-z0-9_]{0,29}\\.[a-z_][a-z0-9_]{0,29}$"
+      "pattern": "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})+$"
     },
     "base": {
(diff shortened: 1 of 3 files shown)


templates/include/glean/glean.1.schema.json
```diff
@@ -26,7 +26,7 @@
     },
     "dot_separated_short_id": {
       "type": "string",
-      "pattern": "^[a-z_][a-z0-9_]{0,29}\\.[a-z_][a-z0-9_]{0,29}$"
+      "pattern": "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})+$"
```
fbertsch: There is now no length limit here, is that expected?

mdboom: Good point.  I think we need to figure out what this limit should be, because obviously it should be longer than before...

@georgf ?

mdboom:
Actually, I'm mistaken -- the length here needs to be 40 (for category) + 20 (for name) + 1 (for the dot).  Updating...

PR #221, opened 2018-11-27
Rename untrusted-modules to untrustedModules

whd:


akkomar:
As discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1508749, this has been moved into https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/223

PR #223, opened 2018-11-29
Bug 1510704 - remove UINT_64 type annotations

akkomar:
This PR:
* removes UINT_64 type annotations from untrusted-modules ping. Fields with this type make parquet files unreadable for Spark.
* untrusted-modules to untrustedModules (previously introduced in [1], discussed in [2])

[1] https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/221
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1508749


whd:
@mreid-moz should be fixed now

PR #228, opened 2018-12-10
WIP: Add activity-stream impression-stats schema

jklukas:
Implements a schema matching the existing ping-centre payloads defined in
https://github.com/mozilla/activity-stream/blob/master/docs/v2-system-addon/data_events.md#top-story-pings

Should allow double-publishing of payloads via ping-centre and via telemetry.

### DIFF
--- a/schemas/activity-stream/impression-stats/impression-stats.1.schema.json
+++ b/schemas/activity-stream/impression-stats/impression-stats.1.schema.json
@@ -5,15 +5,15 @@
       "type": "string"
     },
     "block": {
+      "description": "A 0-based index to record which tile in the 'tiles' list the user just interacted with",
       "type": "integer"
     },
     "click": {
+      "description": "A 0-based index to record which tile in the 'tiles' list the user just interacted with",
       "type": "integer"
     },
-    "client_region": {
-      "type": "string"
-    },
     "impression_id": {
+      "description": "A UUID representing this impression",
       "pattern": "^\\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\\}$",
       "type": "string"
     },
@@ -30,6 +30,7 @@
       "type": "string"
     },
     "pocket": {
+      "description": "A 0-based index to record which tile in the 'tiles' list the user just interacted with",
       "type": "integer"
     },
     "profile_creation_date": {
@@ -41,12 +42,8 @@
     "release_channel": {
       "type": "string"
     },
-    "sample_id": {
-      "maximum": 99,
-      "minimum": 0,
-      "type": "integer"
-    },
     "shield_id": {
+      "description": "A semicolon separated string to store a list of Shield study IDs",
       "type": "string"
     },
     "source": {
--- a/schemas/activity-stream/impression-stats/impression-stats.1.schema.json
+++ b/schemas/activity-stream/impression-stats/impression-stats.1.schema.json
@@ -1,0 +1,94 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "properties": {
+    "action": {
+      "pattern": "^activity_stream_impression_stats$",
+      "type": "string"
+    },
+    "addon_version": {
+      "type": "string"
+    },
+    "block": {
+      "type": "integer"
+    },
+    "click": {
+      "type": "integer"
+    },
+    "client_id": {
+      "pattern": "^n/a$",
+      "type": "string"
+    },
+    "impression_id": {
+      "pattern": "^\\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\\}$",
+      "type": "string"
+    },
+    "locale": {
+      "type": "string"
+    },
+    "page": {
+      "enum": [
+        "about:newtab",
+        "about:home",
+        "about:welcome",
+        "unknown"
+      ],
+      "type": "string"
+    },
+    "pocket": {
+      "type": "integer"
+    },
+    "profile_creation_date": {
+      "type": "integer"
+    },
+    "region": {
+      "type": "string"
+    },
+    "release_channel": {
+      "type": "string"
+    },
+    "session_id": {
+      "pattern": "^n/a$",
+      "type": "string"
+    },
+    "source": {
+      "type": "string"
+    },
+    "tiles": {
+      "items": {
+        "properties": {
+          "id": {
+            "type": "integer"
+          },
+          "pos": {
+            "type": "integer"
+          }
+        },
+        "required": [
+          "id"
+        ],
+        "type": "object"
+      },
+      "type": "array"
+    },
+    "topic": {
+      "type": "string"
+    },
+    "user_prefs": {
+      "type": "integer"
+    },
+    "version": {
+      "type": "string"
+    }
+  },
+  "required": [
+    "impression_id",
+    "addon_version",
+    "locale",
+    "source",
+    "page",
+    "user_prefs",
+    "tiles"
+  ],
+  "title": "impression-stats",
+  "type": "object"
+}
--- a/schemas/activity-stream/impression-stats/impression-stats.1.schema.json
+++ b/schemas/activity-stream/impression-stats/impression-stats.1.schema.json
@@ -1,10 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   "properties": {
-    "action": {
-      "pattern": "^activity_stream_impression_stats$",
-      "type": "string"
-    },
     "addon_version": {
       "type": "string"
     },
@@ -14,8 +10,7 @@
     "click": {
       "type": "integer"
     },
-    "client_id": {
-      "pattern": "^n/a$",
+    "client_region": {
       "type": "string"
     },
     "impression_id": {
@@ -46,8 +41,12 @@
     "release_channel": {
       "type": "string"
     },
-    "session_id": {
-      "pattern": "^n/a$",
+    "sample_id": {
+      "maximum": 99,
+      "minimum": 0,
+      "type": "integer"
+    },
+    "shield_id": {
       "type": "string"
     },
     "source": {
@@ -69,9 +68,6 @@
         "type": "object"
       },
       "type": "array"
-    },
-    "topic": {
-      "type": "string"
     },
     "user_prefs": {
       "type": "integer"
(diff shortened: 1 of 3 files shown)


schemas/activity-stream/impression-stats/impression-stats.1.schema.json
```diff
@@ -0,0 +1,94 @@
+{
```
jklukas: This file is auto-generated. See the new file under `templates` for the user-created definition.

templates/activity-stream/impression-stats/impression-stats.1.schema.json
```diff
@@ -0,0 +1,87 @@
...
+      "pattern": "^activity_stream_impression_stats$"
+    },
+    "client_id": {
+      "type": "string",
+      "pattern": "^n/a$"
+    },
+    "session_id": {
+      "type": "string",
+      "pattern": "^n/a$"
+    },
```
jklukas: These first three fields are all expected to have static values, so we check those via regex patterns, but don't include the fields in the "required" section below. That means we will validate payloads either missing the "action" field or having action set to "activity_stream_impression_stats" exactly.

The idea is that this provides us compatibility with existing payloads today, but would allow us to stop sending these static fields in some future release where we stop needing to support sending payloads via the ping-centre infrastructure.

templates/activity-stream/impression-stats/impression-stats.1.schema.json
```diff
@@ -0,0 +1,87 @@
...
+    "topic": {
+      "type": "string"
+    },
+    "source": {
+      "type": "string"
+    },
+    "page": {
+      "type": "string",
+      "enum": [ "about:newtab", "about:home", "about:welcome", "unknown" ]
+    },
```
jklukas: We define "page" as an enum so that we only accept these explicit values.

jklukas:
For @ncloudioj - does this look reasonable? I wanted to give you a head start of seeing a concrete example of what contributing an Activity Stream schema to this repo would look like.

We can either move towards merging this PR, or we can just use this PR as a reference document for your team's work.

templates/activity-stream/impression-stats/impression-stats.1.schema.json
ncloudioj: `action` is merely a ping-centre specific field as the ping identifier, and we don't even bother storing it in the database, so I think we can leave it out here.

Both `client_id` and `session_id` are also omitable, they are all set as `n/a` in the current A-S payload for backward compatibility. We don't need to bear this burden in the new schema.

As a follow up question, how would the new data pipeline process those fields that are not defined in the schema? Does it simply throw them away, or still pack and store them in the database?

templates/activity-stream/impression-stats/impression-stats.1.schema.json
```diff
@@ -0,0 +1,87 @@
...
+    "version": {
+      "type": "string"
+    },
+    "release_channel": {
+      "type": "string"
+    },
+    "locale": {
+      "type": "string"
+    },
+    "topic": {
```
ncloudioj: `topic` is also a ping-centre specific field, we can drop it.

templates/activity-stream/impression-stats/impression-stats.1.schema.json
jklukas: > how would the new data pipeline process those fields that are not defined in the schema? Does it simply throw them away, or still pack and store them in the database?

In the existing AWS pipeline, extra fields are thrown away. We have a design goal for GCP of packaging up any remaining fields not included in the schema as a JSON blob, and including that as an additional field in the destination BigQuery table. Does that behavior sound good to you?

The code for this has not been written yet, though. I'll do a little more digging into what would be necessary to provide that.

templates/activity-stream/impression-stats/impression-stats.1.schema.json
```diff
@@ -0,0 +1,87 @@
...
+          },
+          "pos": {
+            "type": "integer"
+          }
+        },
+        "required": [ "id" ]
+      }
+    }
+  },
+  "required": [
```
ncloudioj: There are a few more fields that are not listed in the `data-event.md`.

* `shield_id`, varchar(256). A semicolon separated string to store a list of Shield IDs. 
* `hour`, integer
* `minute`, integer
* `sample_id`, server side generated integer ranges from 0 - 99
* `client_region`, varchar(5). Similar to `country_code`, but collected from the client side

Note that `hour` and `minute` are extracted from the timestamp field `receive_at`, which records the timestamp when Onyx receives the ping. Those two fields were requested by the Pocket team, need to confirm with them to see if they are still in use.

jklukas: Added these other than hour and minute.

templates/activity-stream/impression-stats/impression-stats.1.schema.json
```diff
@@ -0,0 +1,87 @@
...
+    "click": {
+      "type": "integer"
+    },
+    "block": {
+      "type": "integer"
+    },
+    "pocket": {
+      "type": "integer"
+    },
+    "tiles": {
```
ncloudioj: There are some extra complexities about this `tiles` field. See more details [here](https://github.com/mozilla/infernyx#payloads-from-firefox)

In a nutshell, both impression ping and other user interaction pings (click, delete, add to pocket) reuse this structure. For example, an impression ping with 3 Pocket stories,

```js
{
  // ... other fields
  "tiles": [
    {"id": 1000, "pos": 0},
    {"id": 1001, "pos": 1},
    {"id": 1002, "pos": 2},
  ]
}
```  

A click on the Pocket story "1002" at the position "2" would generate the following ping:
```js
{
  // ... other fields
  "click": 0,    // 0 indicating the index of the clicked Pocket story stored in the "tiles" array below 
  "tiles": [
    {"id": 1002, "pos": 2},
  ]
}
```
Infernyx (ETL jobs) will store this ping to the database

```js
{
  "tile_id": 1001,
  "position": 2,
  "clicks": 1,
  "blocked": 0,
  "pinned": 0,
  "pocketed": 0 
}
```

The same rule could be applied to `delete`, `pin`, `save to pocket` etc.

This flat schema is very friendly for the aggregate queries for Pocket, such as counting impressions for each story, clicks, deletes, save to pockets etc. Also, it's quite straightforward to query the unique audience of the certain `tile_id`. Wonder if we can still keep those merits if we're looking to keep this `tiles` as an array in the table.

If not, we might want to split this array either on the client side, or somewhere in the data pipeline.

templates/activity-stream/impression-stats/impression-stats.1.schema.json
ncloudioj: Yep, that'll be great!

Also, we will make sure that anything unnecessary won't be included in the payload.

templates/activity-stream/impression-stats/impression-stats.1.schema.json
jklukas: > If not, we might want to split this array either on the client side, or somewhere in the data pipeline.

The model we have in mind for the data pipeline in GCP is that every schema in mozilla-pipeline-schemas will map to exactly one table in BigQuery. The table in BigQuery will have exactly the same structure as the payload, save for some additional metadata fields.

So, we'd still want to define some sort of ETL process to serve the kinds of needs you're talking about here.

This kind of transformation sounds like a good candidate for either an on-demand SQL view or a "materialized view" (called a "destination table" in BigQuery) where we have a query run on some schedule to update the target.

If latency is a bigger issue, we could also consider a dataflow streaming job to do such a transformation.

ncloudioj: > The model we have in mind for the data pipeline in GCP is that every schema in mozilla-pipeline-schemas will map to exactly one table in BigQuery. The table in BigQuery will have exactly the same structure as the payload, save for some additional metadata fields.

Gotcha. Since BigQuery has a great query support on Array types, I think we can just leave the schema as-is, and try it out without introducing other complexities for now.

ncloudioj:
@jklukas Does the pipeline support field mapping or renaming? For example, can we store a field called `block` in the payload to the field `delete` in the database?

Again, for some historical reasons, there are some fields are stored with different names in the database for A-S.

jklukas:
> Does the pipeline support field mapping or renaming? For example, can we store a field called block in the payload to the field delete in the database?

No, but field renaming is a case that's particularly well suited to views. We could define a view like the following:

```
CREATE VIEW table_with_renamed_columns AS (
  SELECT field1, field2, block AS delete, field4
  FROM base_table
);
```

Then you can basically ignore that `base_table` exists and always query `table_with_renamed_columns` instead. I'd expect negligible performance impact in this case for querying the view rather than the base table.

ncloudioj:
Yep, we could use views if folks feel certain fields are confusing or misleading.

I'd assume that creating new views need separate requests, correct? Most of users would only have the read-only access in BQ.

jklukas:
> I'd assume that creating new views need separate requests, correct? Most of users would only have the read-only access in BQ.

Yes, we'd need to figure out permissions. It might be possible to define a separate dataset in BQ for housing these views and give relevant A-S and Pocket folks write access for that schema.

templates/activity-stream/impression-stats/impression-stats.1.schema.json
```diff
@@ -0,0 +1,83 @@
...
+  "title": "impression-stats",
+  "properties": {
+    "impression_id": {
+      "type": "string",
+      "pattern": "^\\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\\}$"
+    },
+    "shield_id": {
+      "type": "string"
+    },
+    "sample_id": {
```
ncloudioj: Correction: `sample_id` is *not* included in the payload from client. Instead, it's a random integer auto generated by the database. My apologies for the confusion. How can we handle this special case? I.e. Having a default (or auto generated) field in the DB schema.

jklukas: > How can we handle this special case? I.e. Having a default (or auto generated) field in the DB schema, but not in the payload schema.

A static default value could be handled in a SQL view definition, same as renaming a field. A SQL view could also handle running a hash function to stably generate a sample_id from properties of the message itself, as we do for telemetry where we hash client_id to get sample_id.

The case of the database auto-generating a random value, though, would require a periodic scheduled ETL job that inserts records into a destination table.

ncloudioj:
Just confirmed with Pocket team, both `hour` and `minute`fields are still in use. So let's keep them in the schema.

jklukas:
Just pushed a new commit that removes the fields you indicated were generated by the server rather than the client. Also added `"description"`s for several of the less-obvious fields.

Any other comments at the moment, or does this look like something we could merge as a starting point?

ncloudioj:
@jklukas LGTM, go ahead merge it. 🚢

I am meeting Pocket team today, will catch them up about this work. Let's see if they have any inputs on this.

PR #229, opened 2018-12-11
Add parquet schema beautification

trink:


Dockerfile
```diff
@@ -19,8 +19,9 @@ RUN pip install jsonschema
 WORKDIR /downloads
 
 # Install hindsight and the luasandbox
-RUN wget -qO- https://hsadmin.trink.com/packages/centos7/all.tgz | tar xvz
-RUN wget https://hsadmin.trink.com/packages/centos7/external/parquet-cpp-1.3.1-1.x86_64.rpm
+RUN wget -qO- https://s3-us-west-2.amazonaws.com/net-mozaws-data-us-west-2-ops-ci-artifacts/mozilla-services/lua_sandbox_extensions/master/centos7/all.tgz | tar xvz
+RUN wget https://s3-us-west-2.amazonaws.com/net-mozaws-data-us-west-2-ops-ci-artifacts/mozilla-services/lua_sandbox_extensions/external/centos7/parquet-cpp-1.3.1-1.x86_64.rpm
```
fbertsch: I'm not seeing the `external` branch in github, is that expected?

trink:
Yeah, the docker file uses a pre-built Linux binary instead of installing Rust/Cargo and building it each time.  The readme installation instructions are for local users (if they happen to be on Linux they could also grab the pre-built binary :) )

Dockerfile
fbertsch: How did these get deployed to the artifacts bucket if there's no associated branch?

trink: Ops put them there

fbertsch: Got it. Anyways feel free to merge unless you want to move parquetfmt to mozilla-services

fbertsch: Given that the docker setup is running fine I think we're good as it is.

mreid-moz: If docker can be used to format the schemas locally, I'm fine with either way. If it's something that runs as part of `cmake`, it'd be nice to keep it as simple as possible for people who don't have a rust environment set up, so I'd lean towards adding pre-built binaries for win+osx.

trink: done

PR #230, opened 2018-12-11
Add glean events docType

mdboom:
Fixes [1510552](https://bugzilla.mozilla.org/show_bug.cgi?id=1510552).


templates/glean/events/events.1.schema.json
```diff
@@ -0,0 +1 @@
+@GLEAN_GLEAN_1_JSON@
```
fbertsch: Oh this is too easy.

PR #231, opened 2018-12-12
Bug 1503256 Add pull request template with reviewer instructions

jklukas:
Also adds caching of the data from S3, which brings runtime for the integration test down to ~60 seconds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,20 +4,6 @@
 
 version: 2.1
 
-commands:
-  early_return_for_forked_pull_requests:
-    description: >-
-      If this build is from a fork, stop executing the current job and return success.
-      This is useful to avoid steps that will fail due to missing credentials.
-    steps:
-      - run:
-          name: Early return if this build is from a forked PR
-          command: |
-            if [ -n "$CIRCLE_PR_NUMBER" ]; then
-              echo "Nothing to do for forked PRs, so marking this step successful"
-              circleci step halt
-            fi
-
 jobs:
   test:
     docker:
@@ -32,18 +18,19 @@ jobs:
           name: Test Code
           command: docker run mps
   integrate:
-    docker:
+    docker: &edge_validator
       - image: mozilla/edge-validator:1.3
     steps:
-      - early_return_for_forked_pull_requests
       - checkout
-      - run:
+      - run: &checkout_upstream
           name: Checkout upstream
           command: |
             git remote add upstream git@github.com:mozilla-services/mozilla-pipeline-schemas.git
             git fetch --all
       - setup_remote_docker
-      - run:
+      - restore_cache:
+          key: mps-integration-data-v1
+      - run: &run_comparison
           name: Run a comparison report
           environment:
             SOURCE_DATA_BUCKET: telemetry-parquet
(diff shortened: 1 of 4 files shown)


jklukas:
Notice that this is a forked PR, but the `integrate` test shows up here because I ran:

    ./.github/push-to-trigger-integration jklukas:review-docs

jklukas:
Requesting review from @acmiyaguchi as the author of the tests and from @frank as someone I know has been a recent reviewer of PRs. Does the workflow encoded here sound reasonable?

.circleci/config.yml
```diff
@@ -63,13 +50,51 @@ jobs:
...
+              branches:
+                # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
+                ignore: /pull\/[0-9]+/
+    nightly:
+      jobs:
+        - integrate-clean
+      triggers:
+        - schedule:
+            # Run a little after midnight Pacific every day.
+            cron: "33 8 * * *"
```
acmiyaguchi: This schedules the integration tests every night? If so, this is nifty.

.circleci/config.yml
```diff
@@ -63,13 +50,51 @@ jobs:
...
+      - store_artifacts: &store_artifacts
           path: /app/test-reports
           destination: /app/test-results
+      - save_cache: &save_cache
+          paths:
+            - ~/resources
+          key: mps-integration-data-v1-{{ epoch }}
+  integrate-clean:
+    docker:
+      - image: mozilla/edge-validator:1.3
```
acmiyaguchi:
```suggestion
      <<: *edge_validator
```

.circleci/config.yml
jklukas: That is correct!

jklukas:
> One thing I noticed is that the build artifacts aren't available on the CircleCI page for integration. Do you know where they went?

I see artifacts uploaded at:

https://circleci.com/gh/mozilla-services/mozilla-pipeline-schemas/329#artifacts/containers/0

Though both of those result files are just empty json blobs. It's not clear to me if that's expected.

.circleci/config.yml
jklukas: Pushed an update that refactors this to an executor.

acmiyaguchi:
How do you get to that page from a PR? I clicked on the green checkbox and looked at the integrate job, but it's not there.

image

Actually, it looks like I just have to watch the `trigger-integration` jobs in order to see the artifacts which makes sense. 

It's unexpected that they're empty. I ran `pipenv run python integration.py report --report-path test.json` in a container on my local machine, which generates a non-empty report. The[`compare` command](https://github.com/mozilla-services/edge-validator/blob/master/integration.py#L303) calls this same reporter twice to generate a diff. It doesn't look like anything in this PR should have introduced to break it, so it could be the integration script itself that's causing issues.

jklukas:
> How do you get to that page from a PR?

If you follow the link to https://circleci.com/gh/mozilla-services/mozilla-pipeline-schemas/333 there's an "Artifacts" tab, which you can see in the bottom right of this screenshot:

screen shot 2018-12-13 at 3 19 01 pm

jklukas:
> It doesn't look like anything in this PR should have introduced to break it, so it could be the integration script itself that's causing issues.

Could the caching of `resources` be potentially causing issues with the reports?

.circleci/config.yml
```diff
@@ -32,18 +23,18 @@ jobs:
...
           command: |
             git remote add upstream git@github.com:mozilla-services/mozilla-pipeline-schemas.git
             git fetch --all
       - setup_remote_docker
-      - run:
+      - restore_cache:
+          key: mps-integration-data-v1
+      - run: &run_comparison
           name: Run a comparison report
           environment:
```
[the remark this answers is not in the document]
jklukas: They should be included as part of

```
      - run:
          <<: *run_comparison
```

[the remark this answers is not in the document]
acmiyaguchi: I ssh'ed into a box to see what the current state of the `resources/data` folder looks like.

The [edge-validator script](https://github.com/mozilla-services/edge-validator/blob/ebbab1af5e902537e7b7dfbfdb98d1eb62d32be1/integration.py#L148-L150) assumes a file structure `resources/data/<submission date>/<namespace>/<doctype>`, but it looks like the cache is missing the submission date sub-folder. It does contain the correct `namespace` data, so it looks like the environment variables are being set correctly.

acmiyaguchi:
I found the bug buried deep in the [rsync script](https://github.com/mozilla-services/edge-validator/blob/ebbab1af5e902537e7b7dfbfdb98d1eb62d32be1/sync.sh#L69-L72) written in bash:

```bash
        submission_date=$(echo "${path}" | cut -d'/' -f4 | cut -d'=' -f2)
        namespace=$(echo "${path}" | cut -d'/' -f5 | cut -d'=' -f2)
        doc_type=$(echo "${path}" | cut -d'/' -f6 | cut -d'=' -f2)
        doc_version=$(echo "${path}" | cut -d'/' -f7 | cut -d'=' -f2)
```
These offsets were assuming the old default location of `pipeline-analysis/amiyaguchi/sanitized-landfill-sample/v3` which then got changed to `telemetry-parquet/sanitized-landfill-sample/v3`. The cut indexes are off by 1, which chops a directory in the final data resource folder. This will need to be fixed in the `edge-validator` code so that reports are valid.

PR #234, opened 2018-12-14
1513939: Add android reference browser

mdboom:
This adds a new application url `android-reference-browser`, which will take vanilla glean telemetry.

Also, this adds the parquet template to `glean`, which was missed in #230.

Question (@fbertsch, @mreid): How do we want to handle testing?  It seems redundant / potentially hard to keep in sync to just copy the tests from `glean` into `android-reference-browser`, but it doesn't appear that the templating works on test content.  Any thoughts?

Checklist for reviewer:


mreid-moz:
Agreed that it's redundant to copy the tests over. I think we should include an actual example of a ping produced by the new application `android-reference-browser`, and (if applicable) test pings for anything that is unique/different about the new application.

mdboom:
So in that case, I suppose that means no tests for now, since the new application isn't integrated with glean yet and we're not aware of any application-specific differences for now...

fbertsch:
Also not really any integration tests we can do, since there's no data available.

schemas/glean/events/events.1.parquetmr.txt
```diff
@@ -0,0 +1,183 @@
...
+      repeated group key_value {
+        required binary key (UTF8);
+        required binary value (UTF8);
+      }
+    }
+  }
+  optional group events (LIST) {
+    repeated group list {
+      required group element (TUPLE) {
+  required int64 timestamp;
```
fbertsch: These issues should have been fixed with the recent parquetfmt changes, can you rebase and rebuild the schemas?

mdboom: Sure thing.  Should be done now.

mdboom:
@fbertsch: I think I got this now.  We'll have to convert `org.mozilla.reference.browser` to `org-mozilla-reference-browser` on the clients end before hitting this URL.

PR #238, opened 2018-12-27
Fix empty integration reports by updating edge-validator image

acmiyaguchi:
Checklist for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ version: 2.1
 executors:
   edge-validator:
     docker:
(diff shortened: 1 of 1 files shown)


acmiyaguchi:
It looks like the integration CI won't work correctly because the existing cache contains data in the wrong format. This should be fixed on the nightly run, but in the meanwhile, I'll add a testing patch for verification.

jklukas:
You should be able to bump the cache version to v2 as a permanent part of this change and that should allow tests to run correctly.

PR #242, opened 2019-01-10
Rename the `glean` namespace to `org-mozilla-samples-glean`

Dexterp37:
This is needed since the namespace changed in the sample application itself. We don't care too much about loosing the data, as this is a sample application.

Checklist for reviewer:

### DIFF
--- a/schemas/glean/events/events.1.parquetmr.txt
+++ b/schemas/glean/events/events.1.parquetmr.txt
@@ -1,0 +1,179 @@
+message baseline {
+  required group metadata {
+    optional binary EnvVersion (UTF8);
+    optional binary Host (UTF8);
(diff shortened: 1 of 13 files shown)


Dexterp37:
This is ready for another look @fbertsch !

Dexterp37:
> LGTM! We also need to add the events parquet output for the reference-browser, but if you don't want to do that here you don't have to.

@fbertsch is [this](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/org-mozilla-reference-browser/events/events.1.parquetmr.txt) the file? If so, looks like it's already there :)

fbertsch:
Well lookie there, you're right!

PR #243, opened 2019-01-10
Add schema for core ping version 10

jklukas:
Version was increased as part of
https://hg.mozilla.org/mozilla-central/rev/5cb4e28e1eb8

Checklist for reviewer:

### DIFF
--- a/schemas/telemetry/core/core.10.schema.json
+++ b/schemas/telemetry/core/core.10.schema.json
@@ -1,0 +1,107 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "properties": {
+    "accessibilityServices": {
+      "items": {
+        "type": "string"
+      },
+      "type": "array"
+    },
+    "arch": {
+      "type": "string"
+    },
+    "bug_1501329_affected": {
+      "type": "boolean"
+    },
+    "campaign": {
+      "type": "string"
+    },
+    "clientId": {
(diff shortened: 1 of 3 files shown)


whd:
Note that the way we're provisioning core ping parquet output on the AWS infra is such that the schema for v9 is applied to all core pings, not just v9, which is why there is no difference between the v9 and v10 parquet schemas. As such, there may be no need to add a v10 parquet output.

What this brings to light is that we're probably not doing validation of core v10 pings json-side, since this schema is missing (but the relevant change was applied to v9). The behavior of the current infra, absent a schema, is to assume the payload is valid.

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1410156

jklukas:
Given that context, I agree it doesn't make sense to add the parquet schemas for v10. Will remove.

whd:
It looks like it would also be technically correct to revert the change on the v9 json schema that is for v10 pings (but not revert it for the v9 parquet output, which is really for all core pings). Given the confusion here, it may make sense to rename the v9 parquet output to remove its version, but that will require a redeploy of the ingestion infrastructure.

whd:
That it did not require a version bump is probably why @fbertsch added the changes to v9. My guess is that he was unaware that the client team was going to bump the version due to this addition.

At any rate, leaving v9 as-is seems reasonable to me.

fbertsch:
Also, @whd is correct - we somehow missed on our end that this was going to be a version bump on the client.

PR #246, opened 2019-01-21
Bug 1518490: Add pipeline schemas for 2 new timing fields in untrusted modules ping

CarlCorcoran:
This modifies the untrusted modules ping Parquet and JSON schemas to add 2 new
fields:
- "xulLoadDurationMS" (optional, number) at the payload root
- "loadDurationMS" (optional, number) per module object

And adds these fields to validation schema: untrustedModules.4.sample.pass.json

These fields are marked optional in order to be compatible with the previous
schema.

The code that generates these fields landed in
https://hg.mozilla.org/mozilla-central/rev/1b1fe5815dff

Checklist for reviewer:

### DIFF
--- a/templates/telemetry/untrustedModules/untrustedModules.4.schema.json
+++ b/templates/telemetry/untrustedModules/untrustedModules.4.schema.json
@@ -70,7 +70,8 @@
           "minimum": 0
         },
         "xulLoadDurationMS": {
-          "type": "number"
+          "type": "number",
+          "minimum": 0
         },
         "events": {
           "type": "array",
@@ -115,7 +116,8 @@
                       "minimum": 0
                     },
                     "loadDurationMS": {
-                      "type": "number"
+                      "type": "number",
+                      "minimum": 0
                     }
                   }
                 },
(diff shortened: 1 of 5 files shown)


templates/telemetry/untrustedModules/untrustedModules.4.schema.json
```diff
@@ -69,6 +69,9 @@
           "type": "integer",
           "minimum": 0
         },
+        "xulLoadDurationMS": {
+          "type": "number"
```
fbertsch: Sometimes it's useful to add `"minimum": 0`, but I'll leave that up to you.

CarlCorcoran: Makes sense here; I added these constraints in my latest push.

Regarding the version, is there any thing for me to do there?

fbertsch: No, ops will handle that with a new deploy.

fbertsch:
cc @whd, this deploy will require a version bump.

whd:
Accordingly this will go out on the next full deploy, which is either next week or the week after.

CarlCorcoran:
> @CarlCorcoran do you care about having the v1 table include complete data (sans the new fields) for some amount of time or is it fine to simply query v2?

@whd Whatever is simplest for you. We're happy to just query against v2.

whd:
Ok, we're doing a hard cutover then.

whd:
This cutover has happened, and the data should be available in re:dash under v2 soon.

PR #247, opened 2019-01-23
Add install ping support for ARM64 platform (fixes #237)

### LINKED ISSUE #237: Add AArch64 platform to installer schema (Bugzilla bug 1514799)
Currently the installer ping has boolean fields for 64bit_build and 64bit_os, which are false for x86 and true for x64. We're now adding support for AArch64, so booleans are no longer enough and those need to be replaced by enums that allow x86, x64, and AArch64.

mhowell-mozilla:
Adds a new version of the install ping schema which removes the `64bit_build` and `64bit_os` booleans, replacing them with `build_arch` and `os_arch` enums so that we can add a third platform (ARM64) and easily add more in the future if needed.

I couldn't find any documentation about the intended processes for changing existing schemas like this, and I haven't started on the patch to send this ping version yet, so if this isn't the right way to go about it, I can easily scrap it and do something different.

Checklist for reviewer:


acmiyaguchi:
Apologies for not having looked at this in almost 2 years (yikes). Is this something that would still be useful? The PR certainly needs a rebase (but it should be straightforward to add).

mhowell-mozilla:
Thanks for the ping, almost forgot about this one myself. I'm going to just close this; so far ARM64 Windows hasn't turned out to really need to get broken out like this, and if that does need to happen some time in the future it'll be just as easy to make a fresh PR.

PR #248, opened 2019-01-24
Bug 1502921 - Add new 'intl' fields to environment data schema

flodolo:
CC @mstriemer

Bug 1502921 added a few field to the Environment ping. Per IRC discussion, we need this data to be validated in order to add it to datasets in STMO.

Landing in m-c (Firefox 66): https://hg.mozilla.org/mozilla-central/rev/de6426e0f491

### DIFF
--- a/templates/include/telemetry/environment.1.schema.json
+++ b/templates/include/telemetry/environment.1.schema.json
@@ -379,13 +379,13 @@
               }
             },
             "systemLocales": {
-              "type": "array",
+              "type": [ "array", "null" ],
               "items": {
                 "type": "string"
               }
             },
             "regionalPrefsLocales": {
-              "type": "array",
+              "type": [ "array", "null" ],
               "items": {
                 "type": "string"
               }
--- a/templates/include/telemetry/environment.1.schema.json
+++ b/templates/include/telemetry/environment.1.schema.json
@@ -357,6 +357,47 @@
         "e10sCohort": {
           "type": "string"
         },
+        "intl": {
+          "type": "object",
+          "properties": {
+            "requestedLocales": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "availableLocales": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "appLocales": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "systemLocales": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "regionalPrefsLocales": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "acceptLanguages": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          }
+        },
         "locale": {
           "type": "string"
         },
--- a/schemas/telemetry/crash/crash.4.schema.json
+++ b/schemas/telemetry/crash/crash.4.schema.json
@@ -443,6 +443,53 @@
             "e10sEnabled": {
               "type": "boolean"
             },
+            "intl": {
+              "properties": {
+                "acceptLanguages": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": "array"
+                },
+                "appLocales": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": "array"
+                },
+                "availableLocales": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": "array"
+                },
+                "regionalPrefsLocales": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": [
+                    "array",
+                    "null"
+                  ]
+                },
+                "requestedLocales": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": "array"
+                },
+                "systemLocales": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": [
+                    "array",
+                    "null"
+                  ]
+                }
+              },
+              "type": "object"
+            },
             "isDefaultBrowser": {
               "type": [
                 "boolean",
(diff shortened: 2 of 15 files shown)


templates/include/telemetry/environment.1.schema.json
```diff
@@ -357,6 +357,29 @@
         "e10sCohort": {
           "type": "string"
         },
+        "intl": {
+          "type": "object",
+          "properties": {
+            "requestedLocales": {
+              "type": "string"
```
mstriemer: I think these should be:

```json
"requestedLocales": {
  "type": "array",
  "items": {
    "type": "string"
  }
```

flodolo: Ah, I missed we have an array type. I think you're right

flodolo: I force pushed since GH these days shows them anyway in the history

mstriemer:
`regionalPrefsLocales` and `systemLocales` can be null instead of an array. This is similar to `locale` which can be a string or null, but that only defines `"type": "string"`. Not sure if anything needs to be done to handle that.

They shouldn't be null, but the code allows for it if the mozIOSPreferences object thinger isn't loaded for some reason (likely just on android tests, but I don't know for sure).

flodolo:
> `regionalPrefsLocales` and `systemLocales` can be null instead of an array. This is similar to `locale` which can be a string or null, but that only defines `"type": "string"`. Not sure if anything needs to be done to handle that.

Would that be this?

```
"regionalPrefsLocales": {
    "type": [ "array", "null" ],
    "items": {
    "type": "string"
    }
},
```

mreid-moz:
> I think so, but I'd defer to a maintainer of this project. Seeing as it isn't set for locale I'm not sure it's needed.

It would be a good idea to add a couple of examples to the validation pings that exercise this case. If the field is simply missing in the JSON, you don't need to include `null` as a type, but if you see a literal `"regionalPrefsLocales": null` in the incoming documents, it _is_ needed.

flodolo:
> It would be a good idea to add a couple of examples to the validation pings that exercise this case

Mark, would you be able to provide that?

acmiyaguchi:
@flodolo To clarify Mark's comment, it's good practice to add validation pings that functionally act as unit tests. They can be added to `mozilla-pipeline-schemas/validation/telemetry`. 

The convention is `<document-type>.<version>.<validation-name>.<[pass | fail]>.json`. You can start with the minimal main document called [`main.4.min.pass.json`](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/validation/telemetry/main.4.min.pass.json) to construct the test cases:

https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/8939bcb719855c6d84fbf2cfb4c256f2123d7026/validation/telemetry/main.4.min.pass.json#L1-L16

flodolo:
> @flodolo To clarify Mark's comment, it's good practice to add validation pings that functionally act as unit tests. They can be added to `mozilla-pipeline-schemas/validation/telemetry`.

Would this be run automatically via integration, or do I need to run tests manually (and how)?

I'd still need some input from Mark on the null cases, because I'm not really a developer, and I'm not sure which cases should be tested.

acmiyaguchi:
They are run during integration, you can "Show all checks" to see the logs from the tests. If you have docker (or have an environment with cmake installed), you can follow the [instructions in the README](https://github.com/mozilla-services/mozilla-pipeline-schemas#running-tests-via-docker) to test locally.

It looks like there are three cases that should be tested from this thread, to be comprehensive:

1. The normal case where the field is an array (new pings)
2. A degenerate case where the field is null per https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/248#issuecomment-457334336
3. A case where the field doesn't exists (for pre-existing pings)

Using this schema as reference:

```
"regionalPrefsLocales": {
    "type": [ "array", "null" ],
    "items": {
        "type": "string"
    }
},
```

You might have something like:

```json
"intl": {
    ...
}
```

```json
"intl": {
    ...
    "regionalPrefsLocales": null,
    ...
}
```

and

```json
"intl": {
    ...
    "regionalPrefsLocales": ["hello", "world"],
    ...
}
```

flodolo:
Thanks for all the info. I'll take a look next week and see what I can come up with.

flodolo:
I think I finally figured out how to get tests pass…

@mstriemer 
Can you verify if the test cases make sense based on the code?

flodolo:
@mstriemer review ping :)

Side question: if/when this is merged, what are the next steps? Is the data going to be backfilled, since pings already include this information?

mstriemer:
The schema and tests look right to me.

flodolo:
Sorry, somehow your reply got completely lost in my mailbox.

I rebase on top of dev just in case, run the command and added it as a separate commit.

acmiyaguchi:
@flodolo Thanks for updating the schema :+1:. Adding this information to the schema will allow us to validate the pings in the pipeline. There's no backfill necessary, since extra fields are generally allowed in the documents. 

The data can be access through the pings via the [moztelemetry API](https://docs.telemetry.mozilla.org/tools/spark.html#the-moztelemetry-library) in a python notebook. If you want to query using sql.telemetry.mozilla.org, the new environment fields  would need to be added explicitly to a dataset like `main_summary`.

flodolo:
> If you want to query using sql.telemetry.mozilla.org, the new environment fields would need to be added explicitly to a dataset like `main_summary`.

How can that be done? All the queries and dashboards I have live on STMO, so it would be great to have that. 

When adding these fields we were also considering removing the `locale` field, but left it alone as a future follow-up.

flodolo:
(For further reference: I normally look at the `client_count_daily` dataset)

acmiyaguchi:
The best next step is to file a bug under `Data Platform and Tools :: Datasets: Main Summary` to add the new environment fields to the dataset. We wouldn't do a backfill on the data here, since the fields haven't hit the release channel yet, and it's time consuming to backfill.

PR #249, opened 2019-01-24
Remove "object" type for monitors

fbertsch:
Keeping the two types makes the type of this field be "string" in BQ, which is not ideal. We'll need to thoroughly check that this doesn't break the schema.

I will be making changes to other parts as well (so they are more easily queryable), if you'd like you can hold off on reviewing until they are all in.

Checklist for reviewer:

### DIFF
--- a/templates/include/telemetry/environment.1.schema.json
+++ b/templates/include/telemetry/environment.1.schema.json
@@ -509,7 +509,7 @@
               }
             },
             "monitors": {
-              "type": [ "array", "object" ],
+              "type": [ "array" ],
               "items": {
                 "type": "object",
                 "properties": {


fbertsch:
A bunch of these came up in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/215, and in talking with @georgf as well it sounds like we don't need the dual types. For example, this one is always an array. I'm going to do them one-at-a-time so we can easily verify each change.

templates/include/telemetry/environment.1.schema.json
```diff
@@ -509,7 +509,7 @@
               }
             },
             "monitors": {
-              "type": [ "array", "object" ],
+              "type": [ "array" ],
```
acmiyaguchi:
```suggestion
              "type": "array",
```

fbertsch: Good point. I did this in the GH editor, not a good decision for something that required a build.

acmiyaguchi:
@frank bump on this PR

This is also causing issues with schema transpilation because there are ["array", "object"] types that don't provide an items field.

find-array-object.py
```python
#!/usr/bin/env python3

import json
import sys
from collections import deque
from pprint import PrettyPrinter

def main():
    # traverse a json document, finding path to leaves that match exactly with the term
    document = json.load(sys.stdin)
    
    search_term = {"object", "array"}
    results = []

    queue = deque([("root", document)])
    while queue:
        name, node = queue.popleft()
        if not isinstance(node, dict):
            continue
        dtype = node.get("type")
        if dtype and isinstance(dtype, list) and search_term.issubset(set(dtype)):
            results.append((name, node))
        for key, value in node.items():
            queue.append((f"{name}.{key}", value))
    
    pp = PrettyPrinter()
    pp.pprint(results)


if __name__ == "__main__":
    main()
```
using command:
`find schemas/ -name "*.schema.json" | xargs -I {} bash -c "echo {}; cat {} | python3 find-array-object.py"`

Gives some useful information about particular schemas that have this error. This PR should clear up quite a few of them.

acmiyaguchi:
See #291 for updates to this field.

acmiyaguchi:
@fbertsch I'm going to close this PR, I fixed this through #291 in the process of developing avro schemas from this repo.

PR #254, opened 2019-02-04
1516527: Add labeled metrics to glean

mdboom:
This adds labeled versions of all the existing metric types to glean.

Checklist for reviewer:

### DIFF
--- a/templates/include/glean/glean.1.schema.json
+++ b/templates/include/glean/glean.1.schema.json
@@ -77,7 +77,7 @@
         "labeled_boolean": {
           @GLEAN_BASE_OBJECT_1_JSON@,
           "additionalProperties": {
-            @GLEAN_BASE_OBJECT_1_JSON@,
+            @GLEAN_LABELED_GROUP_1_JSON@,
             "additionalProperties": @GLEAN_BOOLEAN_1_JSON@
           }
         },
@@ -88,7 +88,7 @@
         "labeled_string": {
           @GLEAN_BASE_OBJECT_1_JSON@,
           "additionalProperties": {
-            @GLEAN_BASE_OBJECT_1_JSON@,
+            @GLEAN_LABELED_GROUP_1_JSON@,
             "additionalProperties": @GLEAN_STRING_1_JSON@
           }
         },
@@ -99,7 +99,7 @@
         "labeled_number": {
           @GLEAN_BASE_OBJECT_1_JSON@,
           "additionalProperties": {
-            @GLEAN_BASE_OBJECT_1_JSON@,
+            @GLEAN_LABELED_GROUP_1_JSON@,
             "additionalProperties": @GLEAN_NUMBER_1_JSON@
           }
         },
@@ -110,7 +110,7 @@
         "labeled_string_list": {
           @GLEAN_BASE_OBJECT_1_JSON@,
           "additionalProperties": {
-            @GLEAN_BASE_OBJECT_1_JSON@,
+            @GLEAN_LABELED_GROUP_1_JSON@,
             "additionalProperties": @GLEAN_STRING_LIST_1_JSON@
           }
         },
@@ -121,7 +121,7 @@
         "labeled_enumeration": {
           @GLEAN_BASE_OBJECT_1_JSON@,
           "additionalProperties": {
-            @GLEAN_BASE_OBJECT_1_JSON@,
+            @GLEAN_LABELED_GROUP_1_JSON@,
             "additionalProperties": @GLEAN_ENUMERATION_1_JSON@
           }
         },
@@ -132,7 +132,7 @@
         "labeled_counter": {
           @GLEAN_BASE_OBJECT_1_JSON@,
           "additionalProperties": {
-            @GLEAN_BASE_OBJECT_1_JSON@,
+            @GLEAN_LABELED_GROUP_1_JSON@,
             "additionalProperties": @GLEAN_COUNTER_1_JSON@
           }
         },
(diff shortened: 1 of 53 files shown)


templates/include/glean/glean.1.schema.json
```diff
@@ -73,8 +73,8 @@
       },
       "required": [
         "range",
-	"bucket_count",
```
Dexterp37: nit: should we align `bucket_count` and `values` to `range`?

templates/include/glean/glean.1.schema.json
```diff
@@ -163,216 +378,196 @@
     "metrics": {
       "properties": {
         "boolean": {
+          "$ref": "#/definitions/boolean"
+        },
+        "labeled_boolean": {
```
Dexterp37: Oh, TIL "labeled" is American English :)

validation/glean/baseline.1.all.pass.json
```diff
@@ -25,33 +25,82 @@
...
+    },
     "string": {
       "examples.subcategory.string_example": "\u82f1\u6f22\u5b57\u5178"
     },
+    "labeled_string": {
+      "examples.subcategory.labeled_string_example": {
+        "longer_key_1": "ABCD",
+        "an_even_longer_key_2": "EFGH"
+      }
+    }
```
Dexterp37: Looks like you're missing a comma here

validation/glean/baseline.1.all.pass.json
```diff
@@ -25,33 +25,82 @@
...
     "number": {
       "examples.number_example": 5
     },
+    "labeled_number": {
+      "examples.labeled_number_example": {
+        "a": 42,
+        "b": 58
+      }
+    }
```
Dexterp37: And here

validation/glean/baseline.1.all.pass.json
```diff
@@ -25,33 +25,82 @@
...
         "B",
         "C"
       ]
     },
+    "labeled_string_list": {
+      "examples.labeled_string_list_example": {
+        "a": ["X", "Y", "Z"],
+        "b": ["μ", "γ"]
+      }
+    }
```
Dexterp37: And here

validation/glean/baseline.1.all.pass.json
```diff
@@ -25,33 +25,82 @@
...
     "enumeration": {
       "examples.enumeration_example": "value1"
     },
+    "labeled_enumeration": {
+      "examples.labeled_enumeration_example": {
+        "a": "value1",
+        "b": "value2"
+      }
+    }
```
Dexterp37: And here

validation/glean/baseline.1.all.pass.json
```diff
@@ -25,33 +25,82 @@
...
     "counter": {
       "examples.counter_example": 42
     },
+    "labeled_counter": {
+      "examples.labeled_counter_example": {
+        "a": 0,
+        "b": 68,
+        "__other__": 2
+      }
+    }
```
Dexterp37: Comma ,

templates/include/glean/glean.1.schema.json
mdboom: Yeah.  It's annoying.  But I guess Mozilla products should default to American English?

validation/glean/baseline.1.all.pass.json
mdboom: Weird.  I guess my passing tests aren't actually running this, then.  It would have caught these syntax errors.

templates/include/glean/glean.1.schema.json
mdboom: Weird.  It looked aligned in my editor.  I'll do a tab-to-space conversion and see if that helps.

mdboom:
It looks like `allOf` can mostly be refactored out.

The tricky one is `timing_distribution` which is a `histogram` but adds an additional property for the `time_unit`.  I don't think there's a clean way to do that with the templating system.  I think I'll leave them in for now -- it will be a good discussion point to include @acmiyaguchi and we can decide whether it's worth the effort to support it in the BigQuery transpiler.

mdboom:
@fbertsch: I refactored this to use templating rather than definitions.  The use of `$ref` and definitions are completely gone.

There is still one use of `allOf`.  Of course, this could be manually expanded (losing the "inheritance" of `timespan` on `histogram`) if we need to, but thought it might be a good discussion point first.

fbertsch:
@mdboom sorry for the delay - your new commit reminded me about commenting here. We decided we could use the `allOf` here, but we want to make sure we properly include that test case in the transpiler, just to make sure everything is converted properly. So feel free to include it for the `timespan` metric, or if you want to remove all instances now and come back later that's fine too.

mdboom:
@fbertsch: No problem.  I'm find with taking `allOf` out for now and revisiting if we grow a lot of types based on histogram (which is the only case where it really reduces duplication here).

What (if anything) do I need to do here on the Parquet side of things?

fbertsch:
@mdboom for parquet, we just need another level of maps for those elements, so each `value` field will instead be a `MAP` with the entire existing `key_value` group for that field. E.g. for `counter`:

```
optional group counter (MAP) {
  repeated group key_value {
    required binary key (UTF8);
    required int64 value;
  }
}
```

becomes

```
optional group counter (MAP) {
  repeated group key_value {
    required binary key (UTF8);
    required group value (MAP) {
      repeated group key_value {
        required binary key (UTF8);
        required int64 value;
      }
    }
  }
}
```

We can use the same templating scheme for inheritance of these fields.

mdboom:
Cool.  The parquet changes are added here, and I've removed the "WIP" status.

PR #257, opened 2019-02-08
[glean] Don't allow additionalProperties in ping_info

mdboom:
I feel like `ping_info` should be pretty locked down, since we want to be able to rely on exactly what's in there for every single ping.  This is in contrast to even things in the baseline ping where end-user code can add custom pings to the `metrics` section by design.  This has already been the source of [one bug](https://github.com/mozilla-mobile/android-components/pull/1984#discussion_r254739261) that slipped through the cracks.

Checklist for reviewer:

### DIFF
--- a/schemas/glean/baseline/baseline.2.schema.json
+++ b/schemas/glean/baseline/baseline.2.schema.json
@@ -646,6 +646,7 @@
       }
     },
     "ping_info": {
+      "additionalProperties": false,
       "properties": {
         "app_build": {
           "type": "string"
--- a/schemas/glean/events/events.2.schema.json
+++ b/schemas/glean/events/events.2.schema.json
@@ -646,6 +646,7 @@
       }
     },
     "ping_info": {
+      "additionalProperties": false,
       "properties": {
         "app_build": {
           "type": "string"
--- a/schemas/glean/metrics/metrics.2.schema.json
+++ b/schemas/glean/metrics/metrics.2.schema.json
@@ -646,6 +646,7 @@
       }
     },
     "ping_info": {
+      "additionalProperties": false,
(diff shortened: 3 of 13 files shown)


mreid-moz:
I like the idea too! This makes the exact set of available fields predictable from the schema version. 

One consideration is that it would require a version bump on the client when adding a new field to `ping_info`, which (in the current backend design) would result in new output tables even if the schema is compatible.

Dexterp37:
> One consideration is that it would require a version bump on the client when adding a new field to `ping_info`, which (in the current backend design) would result in new output tables even if the schema is compatible.

Mh, I see. Would that imply that data scientists would need to look at two different tables to join data from the previous and the current version? Does that already happen? Is that something we could live with?

@georgf FYI

fbertsch:
This would be a schema-compatible change so I don't see any reason to bump the table version. Even if the pipeline did bump (because we automate that), no reason the view can't be the union of the schema-compatible tables.

georgf:
With this landed, can we still iterate quickly on the `ping_info` until Fenix launch?
That would be my only concern here.

Otherwise it sounds like a great way to reduce problems going forward!

mdboom:
I don't think it affects the ability to iterate quickly, since adding new fields will be parquet-schema compatible.  What it will do is ensure that the new fields are in the schema and documented as we add them, rather than catching them later and realizing they were incorrect etc. as has happened once in practice already.

mdboom:
Let's rebase this on #259 after that's merged so we can put this change on v2 of the schema.

mdboom:
This change has been moved to `v2` of the schema.  Ready for one final review, then I think this is good to merge.

PR #259, opened 2019-02-08
1525045: Simplify datetime to just have value, not unit

mdboom:
Checklist for reviewer:

### DIFF
--- a/validation/glean/baseline.2.core.pass.json
+++ b/validation/glean/baseline.2.core.pass.json
@@ -1,0 +1,97 @@
+{
+  "$schema": "moz://mozilla.org/schemas/glean/ping/1",
+  "ping_info": {
+    "ping_type": "full",
+    "app_build": "59f330e5",
+    "app_display_version": "1.0.0",
+    "telemetry_sdk_build": "abcdabcd",
+    "client_id": "6ff20eb7-e80d-4452-b45f-2ea7e63547aa",
+    "seq": 1,
+    "start_time": "2018-10-23 11:23:15-04:00",
+    "end_time": "2018-10-23 11:23:15-04:25"
+  },
+  "metrics": {
+    "boolean": {
+      "examples.boolean_example": true
+    },
+    "string": {
+      "examples.string_example": "\u82f1\u6f22\u5b57\u5178"
+    },
+    "number": {
+      "examples.number_example": 5
+    },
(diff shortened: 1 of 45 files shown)


fbertsch:
@mdboom this is not a schema-compatible change :/ we'll need to bump the version. Sound good?

mdboom:
Sure thing.  To confirm: this would involve copying the `datetime` template and all of the downstream (datetime-using) templates to version 2 (and leaving version 1 as it was prior to this change)?  Or do I need to bump the version on all the glean templates?

fbertsch:
@mdboom exactly. So old clients will route to JSON schema v1, and then parquet tables v1; new clients will need to bump their docVersion to hit v2 of both. On the query side we'll need to include both tables to get client counts (depending on how long it takes everyone to update, which given the reference browser population, won't be long).

mdboom:
Cool.  I've redone this so it's all as a bump to version 2.

fbertsch:
Big fan of the changelog. Consider adding something to the README about it.

Should we update the changelog for schema-compatible changes?

mdboom:
> Big fan of the changelog. Consider adding something to the README about it.

Would you just suggest something like "For glean schemas, we have a changelog here...", or are you suggesting a central changelog or "everybody gets a changelog"?

> Should we update the changelog for schema-compatible changes?

That's probably a good idea, too.

validation/glean/baseline.2.core.pass.json
```diff
@@ -0,0 +1,97 @@
+{
+  "$schema": "moz://mozilla.org/schemas/glean/ping/1",
```
Dexterp37: Quick flyby (the rest looks good to me): should this end with `/2`, given the version bump? Or it doesn't matter/is something else?

mdboom: Good catch.  You are right, this should be updated.

fbertsch:
For now, add a bit about generally how to use the changelog, and then point to the glean one in particular. In the future we can roll it out to other pings if we feel like it's been useful here.

We should probably require updates to the changelog for glean ping updates.

mdboom:
Cool.  I've updated the main `README` about the glean `CHANGELOG`.  I also updated the pull request template -- let me know if that's too noisy, since many (most) PRs probably don't need that.

PR #260, opened 2019-02-13
Loosen label regex

mdboom:
This broadens the `label` regex so it can include `.` and `/`.  This is needed for [error tracking](https://bugzilla.mozilla.org/show_bug.cgi?id=1499761).  Error tracking uses labeled counters where the label is the fully-qualified name of the metric associated with the error, and those have `.` in them.  Additionally the metric that caused the error may itself be a labeled metric, therefore it could have a `/` in it.

This is entirely independent of the allowed labels when using labeled metrics directly on the client, which still doesn't allow these characters.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/glean/baseline/baseline.1.schema.json
+++ b/schemas/glean/baseline/baseline.1.schema.json
@@ -120,8 +120,7 @@
               "type": "boolean"
             },
             "propertyNames": {
-              "maxLength": 30,
-              "pattern": "^[a-z_][a-z0-9_]*$",
+              "pattern": "^[a-z_][a-z0-9_.]*$",
               "type": "string"
             },
             "type": "object"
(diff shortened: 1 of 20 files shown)


Dexterp37:
@fbertsch does having the mentioned characters in metrics names have implications for the pipeline side?

fbertsch:
Unfortunately this won't work with Bigquery. The [limitations](must contain only letters (a-z, A-Z), numbers (0-9), or underscores (_), and it must start with a letter or underscore) specifies: `must contain only letters (a-z, A-Z), numbers (0-9), or underscores (_), and it must start with a letter or underscore`.

Can we record this in an alternate way? Maybe with underscores?

Dexterp37:
> Can we record this in an alternate way? Maybe with underscores?

I think this makes sense. @mdboom what do you think?

mdboom:
The limitations are for any keys on maps in BigQuery?  I think we're already violating that with our metric identifiers which have a `.` in them.

mdboom:
As per our meeting discussion, I updated this to only add `.` in labeled metric labels, not `/`.

mdboom:
For reasons of testing, this also includes #270.

mdboom:
Yeah -- there's no limit on the nesting.  (Though maybe there should be).

PR #263, opened 2019-02-15
Remove Uuid from fire-tv metadata

fbertsch:
This field isn't queryable and borks the whole metadata section: ` Parquet type FIXED_LEN_BYTE_ARRAY supported as DECIMAL; got null`.

We'll need to bump the version with this change.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/pocket/fire-tv-events/fire-tv-events.1.parquetmr.txt
+++ b/schemas/pocket/fire-tv-events/fire-tv-events.1.parquetmr.txt
@@ -15,7 +15,6 @@ message fire_tv_events {
     required binary documentId (UTF8);
     required binary geoCity (UTF8);
     required binary geoCountry (UTF8);
-    required fixed_len_byte_array(16) Uuid;
     required int32 Pid;
     required int64 Timestamp;
   }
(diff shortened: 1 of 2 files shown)


whd:
Bumping the parquet table version doesn't map into the docversion-per-table structure that we have for structured ingestion. I think the correct thing to do here is to probably take some downtime on this dataset and reprocess the existing parquet files, while writing new data out with the new schema.

That said, we can also manage fire-tv the way we manage core pings, which appears to be what you're asking for. We've never done this for structured ingestion, but it's not difficult to do. If you'd prefer not to backfill / take downtime I'm ok with doing the management, but updates to schemas for this dataset may be relatively latent compared to other schema deploys.

fbertsch:
@whd Alternatively, we may be able to just drop everything so far. I know this hasn't even hit a pre-release population, everything so far has been testing; and that may make your life a bit easier. Let me check with them.

fbertsch:
@whd I've confirmed with the team that we can drop all existing data for this ping.

whd:
Excellent, I will coordinate this deploy with a drop of the old data, probably some time today.

whd:
This has been deployed and the old data has been dropped. I'm getting `Schema mismatch, metastore schema for row column metadata has 18 fields but parquet schema has 17 fields` from re:dash now, which should be cleared up in the next p2h run.

PR #269, opened 2019-02-21
Implement the new events API schema

Dexterp37:
This adds the glean specific event format to the templates, since it differs from the desktop ones.
We're dropping object/value, renaming `method` to `name` (as we're dropping the "method" special case too), and keeping `extra` as optional.

I've not bumped the ping schema version, as we don't really care about the old data (there should be none, except from the samples). But I'm happy to address this, if needed.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/glean/CHANGELOG.md
+++ b/templates/include/glean/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Version 2 (2019-02-08)
+# Version 2 (2019-02-21)
 
 - `UUID` was removed from the `metadata` section of the parquet schema.
 
@@ -14,7 +14,11 @@
 
 - Changed datetime values from a (value, time_unit) pair to simply the raw
   value.  (Backward-incompatible change).
-  
+
+- Changed the event schema to drop `object`, `value` and renaming
+  `method` to `name`.
+
+
 # Version 1
 
 - Added support for labeled metrics.
--- a/templates/include/glean/glean.2.parquetmr.txt
+++ b/templates/include/glean/glean.2.parquetmr.txt
@@ -267,8 +267,17 @@ message baseline {
   }
   optional group events (LIST) {
     repeated group list {
-      required group element @COMMON_EVENT_1_TXT@
+      required group element (TUPLE) {
+        required int64 timestamp;
+        required binary category (UTF8);
+        required binary name (UTF8);
+        optional group extra (MAP) {
+          repeated group key_value {
+            required binary key (UTF8);
+            optional binary value (UTF8);
+          }
+        }
+      }
     }
   }
-  
 }
--- a/schemas/glean/baseline/baseline.2.parquetmr.txt
+++ b/schemas/glean/baseline/baseline.2.parquetmr.txt
@@ -338,9 +338,7 @@ message baseline {
       required group element (TUPLE) {
         required int64 timestamp;
         required binary category (UTF8);
-        required binary method (UTF8);
-        required binary object (UTF8);
-        optional binary value (UTF8);
+        required binary name (UTF8);
         optional group extra (MAP) {
           repeated group key_value {
             required binary key (UTF8);
--- a/schemas/glean/baseline/baseline.2.schema.json
+++ b/schemas/glean/baseline/baseline.2.schema.json
@@ -23,29 +23,18 @@
             "type": "string"
           },
           {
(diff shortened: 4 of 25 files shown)


templates/include/glean/CHANGELOG.md
```diff
@@ -1,10 +1,17 @@
+# Version 2 (2019-02-21)
```
mdboom: Maybe just update the date, but don't need to provide a new header?

templates/include/glean/event.1.parquetmr.json
```diff
@@ -0,0 +1,23 @@
+{
+  "type": "array",
+  "items": [
+    {
+      "type": "integer",
+      "minimum": 0
+    },
+    {
+      "type": "string"
```
mdboom: Maybe add `"maxLength": 80` here to match the Kotlin code?

templates/include/glean/event.1.parquetmr.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    {
+      "type": "string"
+    },
+    {
+      "type": [ "object", "null" ],
+      "additionalProperties": {
+        "type": [ "string", "null" ]
+      }
```
mdboom: Let's add:

```
"propertyNames": {
  "type": "string",
  "maxLength": 80
}
```

since the keys of the extra properties must be < 80 characters.

Can the extra values actually be `null`?  That doesn't match the existing implementation (but maybe the proposal proposes to change that.  Haven't read it yet :) )

templates/include/glean/event.1.parquetmr.json
```diff
@@ -0,0 +1,23 @@
...
+      "minimum": 0
+    },
+    {
+      "type": "string"
+    },
+    {
+      "type": "string"
+    },
+    {
+      "type": [ "object", "null" ],
```
mdboom: Do we need `null` here?  If we don't have extra, this array will just have 3 items, right?

templates/include/glean/glean.2.parquetmr.txt
```diff
@@ -267,7 +267,7 @@ message baseline {
   }
   optional group events (LIST) {
     repeated group list {
-      required group element @COMMON_EVENT_1_TXT@
+      required group element @GLEAN_EVENT_1_TXT@
```
fbertsch: Since this is now only user in 1 place, we could remove `include/glean/event.1.parquetmr.txt` and insert the schema directly here.

templates/include/glean/event.1.parquetmr.json
```diff
@@ -0,0 +1,23 @@
+{
```
fbertsch: Should this file be `event.1.json.schema`?

templates/include/glean/event.1.parquetmr.json
Dexterp37: This is a great idea. I like it.

templates/include/glean/event.1.parquetmr.json
Dexterp37: No, they can't be - keys for extra properties are now limited to 40 characters.

templates/include/glean/event.1.parquetmr.json
Dexterp37: Both `value` (which had the 80 chars limit) and `object` where dropped. This here is the category name: I'm not sure we have an hardwire limit in kotlin for that

Dexterp37:
Super weird: turns out we're defining `clientId` in `ping_info`, not `client_id`, and since we're not allowing extra properties there we now fail to validate. I'm not sure why this wasn't caught before.. :|

templates/include/glean/event.1.parquetmr.json
mdboom: Cool.  Let's make that 40, then, and I assume your PR to android-components will update that there, too.

templates/include/glean/event.1.parquetmr.json
mdboom: Oh, good point.  No limit on this one (for now).

templates/include/glean/event.1.parquetmr.json
mdboom: Cool.  So just `"type": "object"` here, then.

PR #284, opened 2019-03-08
Error if a JSON validation file does not match the naming convention

trink:
Checklist for reviewer:

For glean changes:


tests/hindsight/run/input/test_json_schemas.lua
```diff
@@ -103,9 +103,12 @@ function process_message()
         local mode = lfs.attributes(fqdn, "mode")
         if mode == "directory" and not namespace:match("^%.") then
             for fn in lfs.dir(fqdn) do
-                local schema, version, test, outcome = fn:match("(.+)%.(%d+)%.([%w_]+)%.(pass).json$")
+                local schema, version, test, outcome = fn:match("(.+)%.(%d+)%.([%w_]+)%.(pass)%.json$")
```
mreid-moz: we should allow hyphens in the `<test_name>` component

mreid-moz:
I added a commit to rename several files and remove some old cruft (namely the jupyter and python files in the `validation` dir).

validation/glean/baseline.1.pass.json
```diff
@@ -1,98 +0,0 @@
-{
```
mdboom: Don't we want to just rename (not remove) this file?

mreid-moz: I wasn't sure if it exercised anything different than the other test file... if it does I can rename it instead

trink: the .all appear to be inclusive so +1 for the removal

mdboom: Ah, actually that's a good catch.  I think they've become more-or-less redundant at this point, and `baseline.1.all.pass.json` is the more complete and better to keep one.

fbertsch: I think we should keep `baseline.1.pass.json` and remove `baseline.1.all.pass.json`. The former is a ping we are actually likely to see, and the second is not. Now that we have: baseline, events, and metrics samples, the `all` seems redundant.

mdboom: `baseline.1.all.pass.json` exercises all of the metric types, but `metrics.1.all.pass.json` (despite the name) doesn't.  Maybe we keep `baseline.1.pass.json` (since it's a realistic example) and remove `metrics.1.all.pass.json` and rename `baseline.1.all.pass.json` to `metrics.1.all.pass.json`.  (Sorry this all got so messy ... :)

fbertsch: @mdboom I agree with that assessment. Best to match what we're actually sending

mreid-moz: Ok, I think I've got the above. I kept the `"experiments"` section from `metrics.1.all.pass.json`, but otherwise replaced its contents with `baseline.1.pass.json`. I also removed the validation examples in `org-mozilla-samples-glean/` since they were repeats of the stuff in the `glean/` dir, which are straight-up template applications of the main `glean` schema.

@mdboom does this look ok now?

mdboom: Yep.  Thanks.  :+1:

mreid-moz:
I think we're good to go here! Thanks all.

PR #285, opened 2019-03-12
Bug 1533954 - Add Pioneer metadata to parquet schemas

whd:
I ran the tests locally and inspecting the parquet output shows this change does in fact add the (already existing) fields as expected.

I also added the geoip field from https://github.com/mozilla-services/lua_sandbox_extensions/pull/420/files.


whd:
To my knowledge, you cannot have comments in parquet schema files. I initially tried to put the template in its own `pioneer-study` directory, but gave up because I didn't want to figure out how to make the cmake variable inclusion at https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/CMakeLists.txt#L37-L45 work with a hyphenated name. I thought about making the directory `pioneer` instead but didn't want to be inconsistent with that either.

trink:
The hyphen should work `@FOO-BAR_FOO-BAR_1_JSON@` was properly replaced in my test.

As for comments, yeah the original syntax didn't support them and I didn't extend it.  Comments could be added if desired but we would have to touch multiple places in the code and beautifier.

whd:
I've moved the template to its own directory.

PR #290, opened 2019-03-15
Add better definitions for array types missing items

### LINKED ISSUE #288: `range` array in histograms should contain a valid type
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/63028f589c9819f03c0c71d642517f0f22953cb5/templates/include/telemetry/histogram.1.schema.json#L19-L21

While converting some of the schemas into avro, I found that there are some schemas that define arrays with ambiguous types. We should be stricter with array types and define the type for the items:

```
range: {
  "type": "array"
  "items": {
    "type": "number"
  }
}
```

acmiyaguchi:
This fixes #288 by adding a fuller type definition for arrays. The missing fields were found by using the following script:

https://gist.github.com/acmiyaguchi/0786d0cff8d79f8a9662377ad3e891b1

Then running the following command at the toplevel:

```
find schemas/ -name "*.schema.json" | xargs -I {} bash -c "echo {}; cat {} | python3 find-arrays.py" 
```

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/telemetry/mainPayload.1.schema.json
+++ b/templates/include/telemetry/mainPayload.1.schema.json
@@ -10,7 +10,11 @@
       "properties": { }
     },
     "childPayloads": {
-      "type": "array"
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": { }
+      }
     },
     "chromeHangs": {
       "type": "object",
--- a/schemas/telemetry/main/main.4.schema.json
+++ b/schemas/telemetry/main/main.4.schema.json
@@ -998,6 +998,10 @@
           "type": "object"
         },
         "childPayloads": {
+          "items": {
+            "properties": {},
+            "type": "object"
+          },
           "type": "array"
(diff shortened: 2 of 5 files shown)


acmiyaguchi:
I've pushed this branch to integration -- there doesn't seem to be a diff in the artifacts folder along the two validation reports.

templates/include/telemetry/mainPayload.1.schema.json
```diff
@@ -10,7 +10,11 @@
       "properties": { }
     },
     "childPayloads": {
-      "type": "array"
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": { }
```
fbertsch: Why the empty properties here?

acmiyaguchi: This seems to be the convention for empty objects in the templates.

PR #291, opened 2019-03-16
Avoid incompatible array and object union type

acmiyaguchi:
I used the following script to find all occurrences of `"type": ["array", "object"]`:

```python
#!/usr/bin/env python3

import json
import sys
from collections import deque

def main():
    # traverse a json document, finding path to leaves that match exactly with the term
    document = json.load(sys.stdin)
    
    search_term = {"object", "array"}
    results = []

    queue = deque([("root", document)])
    while queue:
        name, node = queue.popleft()
        if not isinstance(node, dict):
            continue
        dtype = node.get("type")
        if dtype and isinstance(dtype, list) and search_term.issubset(set(dtype)):
            results.append((name, node))
        for key, value in node.items():
            queue.append((f"{name}.{key}", value))
    
    results = [x[0] for x in results]
    for item in results:
        print(item)

if __name__ == "__main__":
    main()
```

This is saved as `find-array-object.py`.

This one-liner prints out unique errors (I used some form of this to link them to the documents that they came from).
```bash
find schemas/ -name "*.schema.json" | \
    { xargs -I {} bash -c "cat {} | python3 find-array-object.py"; } | \
    sort | uniq
```

```
root.properties.application.properties.addons.properties.activePlugins
root.properties.application.properties.addons.properties.activePlugins.items.properties.mimeTypes
root.properties.environment.properties.addons.properties.activePlugins
root.properties.environment.properties.addons.properties.activePlugins.items.properties.mimeTypes
root.properties.environment.properties.partner.properties.partnerNames
root.properties.environment.properties.system.properties.gfx.properties.monitors
root.properties.payload.properties.UIMeasurements
root.properties.payload.properties.log
root.properties.payload.properties.threadHangStats
```

To verify that the arrays are not actually objects, I ran the following jq expressions on the sampled landfill data.

```bash
find schemas/ -name "*.schema.json" | \
    { xargs -I {} bash -c "cat {} | python3 find-array-object.py"; } | \
    sort | uniq | \
    sed 's/root//g; s/.properties//g; s/.items/ | .[]? | /g; s/$/ | type/' | \
```

Results:
```
bash-3.2$ pwd
/Users/amiyaguchi/Work/mozilla-pipeline-schemas
bash-3.2$ find data/ -type f | xargs jq ".environment.addons.activePlugins | type" | sort | uniq -c
7617 "array"
4528 "null"
bash-3.2$ find data/ -type f | xargs jq ".environment.addons.activePlugins | .[]? | .mimeTypes | type" | sort | uniq -c
6693 "array"
bash-3.2$ find data/ -type f | xargs jq ".environment.partner.partnerNames | type" | sort | uniq -c
8186 "array"
3959 "null"
bash-3.2$ find data/ -type f | xargs jq ".environment.system.gfx.monitors | type" | sort | uniq -c
8172 "array"
3973 "null"
bash-3.2$ find data/ -type f | xargs jq ".payload.UIMeasurements | type" | sort | uniq -c
 962 "array"
11183 "null"
bash-3.2$ find data/ -type f | xargs jq ".payload.log | type" | sort | uniq -c
1026 "array"
11119 "null"
bash-3.2$ find data/ -type f | xargs jq ".payload.threadHangStats | type" | sort | uniq -c
 299 "array"
11846 "null"
```

These fields are not actually a union of array and objects, they are just arrays.

https://gist.github.com/acmiyaguchi/99efdbe9c84d4c990140ad0ea85ba4af

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/telemetry/mainPayload.1.schema.json
+++ b/templates/include/telemetry/mainPayload.1.schema.json
@@ -67,10 +67,11 @@
       "properties": { }
     },
     "log": {
-      "type": [
-        "array",
-        "object"
-        ]
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": { }
+      }
     },
     "simpleMeasurements": {
       "type": "object",
@@ -85,16 +86,18 @@
       "properties": { }
     },
     "threadHangStats": {
-      "type": [
-        "array",
-        "object"
-        ]
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": { }
+      }
     },
     "UIMeasurements": {
-      "type": [
-        "array",
-        "object"
-        ]
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": { }
+      }
     },
     "ver": {
       "type": "integer"
--- a/templates/include/telemetry/mainPayload.1.schema.json
+++ b/templates/include/telemetry/mainPayload.1.schema.json
@@ -67,11 +67,10 @@
       "properties": { }
     },
     "log": {
-      "type": "array",
-      "items": {
-        "type": "object",
-        "properties": { }
-      }
+      "type": [
+        "array",
+        "object"
+        ]
     },
     "simpleMeasurements": {
       "type": "object",
--- a/schemas/telemetry/crash/crash.4.schema.json
+++ b/schemas/telemetry/crash/crash.4.schema.json
@@ -179,10 +179,7 @@
                     "items": {
                       "type": "string"
                     },
-                    "type": [
-                      "array",
-                      "object"
-                    ]
+                    "type": "array"
                   },
                   "name": {
                     "type": "string"
@@ -196,10 +193,7 @@
                 },
                 "type": "object"
               },
-              "type": [
-                "array",
-                "object"
-              ]
+              "type": "array"
             },
             "persona": {
               "type": [
@@ -364,10 +358,10 @@
               ]
             },
             "partnerNames": {
-              "type": [
-                "array",
-                "object"
-              ]
+              "items": {
+                "type": "string"
+              },
+              "type": "array"
             }
           },
           "type": "object"
@@ -815,10 +809,7 @@
                     },
                     "type": "object"
                   },
-                  "type": [
-                    "array",
-                    "object"
-                  ]
+                  "type": "array"
                 }
               },
               "type": "object"
--- a/schemas/telemetry/event/event.4.schema.json
+++ b/schemas/telemetry/event/event.4.schema.json
@@ -183,10 +183,7 @@
                     "items": {
                       "type": "string"
                     },
-                    "type": [
-                      "array",
-                      "object"
-                    ]
+                    "type": "array"
                   },
                   "name": {
                     "type": "string"
@@ -200,10 +197,7 @@
                 },
                 "type": "object"
               },
-              "type": [
-                "array",
-                "object"
-              ]
+              "type": "array"
             },
             "persona": {
               "type": [
@@ -368,10 +362,10 @@
               ]
             },
             "partnerNames": {
-              "type": [
-                "array",
-                "object"
-              ]
+              "items": {
+                "type": "string"
+              },
+              "type": "array"
             }
           },
           "type": "object"
@@ -819,10 +813,7 @@
                     },
                     "type": "object"
                   },
-                  "type": [
-                    "array",
-                    "object"
-                  ]
+                  "type": "array"
                 }
               },
               "type": "object"
--- a/schemas/telemetry/main/main.4.schema.json
+++ b/schemas/telemetry/main/main.4.schema.json
@@ -1133,11 +1133,10 @@
           "type": "object"
         },
         "log": {
-          "items": {
-            "properties": {},
-            "type": "object"
-          },
-          "type": "array"
+          "type": [
+            "array",
+            "object"
+          ]
         },
         "processes": {
           "properties": {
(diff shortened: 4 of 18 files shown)


templates/include/telemetry/mainPayload.1.schema.json
```diff
@@ -63,10 +63,19 @@
...
+        "items": {
+          "type": [
+            {"type": "string"},
+            {"type": "number"},
+            {"type": "string"},
+            {"type": "string"},
+            {"type": "string"}
+          ]
+        }
+      }
```
acmiyaguchi: I'm pretty uncomfortable with this change because I'm not sure how stable this field is. Based on sampled pings from 2019-03-14, they take the form of a tuple like above:

```
$ jq ".payload.log // empty | select(length != 0)" data/telemetry.main.4.ndjson
[
  [
    "EXPERIMENT_ACTIVATION",
    17077,
    "REJECTED",
    "unaccelerated-compositor-process-nightly56@experiments.mozilla.org",
    "endTime"
  ]
]
[
  [
    "EXPERIMENT_ACTIVATION",
    2293,
    "REJECTED",
    "unaccelerated-compositor-process-nightly56@experiments.mozilla.org",
    "endTime"
  ]
]
[
  [
    "EXPERIMENT_ACTIVATION",
    13025,
    "REJECTED",
    "unaccelerated-compositor-process-nightly56@experiments.mozilla.org",
    "endTime"
  ],
  [
    "EXPERIMENT_ACTIVATION",
    3606165,
    "REJECTED",
    "unaccelerated-compositor-process-nightly56@experiments.mozilla.org",
    "endTime"
  ]
]
...
```

@mreid-moz do you have more context around this particular field?

acmiyaguchi:
It's possible that these fields were once objects and their schemas changed to arrays only recently -- I've run into a similar schema mismatch in gcItems.slices:

https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/63028f589c9819f03c0c71d642517f0f22953cb5/templates/include/telemetry/gcItem.1.schema.json#L26-L28

It seems that older versions of firefox use an array version of this field, whereas new versions use a number. This is legitimate and prevents validation failures. This could occur here too.

mreid-moz:
Re: `slices`, there is some more info in this commit: https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/8c0d152eb364de0041c31b1b319777c0f9f0cab1

Unfortunately the data type of this field _did_ change in an incompatible way. I'm not sure how much data we're still receiving in the old form, maybe it's OK to drop the old form?

templates/include/telemetry/mainPayload.1.schema.json
mreid-moz: I don't know much about this field - @badboy do you know where this is recorded in the Firefox source?

badboy: That's .... an [empty array](https://searchfox.org/mozilla-central/rev/7abb9117c8500ed20833746c9f8e800fce3a4688/toolkit/components/telemetry/pings/TelemetrySession.jsm#897) now since I removed it [about exactly a year ago](https://bugzilla.mozilla.org/show_bug.cgi?id=1443614). 

I don't actually know when we removed the collection code, so if you really need that info I can track it down.

acmiyaguchi: I'm curious about the `LOG_ENTRY_MAX_COUNT` variable in the patch to determine if this is actually a tuple or a variable length list with differing types.

Depending on how this is defined, we could get a different Avro type in the end (e.g. `"additionaltems": "string"` for an Array<Array<String>> vs `"additionaltems": false` for an Array<Struct>). Alternatively, removing the log field would be more reflective of our current schema at the expense of dropping the values.

In either case, I wouldn't want pings from older revisions to be dropped due to the validation schema being too strict.

badboy: [Here's the old code](https://github.com/mozilla/gecko-dev/commit/201d19b399bd63af762bc6c8e1fa29a252bdc4e8#diff-3439f9c7661275a769f1816023db8931) of `TelemetryLog`.
The size of the outer array is limited by `LOG_ENTRY_MAX_COUNT`, but the entries itself are pretty much any array of the form: `[String, Number, String*]` (so 2 or more elements), but not a struct.

mreid-moz:
> I'm not sure how much data we're still receiving in the old form, maybe it's OK to drop the old form?

I checked March 18th and there are still several million pings from versions preceding this change (which landed in Firefox 58 afaict) :(

templates/include/telemetry/mainPayload.1.schema.json
acmiyaguchi: I'm going to introduce the proper schema for this in a separate PR, since it's doing a slightly different thing.

mreid-moz:
Good news and bad news - Good: there was no diff in error reports. Bad: the reports were both empty [here](https://806-48513824-gh.circle-artifacts.com/0/app/test-results/0070083.report.json) and [here](https://806-48513824-gh.circle-artifacts.com/0/app/test-results/9194681.report.json).

mreid-moz:
@acmiyaguchi do you know what's going on with the circleci failures here?

acmiyaguchi:
@mreid-moz It looks it's trying to check out the revision corresponding to `91946812203f3d0b6e9bd4bafb0ab01d83d92b66`, but is unable to parse it. That's the full hash if you click on the short-id:
image

That's strange, and hopefully intermittent.

jklukas:
Perhaps the failure _was_ intermittent? @mreid-moz tried pushing the code to an upstream branch and it still failed, but I tried the same a few minutes later and it succeeded, so this PR now shows as passing tests. So, I don't understand what was happening and whether we should expect similar behavior in the future.

PR #294, opened 2019-03-21
Bug 1526072 - Add fields for clobber builds and resource utilization to build telemetry

chmanchester:
This pr adds two fields to built system telemetry. datareview was granted in bug 1526072.

### DIFF
--- a/schemas/eng-workflow/build/build.1.parquetmr.txt
+++ b/schemas/eng-workflow/build/build.1.parquetmr.txt
@@ -9,6 +9,10 @@ message build {
       required binary element (UTF8);
     }
   }
+  optional group build_attrs {
+    optional boolean clobber;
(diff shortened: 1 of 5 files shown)


chmanchester:
Hi @acmiyaguchi, thanks for taking a look. I added a validation document as suggested. Are the on-PR checks here sufficient to test this? I haven't run the tests locally.

chmanchester:
Hi @acmiyaguchi, thanks for merging! At what point will it be safe to start submitting these new fields? Do we need to wait for a merge to master, or for another process to complete?

acmiyaguchi:
I'm not sure how often schemas are reloaded after merging. However, the new schema is backwards compatible (i.e. the pings won't be rejected), so you can start sending them now. 

@whd, do you know how frequently the schemas updated in production?

PR #301, opened 2019-03-27
Add the optional `android_sdk_version` to glean pings

Dexterp37:
Please note that the new `android_sdk_version` is optional, as this is an Android specific field.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/glean/glean.1.schema.json
+++ b/templates/include/glean/glean.1.schema.json
@@ -26,6 +26,7 @@
         "first_run_date": @GLEAN_DATETIME_1_JSON@,
         "os": @GLEAN_STRING_1_JSON@,
         "os_version": @GLEAN_STRING_1_JSON@,
+        "android_sdk_version": @GLEAN_STRING_1_JSON@,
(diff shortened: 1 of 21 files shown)


templates/include/glean/glean.1.schema.json
```diff
@@ -26,6 +26,7 @@
         "first_run_date": @GLEAN_DATETIME_1_JSON@,
         "os": @GLEAN_STRING_1_JSON@,
         "os_version": @GLEAN_STRING_1_JSON@,
+        "android_sdk_version": @GLEAN_STRING_1_JSON@,
```
mdboom: Strictly speaking, this is always an integer, right?

I don't think there's much value in enforcing that at this level, just wanted to flag it in case there might be...

Dexterp37: Good call. I'll make this an int.

Dexterp37: Actually, as we discussed on Slack, we don't have the "number" plain type in glean, so.. let it be a string :)

PR #305, opened 2019-03-28
Convert glean `event` schema from array to object

travis79:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/glean/event.1.schema.json
+++ b/templates/include/glean/event.1.schema.json
@@ -1,17 +1,18 @@
 {
-  "type": "array",
-  "items": [
-    {
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "timestamp": {
       "type": "integer",
       "minimum": 0
     },
-    {
+    "category": {
       "type": "string"
     },
-    {
+    "name": {
       "type": "string"
     },
-    {
+    "extra": {
       "type": "object",
       "additionalProperties": {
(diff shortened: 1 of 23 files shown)


templates/include/glean/event.1.schema.json
```diff
@@ -1,17 +1,16 @@
 {
-  "type": "array",
-  "items": [
-    {
-      "type": "integer",
-      "minimum": 0
+  "type": "object",
+  "properties": {
```
mdboom: Let's add `"additionalProperties": false` here so these objects can't contain anything extra.  If we add something in the future, we'll want to add that explicitly and I *think* having something additional should be treated as an error.

But I also defer to @fbertsch on that question.

travis79: `"additionalProperties": false` added

templates/include/glean/event.1.schema.json
```diff
@@ -1,17 +1,17 @@
...
-  "type": "array",
-  "items": [
-    {
-      "type": "integer",
-      "minimum": 0
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "timestamp": {
+      "type": "integer"
```
fbertsch: We seem to have lost the `"minimum": 0` on `timestamp`

travis79: Done!

PR #310, opened 2019-04-10
Add `Headless` to Telemetry Environment.

chutten:
Fixes bug 1536175

Might be nice to know if the ping's coming from a headless Firefox.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/crash/crash.4.schema.json
+++ b/schemas/telemetry/crash/crash.4.schema.json
@@ -641,6 +641,12 @@
                     "null"
                   ]
                 },
+                "Headless": {
+                  "type": [
+                    "boolean",
+                    "null"
+                  ]
+                },
                 "LowEndMachine": {
                   "type": "boolean"
                 },
--- a/schemas/telemetry/event/event.4.schema.json
+++ b/schemas/telemetry/event/event.4.schema.json
@@ -645,6 +645,12 @@
                     "null"
                   ]
                 },
+                "Headless": {
+                  "type": [
+                    "boolean",
+                    "null"
+                  ]
+                },
                 "LowEndMachine": {
                   "type": "boolean"
                 },
--- a/schemas/telemetry/main/main.4.schema.json
+++ b/schemas/telemetry/main/main.4.schema.json
@@ -645,6 +645,12 @@
                     "null"
                   ]
                 },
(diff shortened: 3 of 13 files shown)


templates/include/telemetry/environment.1.parquetmr.txt
```diff
@@ -73,6 +73,7 @@ optional group environment {
       }
       optional boolean DWriteEnabled;
       optional boolean D2DEnabled;
+      optional boolean Headless;
```
mreid-moz: Drive-by: I think this change is not compatible with schema evolution in parquet (change to the fields in a nested struct).

chutten: Is there a way I can test that assumption?

(( or I could just drop it. It's only a nice-to-have ))

mreid-moz: I'm not sure if there's an easy way to test it, so +1 for dropping :-/

chutten:
Alrighty, it's been dropped. Ready for review.

validation/telemetry/event.4.headless-sample.fail.json
```diff
@@ -0,0 +1,382 @@
...
+        "system": {
+          "model": null,
+          "revision": null
+        }
+      },
+      "gfx": {
+        "D2DEnabled": null,
+        "DWriteEnabled": null,
+        "ContentBackend": "Skia",
+        "Headless": 0,
```
acmiyaguchi: Could you pare down this validation ping? One thought is to use the `main.4.min.pass` document as a base to avoid the extra boilerplate.

https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/151c1dbbac66f7c52285e32aa7edc494957e664a/validation/telemetry/main.4.min.pass.json#L1-L16

The cases themselves look good otherwise.

chutten: Alrighty. It took a little finagling since introducing the optional `environment` means you need to add some required things (like almost the entirety of `environment.build`), but I think this is now the minimal example of a passing- and failing-due-to-`Headless` schema validation set.

chutten:
Validation set is now minimal. Good for another look

PR #312, opened 2019-04-17
Add schemas for ping metadata in the GCP pipeline

jklukas:
This codifies the metadata structure proposed in [Proposal: Ping Metadata in GCP](https://docs.google.com/document/d/1AsshL_QS1qRVXcHW5lz0jZeM-UKIWFl1oV3XJz7Xewo/edit?ts=5cb4b194#heading=h.5x0d5h95i329).

These two schemas don't define tables on their own; rather they are intended to be merged in with per-doctype json schemas in order to produce BigQuery schemas that include pipeline metadata.


schemas/metadata/structured-ingestion/structured-ingestion.1.schema.json
```diff
@@ -0,0 +1,82 @@
+{
```
jklukas: It's not clear to me where best to put this schema. The "metadata" namespace seemed potentially least confusing and I'm assuming that the other schemas in that folder aren't intended to be tables in BQ (which is the case for this schema as well).

jklukas:
Leaving this in draft form right now so that this can be included in the conversation around the proposal. Once the proposal is closed and accepted, I will add tests and open for final review.

schemas/metadata/structured-ingestion/structured-ingestion.1.schema.json
mreid-moz: assumption confirmed - the other schemas in here are used for stuff in the metadata bucket (`sources.json` which is used by the Dataset API) and credentials files which are used by a few applications to connect to databases.

jklukas: Would people or tooling freak out if I created a `templates/metadata/README.md` to document the intention of the schemas in this namespace?

mreid-moz: I don't think so

fbertsch: :mdboom put a changelog into the glean namespace. That could also be a useful feature here as we change/add metadata in the future. And +1 on the readme.

jklukas: The structure here is complete and should be matching the implementation in https://github.com/mozilla/gcp-ingestion/pull/542

I'll take this out of draft mode on Monday after I add the README/changelog.

jklukas: Added.

CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ endforeach()
 file(GLOB_RECURSE SCHEMAS LIST_DIRECTORIES FALSE RELATIVE "${CMAKE_SOURCE_DIR}/templates" "${CMAKE_SOURCE_DIR}/templates/*")
 foreach(ITEM IN ITEMS ${SCHEMAS})
   if(NOT ITEM MATCHES "^(include/|.*/\\.[^/])" )
-    string(REGEX MATCH "/([-a-zA-Z0-9]+\\.[1-9][0-9]*\\.schema\\.json|[-a-zA-Z0-9_]+\\.[1-9][0-9]*\\.parquetmr\\.txt)$" MATCH ${ITEM})
+    string(REGEX MATCH "/([-a-zA-Z0-9]+\\.[1-9][0-9]*\\.schema\\.json|[-a-zA-Z0-9_]+\\.[1-9][0-9]*\\.parquetmr\\.txt|README.*)$" MATCH ${ITEM})
```
jklukas: @trink Do you see any potential consequences of allowing a README.md file to live inside a `templates` subdirectory and propagate to `schemas`?

fbertsch: On my end, this will be fine; anything in the `schemas/metadata` will not be deployed to the GCP MPS branch. We can special-case these files if we do want to include them there in the future.

templates/metadata/README.md
```diff
@@ -0,0 +1,15 @@
+# Metadata schemas
+
+The schemas in this directory are not intended to correspond directly to pings,
+but rather to describe other metadata structures.
+
+## Ingestion metadata
+
+The `structured-ingestion` and `telemetry-ingestion` schemas define metadata
```
fbertsch: This is a _bit_ confusing to read. Maybe something like:

```
The `structured-ingestion` and `telemetry-ingestion` schemas define metadata
fields that are be added to ping schemas. The GCP pipeline records various metadata
about each ping that comes through, and inserts that metadata into the ping payload.
These schemas define the structure of those new fields. The Bigquery tables are then
updated to include this additional metadata as well.
```

schemas/metadata/README.md
```diff
@@ -0,0 +1,15 @@
...
+The schemas in this directory are not intended to correspond directly to pings,
+but rather to describe other metadata structures.
+
+## Ingestion metadata
+
+The `structured-ingestion` and `telemetry-ingestion` schemas define metadata
+fields that should be added in to ping schemas to define the schema of BigQuery
+tables to hold the pings. The GCP pipeline records various metadata about each
+ping that comes through and inserts them into the ping payloads according to
+these schemas.
```
jklukas: @fbertsch Is this version more clear?

schemas/metadata/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Ingestion metadata
+
+The GCP pipeline captures various metadata about each incoming ping
+and inserts that metadata into the ping payloads following the structure
+defined in the `structured-ingestion` and `telemetry-ingestion` schemas
+that live in this directory.
+Each destination table in BigQuery should stay updated to contain not only the fields from
+the JSON schema for that table's document type, but also all the fields
+defined in the relevant metadata JSON schema.
```
jklukas: @fbertsch - Is this version more clear?

fbertsch: 👍

CMakeLists.txt
trink: Do you want the readme in the artifact output?  You, would have to check the down stream consumers to make sure they can handle it (I don't believe it would be an issue).  Otherwise, you could remove the else clause and only the schemas will be output.

jklukas: Added a README-specific `elseif` clause to specifically ignore READMEs without dropping other files in the future.

PR #314, opened 2019-04-18
Consider speed as optional in parquet output

### LINKED ISSUE #313: output.telemetry_telemetry_pre_account_4_parquet_s3 process_message returned: -1 column 'speedMHz' is required
/CC @badboy 

The pre-account ping schemas were deployed recently and I'm seeing occasional errors of the type above.  It looks like this is because the json schema has the field as required but nullable, but the parquet schema has it as required. I'm a little hazy as to null vs optional, @trink would know more, but making the field optional in parquet is probably sufficient to fix the issue. You should be able to reproduce this in the tests by making the [field](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/validation/telemetry/pre-account.4.sample.pass.json#L56) null.

https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/telemetry/pre-account/pre-account.4.schema.json#L89-L93
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/telemetry/pre-account/pre-account.4.parquetmr.txt#L35-L37

badboy:
Fixes #313

Checklist for reviewer:

### DIFF
--- a/schemas/telemetry/pre-account/pre-account.4.parquetmr.txt
+++ b/schemas/telemetry/pre-account/pre-account.4.parquetmr.txt
@@ -23,17 +23,17 @@ message pre-account {
   required binary id (UTF8);
   required group environment {
     required group settings {
-      required binary locale (UTF8);
+      optional binary locale (UTF8);
     }
     required group system {
-      required int64 memoryMB;
+      optional int64 memoryMB;
       required group os {
-        required binary name (UTF8);
-        required binary version (UTF8);
-        required binary locale (UTF8);
+        optional binary name (UTF8);
+        optional binary version (UTF8);
+        optional binary locale (UTF8);
       }
       required group cpu {
-        required int64 speedMHz;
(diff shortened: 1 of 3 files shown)


mreid-moz:
Please also add the rendered stuff in `schemas/` (sorry!)

badboy:
D'oh! No idea how I managed to _not_ commit those.

badboy:
Can this land? @fbertsch or @mreid-moz?

whd:
I'll roll this out on Monday.

whd:
I've rolled this out and it appears to work from re:dash (athena). If the resulting schema update via p2h update breaks access from some frontend I will try to sort it out with :robotblake. (Aside: the majority of errors for this dataset are because of the unrelated known issue at https://bugzilla.mozilla.org/show_bug.cgi?id=1547563).

PR #315, opened 2019-04-19
glean: 1529226: Make client_id optional

mdboom:
Use case: custom pings need to be able to be sent without the client_id to preserve privacy.

@fbertsch: Does this require a version bump?

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/glean/glean.1.schema.json
+++ b/templates/include/glean/glean.1.schema.json
@@ -35,7 +35,6 @@
         "app_build",
         "app_display_version",
         "architecture",
-        "client_id",
         "device_manufacturer",
         "device_model",
         "first_run_date",
--- a/schemas/glean/baseline/baseline.1.parquetmr.txt
+++ b/schemas/glean/baseline/baseline.1.parquetmr.txt
@@ -23,7 +23,7 @@ message baseline {
     required binary app_display_version (UTF8);
     optional binary app_channel (UTF8);
     required binary architecture (UTF8);
-    required binary client_id (UTF8);
+    optional binary client_id (UTF8);
     required binary device_manufacturer (UTF8);
     required binary device_model (UTF8);
     required binary first_run_date (UTF8);
--- a/schemas/glean/baseline/baseline.1.schema.json
+++ b/schemas/glean/baseline/baseline.1.schema.json
@@ -56,7 +56,6 @@
         "app_build",
         "app_display_version",
         "architecture",
-        "client_id",
         "device_manufacturer",
         "device_model",
         "first_run_date",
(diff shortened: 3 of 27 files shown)


Dexterp37:
@mdboom @fbertsch would this cause data loss?

templates/include/glean/glean.1.schema.json
```diff
@@ -35,7 +35,6 @@
         "app_build",
         "app_display_version",
         "architecture",
-        "client_id",
```
Dexterp37: I wonder if we could make this optional based on the ping type. We definitely want this to be in our owned pings. This could still be optional in the Parquet schema.

I have no clue if this makes sense at all :)

mdboom: It could be done with an `if`/`then` block in JSON schema 7.  I don't think we're using that draft in the pipeline though.

jklukas:
> Does this require a version bump?

I don't see any technical reason that changing a field from required to optional would require a version bump. Any ping that's valid before this change would continue to be valid after the change.

That said, if a version bump were needed, this would be a good time to consider whether client_id could be moved to a top-level field rather than nested in client_info. client_id could potentially be useful as a clustering field in BigQuery, but BQ can only cluster on top-level fields.

mdboom:
> client_id could potentially be useful as a clustering field in BigQuery, but BQ can only cluster on top-level fields.

Hmm...  That's interesting.  I wonder if it's worth a look for that kind of thing in general.  I don't think that's something that has been considered in any of the design.

jklukas:
> Hmm... That's interesting. I wonder if it's worth a look for that kind of thing in general. I don't think that's something that has been considered in any of the design.

This is the only field that surfaced as a potential problem in the recent [Ping metadata proposal](https://docs.google.com/document/d/1AsshL_QS1qRVXcHW5lz0jZeM-UKIWFl1oV3XJz7Xewo/edit?ts=5cb4b194). I feel fairly confident that sample_id is the only field that's going to be problematic in this way. We already know we're going to have to dig into the ping contents to get the source data for making top-level `normalized_*` fields.

fbertsch:
Offhand I'm not sure what clustering by `client_id` would buy us, since it is hard to do predicate push-down on `client_id`. Maybe finding a single `client_id` would be faster but that's not a common query pattern.

Let's leave it out for now, and we can always pull it in later if it's relevant.

templates/include/glean/glean.1.schema.json
fbertsch: This sounds like a nice feature to include in the `glean_parser`, but maybe not worth it here.

PR #326, opened 2019-05-01
Add a new item called `type` to `hdd` object

sefeng211:
This is the schema change for https://bugzilla.mozilla.org/show_bug.cgi?id=1533861

Checklist for reviewer:

For glean changes:


templates/include/telemetry/environment.1.schema.json
```diff
@@ -656,6 +656,12 @@
                     "string",
                     "null"
                   ]
+                },
+                "type": {
+                  "type": [
+                    "string",
+                    "null"
+                  ]
```
acmiyaguchi: It looks like this can be an enumerated type, i.e.

```json
enum: ["SSD", "HDD"]
```

sefeng211: Thanks for pointing this out! Updated the PR. 

I kept the `type` as 
```
"type": [
  "string",
  "null"
]
```

Would `null` still work?

acmiyaguchi: `null` values should still be valid because nothing has changed other than limiting the the values of `string` to the values contained in `enum`. If you aren't sure, I would encourage you to explicitly write another validation test that includes null (and missing!) values for the new column you've added.

sefeng211: Thanks! Added two more tests, for null and empty string.

validation/telemetry/main.4.hdd_type_null.fail.json
```diff
@@ -0,0 +1,43 @@
...
+      "vendor": "Mozilla",
+      "platformVersion": "62.0a1",
+      "xpcomAbi": "x86_64-gcc3"
+    },
+    "partner": {},
+    "settings": {},
+    "system": {
+      "hdd": {
+        "profile": {
+          "type": null
```
acmiyaguchi: I actually didn't know this would fail. It looks like there needs to be a slight modification for this to work properly.

templates/include/telemetry/environment.1.schema.json
```diff
@@ -656,6 +656,16 @@
...
                   ]
+                },
+                "type": {
+                  "type": [
+                    "string",
+                    "null"
+                  ],
+                  "enum": [
+                    "SSD",
+                    "HDD"
```
acmiyaguchi:
```suggestion
                    "HDD",
                    null
```

This should allow null values for this field, while also limiting the values to the enumerated set.

validation/telemetry/main.4.hdd_type_null.fail.json
sefeng211: Ah yep, `null` should pass. It is fixed now

templates/include/telemetry/environment.1.schema.json
sefeng211: PR updated!

sefeng211:
The patch is reached `mozilla-central`, do you mind look it again @acmiyaguchi ? Thanks!

PR #327, opened 2019-05-01
Rename spocs-fills to spoc-fills for Activity Stream

ncloudioj:
Checklist for reviewer:

r? @jklukas


jklukas:
Discussed offline with @ncloudioj. Although this change would normally require a version bump, the GCP pipeline is not yet in production and there is not yet any data in the table, so it's easy for us to make the manual fixups to BQ tables, etc. to handle this change without bumping the version.

PR #328, opened 2019-05-02
Make GCP sumbmission_timestamp datetime

fbertsch:
This has been added to the jsonschema-transpiler and should fix our schemas deploys.

There's an issue in that we seem to be using v4 of json schemas for testing in this repository. The package we use in beam can use 4, 6, or 7; we'll need to ensure that it correctly verifies with v6.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/metadata/structured-ingestion/structured-ingestion.1.schema.json
+++ b/schemas/metadata/structured-ingestion/structured-ingestion.1.schema.json
@@ -102,7 +102,7 @@
     },
     "submission_timestamp": {
       "description": "Time when the ingestion edge server accepted this message",
-      "type": "string"
+      "type": "date-time"
     }
   },
   "required": [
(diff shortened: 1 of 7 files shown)


jklukas:
> we seem to be using v4 of json schemas for testing in this repository

I was under the impression v4 was the target version of schemas we supported. I'm not opposed to upgrading, but it will take some care to do so safely. There are some breaking changes: https://json-schema.org/draft-06/json-schema-release-notes.html

If we're going to upgrade, we may want to go all the way to the current v7.

fbertsch:
Closing this, pending some decisions.

jklukas:
Note that in the pipeline, we do schema validation _before_ merging metadata into the payload. We never validate the merged payload with the merged schema that includes metadata. I think that's reasonable since we a limited team has control over the Dataflow jobs and it seems feasible to handle validation of that code in unit tests rather than having to validate the merged payloads.

Considering that, we may be able to side-step the json schema version issue for now.

PR #329, opened 2019-05-02
Make submission_timestamp datetime formatted

fbertsch:
Okay, this is actually v4 compatible. I made a mistake in that `date-time` is not a type but a format.

Checklist for reviewer:

For glean changes:


fbertsch:
However, it looks like rapid-json isn't actually validating the date-time format.

jklukas:
> However, it looks like rapid-json isn't actually validating the date-time format.

The json spec for validators explicitly states that support for validating formats is optional. For the case of submission_timestamp, which is injected by the pipeline, I don't think being able to validate the format is a concern.

If we want to support datetime types more generally, we'll need to figure out how to validate them, but I don't see that as a blocker here.

PR #331, opened 2019-05-07
Add mobile/activation ping

jklukas:
As described in the [Fennece Mobile Activation Ping design doc](https://docs.google.com/document/d/1D-0m3TRvIAOzGtkBNjaddyVr7baE7uEhiBUFyVQr2pg).

### DIFF
--- a/templates/mobile/activation/activation.1.schema.json
+++ b/templates/mobile/activation/activation.1.schema.json
@@ -43,15 +43,13 @@
       "type": "string"
     },
     "app_name": {
-      "enum": [
-        "Fennec"
-      ]
+      "type": "string"
     },
     "channel": {
       "type": "string"
     }
   },
-  "required": [ "manufacturer", "model", "locale", "os", "osversion" ],
+  "required": [ "manufacturer", "model", "locale", "os", "osversion", "created", "tz", "app_name", "channel" ],
   "oneOf": [
     { "required": [ "identifier" ] },
     { "required": [ "client_id" ] }
--- a/templates/mobile/activation/activation.1.schema.json
+++ b/templates/mobile/activation/activation.1.schema.json
@@ -1,0 +1,59 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "activation",
+  "properties": {
+    "identifier": {
+      "type": "string",
+      "description": "hashed version of Google ID"
+    },
+    "client_id": {
+      "type": "string",
+      @COMMON_PATTERN_UUID_1_JSON@,
+      "description": "telemetry client_id, included only if identifer is empty"
+    },
+    "manufacturer": {
+      "type": "string",
+      "description": "Build.MANUFACTURER"
+    },
+    "model": {
+      "type": "string",
+      "description": "Build.MODEL"
+    },
+    "distribution_id": {
+      "type": "string"
+    },
+    "locale": {
+      "type": "string",
+      "description": "application locale, such as 'en-US'"
+    },
+    "os": {
+      "type": "string"
+    },
+    "osversion": {
+      "type": "string"
+    },
+    "created": {
+      "type": "string"
+    },
+    "profile_date": {
+      "type": "integer"
+    },
+    "tz": {
+      "type": "string"
+    },
+    "app_name": {
+      "enum": [
+        "Fennec"
+      ]
+    },
+    "channel": {
+      "type": "string"
+    }
+  },
+  "required": [ "manufacturer", "model", "locale", "os", "osversion" ],
+  "oneOf": [
+    { "required": [ "identifier" ] },
+    { "required": [ "client_id" ] }
+  ]
+}
--- a/schemas/mobile/activation/activation.1.schema.json
+++ b/schemas/mobile/activation/activation.1.schema.json
@@ -14,9 +14,7 @@
   ],
   "properties": {
     "app_name": {
-      "enum": [
-        "Fennec"
-      ]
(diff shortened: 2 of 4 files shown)


templates/mobile/activation/activation.1.schema.json
```diff
@@ -0,0 +1,55 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "activation",
+  "properties": {
+    "identifier": {
```
jklukas: We don't include a "v" property (as exists in the "core" ping) because this is a modern "structure ingestion" ping (the namespace is not telemetry). For structured ingestion, we capture the schema version as one of the URI parameters, so we don't need to include it in the body of the message.

templates/mobile/activation/activation.1.schema.json
```diff
@@ -0,0 +1,55 @@
...
+      "description": "Build.MANUFACTURER"
+    },
+    "model": {
+      "type": "string",
+      "description": "Build.MODEL"
+    },
+    "device": {
+      "type": "string"
+    },
+    "distribution_id": {
```
jklukas: Note use of snake case here rather than camelCase `distributionId`. We have historical pings that use camelCase field names, but snake case is the modern standard and I don't think we have any technical reason to keep this consistent with the core ping schema.

templates/mobile/activation/activation.1.schema.json
```diff
@@ -0,0 +1,55 @@
...
+      "type": "string"
+    },
+    "locale": {
+      "type": "string",
+      "description": "application locale, such as 'en-US'"
+    },
+    "os": {
+      "type": "string"
+    },
+    "osversion": {
```
jklukas: We do have some special handling in the GCP pipeline to find core-ping-style "os" and "osversion" fields, so we leave this name be rather than changing it to "os_version".

templates/mobile/activation/activation.1.schema.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "tz": {
+      "type": "string"
+    }
+  },
+  "required": [ "manufacturer", "model", "device", "locale", "os", "osversion" ],
+  "oneOf": [
+    { "required": [ "identifier" ] },
+    { "required": [ "client_id" ] }
+  ]
```
jklukas: This "oneOf" section enforces the requirement that we must not have both identifier and client_id populated in a single ping. The ping will fail schema validation if it contains both. The data will still be written to "landfill", a restricted-access dump of raw pings we use for backfilling in case of pipeline failures, but the ping will never make it downstream to any parquet or BigQuery outputs that analysts would have access to.

templates/mobile/activation/activation.1.schema.json
```diff
@@ -0,0 +1,54 @@
...
+      "type": "string"
+    },
+    "profile_date": {
+      "type": "integer"
+    },
+    "tz": {
+      "type": "string"
+    }
+  },
+  "required": [ "manufacturer", "model", "device", "locale", "os", "osversion" ],
```
jklukas: It's unclear to me which of these fields we should mark as required, so I've generally copied over the list of required fields from the core ping schema.

If one of these required fields is missing, the ping will be rejected in the pipeline and not end up in the dataset.

templates/mobile/activation/activation.1.schema.json
```diff
@@ -0,0 +1,62 @@
...
+      "type": "integer"
+    },
+    "tz": {
+      "type": "string"
+    },
+    "app_name": {
+      "enum": [
+        "Fennec"
+      ]
+    },
```
jklukas: The "enum" construction here means any value other than "Fennec" will cause the ping to fail validation and be rejected.

We could drop this `app_name` field entirely if we never expect this ping to be used by any other application.

jklukas: Changed this back to just a plain string. Seems safer that way.

templates/mobile/activation/activation.1.schema.json
jklukas: Discussed with :vbaicu and updated this list.

templates/mobile/activation/activation.1.schema.json
sunahsuh: That's pretty cool that we can specify this at the schema level :D

templates/mobile/activation/activation.1.schema.json
```diff
@@ -0,0 +1,57 @@
...
+      "description": "application locale, such as 'en-US'"
+    },
+    "os": {
+      "type": "string"
+    },
+    "osversion": {
+      "type": "string"
+    },
+    "created": {
+      "type": "string"
```
sunahsuh: Does our schema matching implementation allow the use of `format`? If so, seems like this could use `date-time`

jklukas: json schema v4 does allow specifying "format": "date-time". The validator that hindsight uses does not support checking formats, but the one we're using in the GCP pipeline does. I think there's a good possibility this ping will be able to skip the old pipeline, so perhaps this is indeed worth specifying here.

jklukas: Per the client-side implementation (https://phabricator.services.mozilla.com/D29668), this string is going to be in the format "YYYY-MM-DD" without a time within the day. JSON schema v7 includes a "data" format that corresponds to this, but the "date-time" of v4 is limited to strings that include both date and time. So going to leave this as a simple string.

PR #333, opened 2019-05-10
Integer tz for mobile/activation

jklukas:
This makes mobile/activation consistent with existing pings (core
and mobile events) that have an integer "tz" field given in minutes.

This is an incompatible change, but since it's a new ping that's not yet deployed, we can simply delete the existing BQ table before merging and let the machinery recreate it.


jklukas:
I've deleted the BQ table, so this should be mergeable without causing errors downstream.

PR #268, opened 2019-02-20
Add a schema for sync pings

mreid-moz:
As part of #262 we need a schema for `sync` pings. 

I started by creating a new schema based on [the code that submits pings](https://searchfox.org/mozilla-central/source/services/sync/modules/telemetry.js#490) and the [`sync` ping documentation](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/sync-ping.html), then updated it to use [the `payload` schema available in `mozilla-central`](https://hg.mozilla.org/mozilla-central/annotate/c2593a3058afdfeaac5c990e18794ee8257afe99/services/sync/tests/unit/sync_ping_schema.json#l)

This raises a few questions about how we want to handle externally-defined schemas more generally, and whether we want to make the changes in 79843ad in mozilla-central as well.

Checklist for reviewer:

- [x] Scan the PR and verify that no changes (particularly to `.circleci/config.yml`) will cause environment variables (particularly credentials) to be exposed in test logs
- [x] Trigger the `integration` CI test by pushing this revision [as discussed in the README](https://github.com/mozilla-services/mozilla-pipeline-schemas#packaging-and-integration-tests-optional)

Issue #262, opened 2019-02-15
Sherlock Holmes and the case of the missing schemas


whd:
@mreid-moz for the title

Here's a list of every known ping type, more or less. For GCP we require schemas to decode submissions. I'm guessing at least some of these pings (e.g. loop) we no longer care about and can therefore ignore, but there are definitely some (e.g. sync) that we need to create a schema for. The current plan is to create a GCP testing branch to add these new schemas to, and backport those schemas into AWS if/when necessary. I don't have volume estimates for a lot of these (essentially those that we also don't have schemas for, sadly) but with @trink's upcoming auto-bootstrapped doctype monitoring we might be able to get that information automatically.

There's also nothing as to `docVersion` in the list I generated, which may end up being relevant for pings with multiple `docVersions`.

In terms of prioritization, I've made a first pass by **emboldening** pings that I think are probably important and _italicizing_ pings that we probably don't care about anymore.

  - [ ] **activation**
  - [ ] addon-install-blocked
  - [ ] advancedtelemetry
  - [ ] android-anr-report
  - [X] anonymous
  - [ ] appusage
  - [ ] _b2g-installer-device_
  - [ ] _b2g-installer-flash_
  - [ ] **bhr**
  - [X] block-autoplay
  - [ ] certificate-checker
  - [X] core
  - [X] crash
  - [ ] **deletion**
  - [ ] deployment-checker
  - [ ] _disableSHA1rollout_
  - [X] downgrade
  - [X] event
  - [ ] **first-shutdown**
  - [ ] _flash-shield-study_
  - [ ] flash-video
  - [X] focus-event
  - [X] frecency-update
  - [ ] ftu
  - [X] health
  - [X] heartbeat
  - [ ] idle-daily
  - [ ] _loop_
  - [X] main
  - [ ] malware-addon-states
  - [X] mobile-event
  - [X] mobile-metrics
  - [X] modules
  - [X] new-profile
  - [X] optout
  - [ ] outofdate-notifications-system-addon
  - [X] pioneer-study
  - [X] saved-session
  - [ ] **searchvol**
  - [ ] **searchvolextra**
  - [X] shield-icq-v1
  - [X] shield-study
  - [X] shield-study-addon (sort of)
  - [X] shield-study-error
  - [ ] _shield-study-icq_
  - [ ] **sync**
  - [ ] system-addon-deployment-diagnostics
  - [X] testpilot
  - [ ] _testpilottest_
[...]

PR #334, opened 2019-05-16
Add default schemas for most known docTypes

whd:
The default schema is taken from [here](https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/moz_telemetry/io_modules/decoders/moz_ingest/telemetry.lua#L87-L96) but with the empty required properties removed to [match our meta schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/tests/hindsight/jsonschema.4.json#L118).

I generated this PR via [this gist](https://gist.github.com/whd/14630dcfb5167115eff5086dfc3f601d). See also for the full list of doctype / docversions. It may be instructive to look at pings for which we are missing a docversion in some cases (crash, event, first_shutdown, health, main, modules, new_profile, shield_study_addon, update) and verify it's ok to throw those data away as errors in future.

Shield studies schema handling has always been a strange case and I don't know how we want to handle those, but this PR should in theory do nothing on the AWS pipeline because this default schema is already being applied there.

I decided not to check `UNKNOWN`/`OTHER` bucket objects for further potential acceptable
doctypes; the volume is sufficiently small that it shouldn't affect error rates much and if we care about any data there I think it's fine to backfill those from errors.

I used actual data we've collected instead of the list at https://github.com/mozilla-services/mozilla-pipeline-schemas/issues/262 to generate these schemas. The difference between them looks something like this:

    $ comm -3 founddoctypes issueandschemasrepodoctypes
    activation
    android-anr-report
    appusage
    b2g-installer-device
    b2g-installer-flash
    certificate-checker
    flash-video
    idle-daily
    loop
            pre-account
            prio
    shield-study-icq
    tls13-middlebox-testing
    tls-13-study
    tls-13-study-v2

@mreid-moz take a look at things in the first column and let me know if you think any of them should have default schemas. Most are old pings we're not collecting anymore. The things in the second column are new pings that we haven't deployed schemas for or don't have data for yet.

As to the naming, I'm using "default" schema here but the terminology may be confusing and something along the lines of "legacy" may be more appropriate.

There's a competing sync ping implementation in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/268. My understanding is that we should use this version, since there may be some source data inconsistencies that are difficult to deal with in that one (recognizing that this means "promoting" sync pings to split things out of additionalProperties may be fraught with peril). I also notice that there are v4 and v5 sync pings, which is perhaps related.

Lastly, I haven't verified that https://github.com/mozilla/mozilla-schema-generator, when applied to these schemas, produces what we expect as bq schema. We should verify that before merging.

r? @mreid-moz but probably other people should look at this as well.


whd:
Updating r?s since :mreid is out: trink for AWS-side and :frank for GCP-side, with delegation as-needed. We can always add additional schemas for old doctypes as necessary.

mreid-moz:
- activation - this is a glean ping, sent by fenix, and [it has a schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/org-mozilla-fenix/activation/activation.1.schema.json) (maybe needs deploying still).
- android-anr-report - I don't know details on this, but they report "app not responding" and [the code appears to still be active](https://searchfox.org/mozilla-central/rev/aba472751e24763d0c18bae8408e9d7106e9acea/mobile/android/base/java/org/mozilla/gecko/ANRReporter.java#273-294). Let's use the default schema here.
- appusage - This is an ancient leftover from b2g. Rejecting these pings seems safe to me.
- b2g-installer-device - ditto
- b2g-installer-flash - ditto
- certificate-checker - I don't know what these are, can we look at a few examples?
- flash-video - these are [old and ignorable](https://bugzilla.mozilla.org/show_bug.cgi?id=1119291).
- idle-daily - I believe these are a leftover from pre-unified telemetry - is that possible? If that is the case, it's safe to reject these.
- loop - these are also [old and safe to reject](https://bugzilla.mozilla.org/show_bug.cgi?id=1288470).
- pre-account - This [has a schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/telemetry/pre-account/pre-account.4.schema.json), maybe needs a re-deploy? 
- prio - [ditto](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/telemetry/prio/prio.4.schema.json)
- shield-study-icq - This looks like a variant of [shield-icq-v1](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/telemetry/shield-icq-v1/shield-icq-v1.4.schema.json), 
- tls13-middlebox-testing - There have been a number of TLS-related studies, I am not confident in saying we can reject them, so let's use the default schema here.
- tls-13-study - ditto, let's use the default schema here.
- tls-13-study-v2 - ditto, let's use the default schema here.

mreid-moz:
> As to the naming, I'm using "default" schema here but the terminology may be confusing and something along the lines of "legacy" may be more appropriate.

I'd be inclined to call it `placeholder` - for all future pings there should be no concept of "default", and we should use real schemas.
 
> There's a competing sync ping implementation in #268. My understanding is that we should use this version, since there may be some source data inconsistencies that are difficult to deal with in that one (recognizing that this means "promoting" sync pings to split things out of additionalProperties may be fraught with peril). I also notice that there are v4 and v5 sync pings, which is perhaps related.

If we have a documented path to promoting info out of `additionalProperties`, I am good with starting from a dummy schema for `sync` pings. It appears that desktop and mobile are sending slightly inconsistent variants which makes it difficult to finish #268 without some help from the FxA / Sync team. I'd be happy to work on the sync ping schema more incrementally if we can get that process documented.

> Lastly, I haven't verified that https://github.com/mozilla/mozilla-schema-generator, when applied to these schemas, produces what we expect as bq schema. We should verify that before merging.

r? @fbertsch for this piece.

mreid-moz:
> I generated this PR via [this gist](https://gist.github.com/whd/14630dcfb5167115eff5086dfc3f601d). See also for the full list of doctype / docversions. It may be instructive to look at pings for which we are missing a docversion in some cases (crash, event, first_shutdown, health, main, modules, new_profile, shield_study_addon, update) and verify it's ok to throw those data away as errors in future.

It seems reasonable to throw these away if they're missing a version, but let's file a follow-up issue to do this verification.
 
> Shield studies schema handling has always been a strange case and I don't know how we want to handle those, but this PR should in theory do nothing on the AWS pipeline because this default schema is already being applied there.

This sounds good to me.

> I decided not to check `UNKNOWN`/`OTHER` bucket objects for further potential acceptable
> doctypes; the volume is sufficiently small that it shouldn't affect error rates much and if we care about any data there I think it's fine to backfill those from errors.

Seems fine to skip `UNKNOWN`, but let's file a follow-up to review doctypes from `OTHER` - it's quite likely that there are things in there we want to keep.

whd:
>     * activation - this is a glean ping, sent by fenix, and [it has a schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/org-mozilla-fenix/activation/activation.1.schema.json) (maybe needs deploying still).

The activation ping referenced here is actually something totally different, that in the minimal archaeology I did dates back to at least 2015. I think it's safe to ignore as such (we don't have any data to examine from the last year of this type). The glean one is in the structured ingestion namespace so wasn't considered for this round of schemas.

> 
>     * android-anr-report - I don't know details on this, but they report "app not responding" and [the code appears to still be active](https://searchfox.org/mozilla-central/rev/aba472751e24763d0c18bae8408e9d7106e9acea/mobile/android/base/java/org/mozilla/gecko/ANRReporter.java#273-294). Let's use the default schema here.
> 

I will add this.

>     * certificate-checker - I don't know what these are, can we look at a few examples?

We can't, because we don't have recent enough data . For what it's worth, [this commit](https://github.com/mozilla-services/puppet-config/commit/2b43eb4c750f37071ffb3b0c3fb228b572655355) dated March 2017 is when it was added as a known doc type. Unfortunately it doesn't link to e.g. a relevant client bug.

>     * idle-daily - I believe these are a leftover from pre-unified telemetry - is that possible? If that is the case, it's safe to reject these.

Yes, and in fact, I do not see any decoded data of this type for the last year. 

>     * pre-account - This [has a schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/telemetry/pre-account/pre-account.4.schema.json), maybe needs a re-deploy?

This was originally listed as "new pings that we haven't deployed schemas for or don't have data for yet.", which is accurate. Ditto prio.

>     * shield-study-icq - This looks like a variant of [shield-icq-v1](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/telemetry/shield-icq-v1/shield-icq-v1.4.schema.json), it may be safe to reject them, I will follow up on this one. **Update** I talked to @Dexterp37 and he linked me to the [icq shield addon telemetry docs](https://github.com/Dexterp37/icq-shield-study/blob/master/docs/TELEMETRY.md), and confirmed that this data is no longer needed.
> 

Sounds good.

>     * tls13-middlebox-testing - There have been a number of TLS-related studies, I am not confident in saying we can reject them, so let's use the default schema here.
> 
>     * tls-13-study - ditto, let's use the default schema here.
> 
>     * tls-13-study-v2 - ditto, let's use the default schema here.

Ok, I will add these.

whd:
With respect to the schemas I'm adding, I don't have source data to inspect for them and as such I'm making educated guesses as to the acceptable doc versions. For `android-anr-report`, that will be 1-10 to correspond to core pings. For tls studies, it will correspond to the docversions found for the other tls studies, which I believe is just 4. For certificate-checker, (if we decide to add a schema), I will use 4 because it appears to have been added during the era of unified telemetry.

whd:
> I'd be inclined to call it `placeholder` - for all future pings there should be no concept of "default", and we should use real schemas.

`placeholder` implies it will eventually be replaced, which I doubt is the case for the majority of these pings. It is perhaps a thing to aspire to however, so I'll update the PR as such.

> 
> If we have a documented path to promoting info out of `additionalProperties`, I am good with starting from a dummy schema for `sync` pings. It appears that desktop and mobile are sending slightly inconsistent variants which makes it difficult to finish #268 without some help from the FxA / Sync team. I'd be happy to work on the sync ping schema more incrementally if we can get that process documented.

Regarding documentation, @fbertsch wrote a google doc with an outline of the operational side of the schemas update process, which includes information as to what kinds of schema updates are permissible. We don't have client/developer documentation anywhere as far as I know, but it is my understanding that in the current pipeline if the source types are inconsistent there is no way to handle this data without writing a special case, so the upgrade path for this ping would still be TBD. I'm separately writing up a PR for reprocessing pathways we'll support in GCP, of which this might be a special case, and perhaps an annotation for forcing type conversion on specific fields may be something we want to support (and document) explicitly.

Regardless, I think any path forward for sync pings will start with this base schema, and involve updating it to become more like the schema in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/268.

whd:
@mreid-moz I've updated this with the changes that were suggested and filed issues for the followup items. I added certificate-checker even though I don't have any reference pings for it.  Please re-r? and then I will merge this. We can verify the schema generator works as expected as we add more items to the allowlist maintained in that repo.

fbertsch:
The biggest limitations for schema changes in BQ are: no type changes, and no column deletions. I'm assuming the differences between the sync pings are different types, which means they are definitely not compatible in BQ.

These will output the following BQ table:
```
[
  {
    "mode": "REQUIRED",
    "name": "root",
    "type": "STRING"
  }
]
```

We'll need to test these against the pipeline.

whd:
> The biggest limitations for schema changes in BQ are: no type changes, and no column deletions. I'm assuming the differences between the sync pings are different types, which means they are definitely not compatible in BQ.

I have an agenda item for Monday to discuss specifically supporting type coercion of known bad fields, because this is the third such instance of "bad input data". Regardless, starting with a schema-less sync ping I think is the way to go.

> 
> These will output the following BQ table:
> 
> ```
> [
>   {
>     "mode": "REQUIRED",
>     "name": "root",
>     "type": "STRING"
>   }
> ]
> ```
> 
> We'll need to test these against the pipeline.

The piece I'm concerned about is that [additional_properties](https://github.com/mozilla/gcp-ingestion/blob/master/docs/bigquery_sink.md#accumulate-unknown-values-as-additional_properties) is correctly added in the metadata merge.

fbertsch:
Gotcha. I ran them both:

[Telemetry JSON Schema](https://gist.github.com/fbertsch/571b48d3762deae07c2f4d2d29af79db)
[Telemetry BQ Schema](https://gist.github.com/fbertsch/1e9ad135edc76d6a26476b078f2f0b99)

[Structured Ingestion JSON Schema](https://gist.github.com/fbertsch/18ac9e08d5078b5f32ca2558e52044dc)
[Structured Ingestion BQ Schema](https://gist.github.com/fbertsch/d6de5ccf4dded7bf650dc3b2d86f4628)

`additional_properties` looks to get added successfully to every table.

whd:
Cool, that looks correct to me (i.e. additional_properties should be top-level). I think with final approval from @mreid-moz this is ready to merge.

templates/include/common/placeholder.1.schema.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema" : "http://json-schema.org/draft-04/schema#",
+  "type" : "object",
+  "title" : "default_schema",
```
mreid-moz: nit: "placeholder_schema"

PR #336, opened 2019-05-23
Add stub ping specific fields

relud:
cc @matt-howell


robotblake:
So just to clarify (because I'm a little rusty on schema definitions), these are optional fields, yes?

relud:
correct

whd:
@robotblake can you re-r? this? Once this and https://github.com/mozilla/mozilla-schema-generator/pull/24 are merged and deployed I believe we have everything we need to process stub installer pings into BQ.

PR #338, opened 2019-06-05
1556684: Fix label length so it can support any metric name

mdboom:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/glean/labeled_group.1.schema.json
+++ b/templates/include/glean/labeled_group.1.schema.json
@@ -2,5 +2,6 @@
 "propertyNames": {
   "type": "string",
   "pattern": "^[a-z_][a-z0-9_]{0,29}(\\.[a-z0-9_]{0,29})*$",
-  "maxLength": 30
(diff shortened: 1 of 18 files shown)


mdboom:
@fbertsch: It seems like this doesn't need parquet changes because the labeled keys aren't length restricted there, but I'd appreciate your confirmation on that.

templates/include/glean/labeled_group.1.schema.json
```diff
@@ -2,5 +2,6 @@
 "propertyNames": {
   "type": "string",
   "pattern": "^[a-z_][a-z0-9_]{0,29}(\\.[a-z0-9_]{0,29})*$",
-  "maxLength": 30
+  "comment": "This must be at least the length of 'category.name' metric names to support error reporting",
+  "maxLength": 61
```
Dexterp37: Do we want to update the changelog with that? If so, let's also mention the related bug and the rationale.

mdboom: Good point.  I forgot we had a glean changelog over here.

PR #340, opened 2019-06-06
Update payload section for the Prio ping (fx69.0a1)

acmiyaguchi:
This adds a validation ping that was generated from Firefox 69.0a1. This updated schema portion will let me write ETL directly from BigQuery.

I think there's a bug with the version number, since all the pings that I've seen so far have had the version encoded as a string. I updated the schema to reflect this change.


acmiyaguchi:
This is a schema incompatible change with the most recent generated BigQuery schemas in  9437f7c. The table has been dropped in preparation for this change.

PR #342, opened 2019-06-11
Add schemas for Glean history and bookmarks sync pings

linacambridge:
These were added in mozilla-mobile/android-components#3092. I cribbed this PR from mozilla-services/mozilla-pipeline-schemas#325, please let me know if there's anything that needs to happen next for us to see the data in re:dash!

(CC @Dexterp37 @mdboom)

Checklist for reviewer:

For glean changes:


linacambridge:
Hmm, was there a reason underscores were disallowed in schema names? I'm wondering if that's causing the failure...but also, the [debug ping viewer](https://debug-ping-preview.firebaseapp.com/pings/a90b3646-dc2e-437c-84af-782d97e1c917/lina) reports `No such docType: org-mozilla-fenix-debug/bookmarks_sync`, is there something else I should call it if I can't use underscores?

CMakeLists.txt
```diff
@@ -53,9 +53,9 @@ endforeach()
 file(GLOB_RECURSE SCHEMAS LIST_DIRECTORIES FALSE RELATIVE "${CMAKE_SOURCE_DIR}/templates" "${CMAKE_SOURCE_DIR}/templates/*")
 foreach(ITEM IN ITEMS ${SCHEMAS})
   if(NOT ITEM MATCHES "^(include/|.*/\\.[^/])" )
-    string(REGEX MATCH "/([-a-zA-Z0-9]+\\.[1-9][0-9]*\\.schema\\.json|[-a-zA-Z0-9_]+\\.[1-9][0-9]*\\.parquetmr\\.txt|README.*)$" MATCH ${ITEM})
+    string(REGEX MATCH "/([-a-zA-Z0-9_]+\\.[1-9][0-9]*\\.schema\\.json|[-a-zA-Z0-9_]+\\.[1-9][0-9]*\\.parquetmr\\.txt|README.*)$" MATCH ${ITEM})
```
Dexterp37: Mh, this is for the schema filenames, I think. @trink , thoughts?
I think, at the very least, this should allow the same characters in the ping filename that we allow for the ping type [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/2dd16ff73b5c857a4d9b90325e20f32145fbbbaf/schemas/glean/baseline/baseline.1.schema.json#L754-L756).

trink: The current regex matches what is documented here (the docs actually need to be updated to include the numeric range): http://docs-origin.telemetry.mozilla.org/cookbooks/new_ping.html?highlight=namespace#choose-a-namespace-and-doctype.  There is no restriction in the ingestion ETL the URL is simply split on the `/` and the corresponding values are used for the schema lookup.

Dexterp37: I see, thank you @trink ! Then I think it's now up to @fbertsch or @mreid-moz to decide if this is ok to extend or not :)

linacambridge: > There is no restriction in the ingestion ETL the URL is simply split on the / and the corresponding values are used for the schema lookup.

That's what I was wondering about. 😄 The [Glean debug ping viewer](https://debug-ping-preview.firebaseapp.com/pings/UNKNOWN/lina) shows errors like:

> ParsePayload com.mozilla.telemetry.schemas.SchemaNotFoundException: No such docType: org-mozilla-fenix-debug/bookmarks_sync

I guess debug builds [wouldn't send telemetry anyway](https://github.com/mozilla-mobile/fenix/blob/a58a77317e49814c3e1f046b2d4fda2a49abcff5/app/build.gradle#L207-L211), but, for release builds—does the file also need to be named `org-mozilla-fenix/bookmarks_sync`, or can it have a different name?

Dexterp37: It needs to have the same name as the ping name in the `pings.yaml` file, and it must be in a directory with the application id as the directory name. Your current PR seems to look good.

PR #344, opened 2019-06-11
Add missing string types for some enum fields.

mreid-moz:
Similar to #343 but for glean. Should be a no-op from a validation / structure standpoint. See [this gist](https://gist.github.com/acmiyaguchi/bbfb78590f0c5030c87ade879405d73e) for context.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/glean/baseline/baseline.1.schema.json
+++ b/schemas/glean/baseline/baseline.1.schema.json
@@ -324,7 +324,8 @@
                     "minute",
                     "hour",
                     "day"
-                  ]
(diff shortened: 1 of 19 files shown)


mreid-moz:
> This is basically just for schema generator so it doesn't have to infer from the `enum` values that the type is a `string`? If so, LGTM.

Yep, exactly.

In terms of validation, we have test docs that include fields like "time_unit", so I'm confident we won't introduce new schema validation errors here.

PR #362, opened 2019-07-24
Rename `string_list` to `string_set`

Dexterp37:
This additionally changes the schema to check for unique values within a `string_set`.

Note: nobody is using `string_list` yet so this is a safe change.

Checklist for reviewer:

For glean changes:


acmiyaguchi:
Just noting that renaming/deleting a field is typically an incompatible with [schema evolution rules for BigQuery](https://cloud.google.com/bigquery/docs/managing-table-schemas). Generally this would require dropping the existing table or bumping the schema version, but it looks like the generated schema is missing these fields in the metrics section.

Dexterp37:
> Edit: This is currently an unused fields without any instances, so normal rules don't apply since they're dropped before they make it into the table schema. 👍

Yes ;) That's why we're pushing hard to make this change *now*, before anyone uses it. We have the first consumer lined up for 2 weeks from now ;)

Dexterp37:
Good point Frank. I think that, as of today, this too invasive and too late. Let's defer this for future discussion and consideration.

PR #365, opened 2019-07-24
Fix #363 - Update `payload.info` in main ping schema

### LINKED ISSUE #363: Add `payload.info` to the main ping schema
The main ping schema is currently missing most of the fields inside of `payload.info`. The Firefox source docs has [some information on this](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/main-ping.html#info).

acmiyaguchi:
This fixes #363 using the environment information on [this page](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/main-ping.html). I've also checked the additionalProperties to double check the types of each of the fields. The table in [this issue](https://github.com/mozilla/gcp-ingestion/issues/711#issue-472594034) shows the types and frequency of each field.

I left out `addons` and `flashVersion` because the docs say the fields are obsolete. Based on the data, it seems that they're being reported by a significant number of clients. It might be useful to add here just to increase the coverage of the schema and to reduce the size of additional properties.

Checklist for reviewer:

### DIFF
--- a/schemas/telemetry/main/main.4.schema.json
+++ b/schemas/telemetry/main/main.4.schema.json
@@ -1115,8 +1115,71 @@
         },
         "info": {
           "properties": {
-            "subsessionLength": {
+            "previousBuildId": {
+              "description": "null if this is the first run, or the previous build ID is unknown",
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "previousSessionId": {
+              "description": "session id of the previous session, null on first run.",
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "previousSubsessionId": {
+              "description": "subsession id of the previous subsession (even if it was in a different session), null on first run.",
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "profileSubsessionCounter": {
+              "description": "the running no. of all subsessions for the whole profile life time",
               "minimum": 0,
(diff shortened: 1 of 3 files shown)


mreid-moz:
The reports from the integration test are empty again - can you do a manual run of the error report?

Otherwise, this looks good!

acmiyaguchi:
@mreid-moz integration tests require #361 to work

mreid-moz:
Do we need special handling here to address splitting these fields out of `additional_properties` in BQ?

acmiyaguchi:
@mreid-moz: Yes, queries at that cross today's date boundary are going to need to read from both fields. It makes @relud's [main summary reimplementation](https://github.com/mozilla/bigquery-etl/pull/258) a bit tricky. It might be possible to backfill the fields using DML, but it's not a small change.

relud:
as per [whd's comment on snake casing addtional_properties](https://github.com/mozilla/gcp-ingestion/pull/713#issuecomment-515595703), the table should get backfilled from protobuf after it moves to prod.

We should merge this before then, so the [main summary reimplementation](https://github.com/mozilla/bigquery-etl/pull/258) can simply read from the new column.

if it does impact that query, then i'll just modify it to use `COALESCE` between `additionalProperties` and the new column.

PR #366, opened 2019-07-30
Remove labeled timespan

badboy:
We are currently removing support for it.
I did the research and it seems to never have been in use anywhere:
* I looked through the code bases of Fenix, a-c & reference-browser. Except for Glean itself in tests, no `labeled_timespan` was used
* I looked through the data available in parquet tables for Fenix, Fenix Nightly, Reference browser, the sample app and Firefox for FireTV. None contained any labeled timespan data.

cc @Dexterp37, @mdboom (can't add them as reviewers)

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/glean/baseline/baseline.1.parquetmr.txt
+++ b/schemas/glean/baseline/baseline.1.parquetmr.txt
@@ -172,20 +172,6 @@ message baseline {
         }
       }
     }
-    optional group labeled_timespan (MAP) {
-      repeated group key_value {
-        required binary key (UTF8);
-        required group value (MAP) {
-          repeated group key_value {
-            required binary key (UTF8);
-            required group value {
-              required int64 value;
-              required binary time_unit (UTF8);
-            }
-          }
-        }
-      }
-    }
     optional group timing_distribution (MAP) {
       repeated group key_value {
         required binary key (UTF8);
--- a/schemas/glean/baseline/baseline.1.schema.json
+++ b/schemas/glean/baseline/baseline.1.schema.json
@@ -311,47 +311,6 @@
           },
           "type": "object"
         },
-        "labeled_timespan": {
-          "additionalProperties": {
-            "additionalProperties": {
-              "properties": {
-                "time_unit": {
-                  "enum": [
-                    "nanosecond",
-                    "microsecond",
-                    "millisecond",
-                    "second",
-                    "minute",
-                    "hour",
-                    "day"
-                  ],
-                  "type": "string"
-                },
-                "value": {
-                  "type": "integer"
-                }
-              },
-              "required": [
-                "value",
-                "time_unit"
-              ],
-              "type": "object"
-            },
-            "propertyNames": {
-              "comment": "This must be at least the length of 'category.name' metric names to support error reporting",
-              "maxLength": 61,
-              "pattern": "^[a-z_][a-z0-9_-]{0,29}(\\.[a-z0-9_-]{0,29})*$",
-              "type": "string"
-            },
-            "type": "object"
-          },
-          "propertyNames": {
-            "maxLength": 61,
-            "pattern": "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})+$",
-            "type": "string"
-          },
-          "type": "object"
-        },
         "labeled_timing_distribution": {
           "additionalProperties": {
             "additionalProperties": {
--- a/schemas/glean/events/events.1.parquetmr.txt
+++ b/schemas/glean/events/events.1.parquetmr.txt
@@ -172,20 +172,6 @@ message baseline {
         }
       }
     }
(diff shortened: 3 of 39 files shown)


mdboom:
`labeled_timespan` does occur in a few of the bq schemas for Fenix, for example [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/generated-schemas/schemas/org-mozilla-fenix/bookmarks_sync/bookmarks_sync.1.bq#L546).

@fbertsch: Does that mean this is a no-go?  It's not quite clear to me where that came from, since it's not in any of the metrics.yaml (as @badboy mentioned above).

fbertsch:
Checking in on this now. It looks to me like those pieces should _not_ be in those schemas, and I'm trying to figure out why they are.

fbertsch:
Okay, bad news. The schema generator doesn't know anything about those sync pings, because sync is not a product. Because it doesn't know about sync pings, it doesn't fill in those schemas; so those have all the fields and no top-level metrics. No good.

We'll need to update the schema generator to read in each dependency's `metrics.yaml` to see if there's any pings it missed, and fill those in as well.

mdboom:
> We'll need to update the schema generator to read in each dependency's metrics.yaml to see if there's any pings it missed, and fill those in as well.

I guess I didn't track what schema_generator was doing with the output of probe_scraper.  Is it doing anything with the dependencies now?

Also, would it make more sense for probe_scraper to do this merging of all the metrics for each app and its dependencies?

Dexterp37:
> Also, would it make more sense for probe_scraper to do this merging of all the metrics for each app and its dependencies?

I think I agree with @mdboom here: other consumers shouldn't really know too much about `metrics.yaml` files :(

fbertsch:
@mdboom What it's doing right now is reading in those other `metrics.yaml` files, and possibly filling in additional metrics to the pings that the _application_ sends (so: baseline, metric, event, and any custom pings the app sends). It does _not_ check to see if those dependencies will be sending any of their own pings.

fbertsch:
We agreed to not do that in the probe_scraper to keep things simpler, and that downstream from the probe_scraper could handle merging dependencies + app metrics.

Dexterp37:
@fbertsch , just to clarify (and sorry if you already replied somewhere else to this same question): is the outcome of your investigation that we **cannot** remove `labeled_timespans` from the schema?

fbertsch:
@Dexterp37 we should be able to once we deploy the new schema for the sync pings. The problem was special for that case:

0. New pings were added for a new library
1. Those same pings schemas were added to mozilla-pipeline-schemas
2. That library was _not_ added to the probe-scraper

So what happened was that the version of the schema in mozilla-pipeline-schemas was deployed, instead of the one we create on our end.

This won't be an issue in the future because no one should be submitting glean-pings to mozilla-pipeline-schemas; instead they will all be generated.

Dexterp37:
> This won't be an issue in the future because no one should be submitting glean-pings to mozilla-pipeline-schemas; instead they will all be generated.

Oh, I see! Thanks for explaining this, the problem is much clearer now!

badboy:
@fbertsch Did a new sync ping schema land? What's the way forward on this PR? 

We potentially have some (now-parser-disabled) other metric types that are unused and I wonder if we should also remove them before while we're at it.

fbertsch:
It did, the [schemas](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/generated-schemas/schemas/org-mozilla-fenix/bookmarks_sync/bookmarks_sync.1.bq) have the correct fields, and it has been deployed. Please do one more check on those Glean schemas to ensure there are no `metrics/labeled_timespan` fields.

badboy:
I can still find it in 3 bq schemas from the sample app (`org-mozilla-samples-glean/`):

1. https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/fbfee46a2d38478f3542ae22f12402d2627f8ec9/schemas/org-mozilla-samples-glean/baseline/baseline.1.bq#L587
2. https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/fbfee46a2d38478f3542ae22f12402d2627f8ec9/schemas/org-mozilla-samples-glean/events/events.1.bq#L587
3. https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/generated-schemas/schemas/org-mozilla-samples-glean/metrics/metrics.1.bq#L587

It's not present in any other schema.

fbertsch:
We can probably safely ignore those, that app was for testing (it is probably not even in use anymore). @Dexterp37 can you confirm that you're not using `org-mozilla-samples-glean`?

Dexterp37:
> 
> 
> We can probably safely ignore those, that app was for testing (it is probably not even in use anymore). @Dexterp37 can you confirm that you're not using `org-mozilla-samples-glean`?

Yes, not using it, we can burn it with fire :)

PR #371, opened 2019-08-05
Add new scalar data to new-profile ping schema

badboy:
This explicitely leaves out changes to the parquet schema, as we still
don't have a good way to coerce potentially different data types (numbers, strings, booleans) into a single type.
(Even though for now it's definitely only a string in the new-profile ping)

m-c Patch: https://bugzilla.mozilla.org/show_bug.cgi?id=1570652
This should land first so that we're sure we don't break any data in the pipeline (though everything in the payload is optional already)

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/new-profile/new-profile.4.schema.json
+++ b/schemas/telemetry/new-profile/new-profile.4.schema.json
@@ -1019,6 +1019,303 @@
     },
     "payload": {
       "properties": {
+        "processes": {
+          "properties": {
+            "parent": {
+              "properties": {
+                "events": {
+                  "items": {
+                    "items": [
+                      {
+                        "minimum": 0,
+                        "type": "integer"
+                      },
+                      {
+                        "type": "string"
+                      },
+                      {
+                        "type": "string"
+                      },
+                      {
+                        "type": "string"
+                      },
+                      {
+                        "type": [
+                          "string",
+                          "null"
+                        ]
+                      },
+                      {
+                        "additionalProperties": {
+                          "type": [
+                            "string",
+                            "null"
+                          ]
+                        },
+                        "type": [
+                          "object",
+                          "null"
+                        ]
+                      }
+                    ],
+                    "maxItems": 6,
+                    "minItems": 4,
+                    "type": "array"
+                  },
+                  "type": "array"
+                },
+                "gc": {
+                  "additionalProperties": false,
+                  "properties": {
+                    "random": {
+                      "items": {
+                        "additionalProperties": {
+                          "type": [
+                            "number",
+                            "string",
+                            "null",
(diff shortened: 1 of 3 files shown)


badboy:
cc @mreid-moz

mreid-moz:
In `main` pings, scalar data is at `payload/processes/<processtype>/scalars` - is it expected that it's in a different location in the `new-profile` ping?

fbertsch:
@badboy is there anything `scalars.yaml` that indicates whether a scalar belongs in this ping? If we have anything we can get this added to the schema generator and it will fill-in the appropriate types.

badboy:
> In `main` pings, scalar data is at `payload/processes/<processtype>/scalars` - is it expected that it's in a different location in the `new-profile` ping?

Hm, we indeed switch these around for the main ping. I could do the same.

> @badboy is there anything `scalars.yaml` that indicates whether a scalar belongs in this ping? If we have anything we can get this added to the schema generator and it will fill-in the appropriate types.

[Yes, this:](https://phabricator.services.mozilla.com/D40596#C1403281NL4297):

```yaml
    record_into_store:
      - main
      - new-profile
```

fbertsch:
Oh that's great. let's get it added once we merge this.

badboy:
@mreid-moz 	updated the PR (the m-c patch was also updated)

badboy:
@mreid-moz 	Can we get this in?

mreid-moz:
@acmiyaguchi The integration test reports are empty in this PR - could you take a look?

acmiyaguchi:
@mreid-moz It ended up being an intermittent failure, I reran the integration tests and it looks fine.

The one thing to consider with this PR is that merging in the change will update the BigQuery table schema to include the parent process data structure on the same day around 5pm PST. If we do end adding functionality to the schema-generator (@fbertsch), then the changes should be merged around the same time. If not, we could run into a difficult situation that requires dropping the table. 

To be more concrete, filling in the schema from the schema generator side will allow querying the data like this:

```sql
SELECT payload.processes.parent.scalars.startup_profile_selection_reason
FROM telemetry_live.new_profile_v4
```

Each metric gets its own column, instead of being part of a large nested row (at least for histograms). Since scalars can have multiple types, they will get moved into the "additional_properties" section of the table:

```sql
-- extract json since `type: [integer, string, boolean]` does not have a SQL analog
SELECT JSON_EXTRACT(additional_properties, "$.payload.processes.parent.scalars.startup.profile_selection_reason") as startup_profile_selection_reason
FROM telemetry_live.new_profile_v4
```

Filling in the schema using the `scalars.yaml` solves the problem of being able to handle variable types, and have it in a natural place in the table.

PR #372, opened 2019-08-05
Add schemas for non-lossy BQ sink

relud:
Checklist for reviewer:

### DIFF
--- a/schemas/metadata/error/error.1.schema.json
+++ b/schemas/metadata/error/error.1.schema.json
@@ -1,0 +1,108 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://jsonschema.net",
+  "properties": {
+    "args": {
+      "type": "string"
+    },
+    "content_length": {
+      "type": "string"
+    },
+    "date": {
+      "type": "string"
+    },
+    "dnt": {
+      "type": "string"
+    },
+    "error_message": {
+      "type": "string"
+    },
+    "error_type": {
+      "type": "string"
+    },
+    "exception_class": {
+      "type": "string"
+    },
+    "geo_city": {
+      "type": "string"
+    },
+    "geo_country": {
+      "type": "string"
+    },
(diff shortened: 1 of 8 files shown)


schemas/metadata/error/error.1.schema.json
```diff
@@ -0,0 +1,108 @@
...
+    "geo_city": {
+      "type": "string"
+    },
+    "geo_country": {
+      "type": "string"
+    },
+    "geo_subdivision1": {
+      "type": "string"
+    },
+    "geo_subdivision2": {
```
relud: @whd the deployed schemas are currently missing the `geo_` and `user_data_` prefixed fields

schemas/metadata/error/error.1.schema.json
```diff
@@ -0,0 +1,108 @@
...
+      "type": "string"
+    },
+    "host": {
+      "type": "string"
+    },
+    "method": {
+      "type": "string"
+    },
+    "payload": {
+      "format": "bytes",
```
relud: @acmiyaguchi i'm unclear on whether the transpiler already does the right thing, or we need to add supports for the bytes format somehow

schemas/metadata/error/error.1.schema.json
whd: ACK

whd:
One thing to note here is that if we're planning on using the standard transpiler to convert these into BQ schemas (and then wind up in the generated-schemas branch), we're going to also need to update MSG to that effect. Currently we [simply remove them](https://github.com/mozilla/mozilla-schema-generator/blob/master/bin/schema_generator.sh#L92).

The BQ table management (WIP [here](https://github.com/mozilla-services/cloudops-infra/tree/data_bq_storage_api/projects/data-shared/tf/modules/bigquery)) for these tables will have special case logic since they are generated by "pipeline family", which isn't a concept propagated to this repository.

relud:
filed #373 to handle that

PR #375, opened 2019-08-06
1571744: Add custom_distribution metric type

mdboom:
This adds a custom_distribution metric type.

Checklist for reviewer:

For glean changes:


templates/include/glean/custom_distribution.1.schema.json
```diff
@@ -0,0 +1,46 @@
...
+      "type": "integer"
+    },
+    "overflow": {
+      "type": "integer"
+    },
+    "sum": {
+      "type": "integer"
+    }
+  },
+  "required": [
```
fbertsch: Shouldn't `histogram_type` be required, so we can always calculate what the buckets are?

templates/include/glean/custom_distribution.1.schema.json
```diff
@@ -0,0 +1,46 @@
+{
+  "type": "object",
+  "properties": {
+    "range": {
```
fbertsch: Is this implementation already in flight? We could do `min_bucket` and `max_bucket` properties instead of an array here.

templates/include/glean/custom_distribution.1.schema.json
```diff
@@ -0,0 +1,46 @@
...
+      "minItems": 2,
+      "maxItems": 2,
+      "items": {
+        "type": "number"
+      }
+    },
+    "bucket_count": {
+      "type": "integer"
+    },
+    "histogram_type": {
```
fbertsch: Interesting, `histogram_type` is an int on desktop. Is this translation to a string happening in GV or Glean?

templates/include/glean/custom_distribution.1.schema.json
mdboom: Good catch.

templates/include/glean/custom_distribution.1.schema.json
mdboom: Implementation is totally fungible.

However, this is based directly on the payload format in legacy desktop telemetry (which already made an appearance in Glean in the "old" timing distribution).  If there are advantages to retaining the same payload format, I'd say we leave this as-is, otherwise, I'm happy to simplify.

templates/include/glean/custom_distribution.1.schema.json
mdboom: Ok, so here, we've gone against desktop. :(

mdboom:
Nothing's set in stone.  I guess we just need a call on whether the guiding principle here is to be the same as desktop or to make things as nice as we can.  @Dexterp37 ?

templates/include/glean/custom_distribution.1.schema.json
fbertsch: Well one issue is that while this is _technically_ an array, it's really a tuple of length two, and we don't have a great way of dealing with tuples in BQ. Luckily because the elements are all the same type this ends up working out.

fbertsch:
@mdboom here's a thoughts: We remove `range`, `histogram_type`, and `bucket_count` and instead send all of the buckets in `values`. That would also make querying cleaner since no one would need to calculate the buckets in SQL.

chutten:
> @mdboom here's a thoughts: We remove `range`, `histogram_type`, and `bucket_count` and instead send all of the buckets in `values`. That would also make querying cleaner since no one would need to calculate the buckets in SQL.

I think it's important to preserve the ability to use the new functional binning if we choose to. At present there are probes that want the new binning but use one-off complex units (or are unitless) so don't fit into even potential future datatypes like `memory_distribution` let alone existing ones like `timing_distribution`

mdboom:
@chutten: I'm not sure if whether we include this metadata here impacts the functional binning options, since those will be on different metric types.  But quite possible I'm missing something...

chutten:
I'm thinking that `custom_distribution`s might want to opt to use the functional binning... in which case we won't (can't) send all the buckets like Frank proposes. I'm raising the point here to make sure we can write the schema in such a way that we don't prevent the fancy binning from being used (or, if we do prevent it, we do so on purpose).

mdboom:
@chutten: I guess I was assuming if a metric wanted fancy binning, they wouldn't want a "custom" distribution metric type.  Were you thinking that fancy binning might just be another enumeration value on the `histogram_type` parameter?

Dexterp37:
> 
> 
> @chutten: I guess I was assuming if a metric wanted fancy binning, they wouldn't want a "custom" distribution metric type. Were you thinking that fancy binning might just be another enumeration value on the `histogram_type` parameter?

We might want to use the fancy binning with no time unit, for example.

mdboom:
> We might want to use the fancy binning with no time unit, for example.

Ok -- I could see that going two ways -- a new metric type called "???Distribution" -- or another enumeration value to `histogram_type`.  Not sure I have a strong argument either way there.

The larger question is: do we need to communicate the parameters of the distribution to the backend through the ping, or do we just (always) send all of the bucket edges so the backend doesn't have to know anything about bucket calculation.  I think I would prefer the latter -- it puts the responsibility of calculating buckets all in one place.  I would also argue we don't need to send *all* buckets, just the contiguous set of buckets between the minimum and maximum buckets (which is what the fancy binning PR already does).

fbertsch:
My vote is the following:
- New metric types for these fancy distributions and special Glean pieces. I believe we should leave `CustomDistribution` to exactly that, and we can have nicer, more advanced Glean distributions down the line
- Send all the buckets in the ping. If a `CustomDistribution` ping is using a function, we can send all the buckets between [0, max]; this would guarantee that any aggregation wouldn't be "missing" any buckets that are less than max.

Dexterp37:
> 
> 
> > We might want to use the fancy binning with no time unit, for example.
> 
> Ok -- I could see that going two ways -- a new metric type called "???Distribution" -- or another enumeration value to `histogram_type`. Not sure I have a strong argument either way there.

Ok, I see your point. Let's not cram everything into *this* `CustomDistribution`. This metric is about a distribution that has parameters that can be tweaked. We'll figure out another name for a "fancy distribution metric with no unit" in a separate pr.
 
> The larger question is: do we need to communicate the parameters of the distribution to the backend through the ping, or do we just (always) send all of the bucket edges so the backend doesn't have to know anything about bucket calculation. I think I would prefer the latter -- it puts the responsibility of calculating buckets all in one place. I would also argue we don't need to send _all_ buckets, just the contiguous set of buckets between the minimum and maximum buckets (which is what the fancy binning PR already does).

I think that, for this specific `CustomDistribution`, since it's targetting legacy Gecko metrics, we should have a non-surprising behaviour for analysts and data engineers. I'm fine with sending buckets between [0, max], if this won't impact storage costs and ping size. 

Let's take the "fancy binning custom thing" to a separate bug/pr.

mdboom:
Ok.  So I think consensus is:

1) custom_distribution is only for handling the legacy linear or exponential binning with min/max/bucket_count as parameters.  Functional binning will be implemented in other metric types.
2) We don't need to communicate the bucketing parameters (range/bucket_count/histogram_type) in the ping payload.
3) Instead, we will send all buckets from [0, max) in every ping.  I've already confirmed (in the course of other investigations), that the aggregation should handle this just fine.

I think this matches with what @fbertsch and @Dexterp37 said above (but please confirm) and @chutten, please weigh in as well?

chutten:
That interpretation agrees with mine.

fbertsch:
Sounds like we are all in agreement!

Dexterp37:
> 
> 
> Ok. So I think consensus is:
> 
>     1. custom_distribution is only for handling the legacy linear or exponential binning with min/max/bucket_count as parameters.  Functional binning will be implemented in other metric types.
> 
>     2. We don't need to communicate the bucketing parameters (range/bucket_count/histogram_type) in the ping payload.
> 
>     3. Instead, we will send all buckets from [0, max) in every ping.  I've already confirmed (in the course of other investigations), that the aggregation should handle this just fine.
> 
> 
> I think this matches with what @fbertsch and @Dexterp37 said above (but please confirm) and @chutten, please weigh in as well?

Sounds good to me!

Dexterp37:
Whoops, I pressed the wrong button :P

PR #378, opened 2019-08-08
Add document_* attributes to metadata in decoded bytes schema

### LINKED ISSUE #377: Add back missing attributes to payload_bytes_decoded schema
/CC @jklukas 

[Here's](https://github.com/mozilla-services/cloudops-infra/commit/a0d1bb4d94e81b72b6ff02031a95bf0391a0bf26#diff-1fec0cabf755a2a7b5c0b38b3963eec2) the version that is currently deployed. [Here's](https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/59fedd57ea36c3ed9e09dbb32dc9a68761352f7c#diff-90730b5111929a4fca352a5b0f5584b4) the version newly-created on the generated-schemas branch. It appears we've dropped `metadata.document_*` somewhere along the line:

 	* google_bigquery_table.telemetry_tls_13_study_v3_v4_telemetry_decoded: googleapi: Error 400: Provided Schema does not match Table moz-fx-data-shar-nonprod-efed:payload_bytes_decoded.telemetry_telemetry__tls_13_study_v3_v4. Field metadata.document_namespace is missing in new schema, invalid

I'm not sure if we wanted these fields at the top level (as in the errors schema chase) or within `metadata` so I'm filing this as an issue instead of a PR. I'm thinking these are probably meant to be added to `ingestionCommonMetadata.1.schema.json` but I'm not 100%.

jklukas:
Closes #377


templates/metadata/decoded/decoded.1.schema.json
```diff
@@ -3,7 +3,23 @@
   "id": "http://jsonschema.net",
   "type": "object",
   "properties": {
-    @METADATA_TELEMETRYINGESTION_1_JSON@,
+    @METADATA_INGESTIONTOPLEVEL_1_JSON@,
+    "metadata": {
```
jklukas: I refactor a bit in this PR such that the `metadata` struct always appears explicitly. This allows inserting the additional document_* attributes here, but also makes for the definitions of telemetry-ingestion and structured-ingestion to be more parallel. We factor out `uri` to its own include file instead, so that it can be used in the decoded schema and the telemetry-ingestion schema.

PR #380, opened 2019-08-12
Remove error fields from raw schema

relud:
Checklist for reviewer:


relud:
cc @whd

PR #387, opened 2019-08-15
Enforce snake_case naming for namespaces and docTypes

jklukas:
This is a draft of a large change in this repo to make file names and paths match exactly with how datasets and tables are named in BigQuery. The benefit here is both for humans (it's more obvious what the BQ table name is going to be) and for machines (we currently have to support special case naming conversions in several different repositories and keep those in sync).

Importantly, this PR also adds in a validation check such that schema generation fails if namespace or doctype doesn't match the snake_case convention, so that there is not a possibility of non-compliant names creeping in in the future.

Merging these changes as-is would likely break many things. At a minimum, we need some mechanism for "aliasing" non-compliant names to the snake_cased variant. [mozilla-schema-generator is gaining that feature](https://github.com/mozilla/mozilla-schema-generator/pull/58) which could be a way to solve this problem for the GCP pipeline. I'm not sure what happens with the AWS pipeline and if this would cause things to break there.

So, this PR is meant to be a starting point for a conversation. Does it seem reasonable to normalize this repository itself and push responsibility for backwards-compatibility to mozilla-schema-generator and potentially some code changes in end consumers? Are there other approaches we should consider?


jklukas:
cc @whd @fbertsch @relud @trink @mreid-moz

whd:
Whoo boy, I've been puzzling on this for almost an hour and don't have a full response written. This may be worth discussing on Monday, or else I will try to get my thoughts written down, but that they're not clearly formed yet is indicative that there is a lot of nuance here.

I will say as to "other approaches to consider" we may be able to get away with leveraging the schemas aliases functionality we've mostly deprecated e.g. something like:

    {
      "aliases": [
        {
          "alias": "org-mozilla-fenix/bookmarks-sync",
          "base": "org-mozilla-fenix/bookmarks_sync"
        },
        {
          "alias": "org-mozilla-fenix/history-sync",
          "base": "org-mozilla-fenix/history_sync"
        }
      ]
    }

This issue we're primarily trying to deal with is https://github.com/mozilla/gcp-ingestion/issues/744 and how we look up schemas on output via our BQ schema store (input schemas via JSON schema store are fine and dandy), specifically that the BQ store assumes that `_`s it sees should be converted back to their original `-`s. This is the case for all namespaces but not all doctypes due to this newly uncovered special case. Perhaps leveraging beam-specific aliasing functionality to patch lookups would be a sufficient resolution.

jklukas:
> Perhaps leveraging beam-specific aliasing functionality to patch lookups would be a sufficient resolution

I agree that's the easiest thing to do for the short term and I'm happy to do something there to get doctypes flowing for the prod pipeline, but I still think we need to enforce a stricter naming standard in this repo going forward.

A less disruptive alternative would be to standardize on "kebab-case" which is the _de facto_ standard in this repo currently. That at least makes the behavior more predictable for tooling, though it's still more complicated than having the schema paths match the BQ table names exactly.

jklukas:
I am also willing to consider that this is a cause that would be better taken up after the AWS pipeline is torn down so that there's less complexity involved in adapting to changes.

whd:
> I am also willing to consider that this is a cause that would be better taken up after the AWS pipeline is torn down so that there's less complexity involved in adapting to changes.

I would prefer this, and I agree that we need to enforce a stricter naming standard in this repo going forward. I'm writing my thoughts on this down for consideration at a later time post AWS-teardown.

I think it's useful to maintain a distinction between (1) input data standards, (2) output table standards, (3) what is codified in this repo.

1. is kebab-case for all namespaces and doctypes, but there are deviations e.g. `untrustedModules` `bookmarks_sync` 
2. is snake_case, for all namespaces and doctypes
3. is currently
   a. for json schemas, `schemas/namespace/doctype/doctype.docversion.schema.json`, using (1) for the casing of all its constituents (i.e. no normalization occurs).
   b. for bq schemas, `schemas/namespace/doctype/doctype.docversion.bq`, where doctype is kebab-cased to match the input convention of (1).

(3a) is currently entirely consistent with (1), which is in my opinion desirable. To take the case of `untrustedModules`, the resultant schemas live in the `generated-schemas` branch as:

`schemas/telemetry/untrustedModules/untrustedModules.4.schema.json`
`schemas/telemetry/untrusted-modules/untrusted-modules.4.bq`

(3b) is primarily what I would like to consider. We have the following the end users of these schemas: (A) the table management infrastructure, (B) the beam job that decodes data, (C) the beam job that writes data to live tables, and (D) humans reading the repository. Of these (3b) is used by (A), (C), and (D).

A. Assumes schemas from (3b) are in the format of (1), and normalizes (snake_cases) kebab-case namespaces and doctypes in the table names it generates, enforcing (2). Notably there is no special casing logic here because the special casing logic was pushed to MSG.
B. Currently is able to read the schema data from (3a) without applying any normalization.
C. Currently uses snake_cased attributes in its logic for parsing destination table, for both namespace and doctype. It assumes there are no special cases in (1) and thus performs a kebab-case of the snake-cased doctype, which is where our error occurs.
D. As a human, you can't presently infer from the BQ table path for `untrusted_modules` that there is a special input case of `untrustedModules`. You could see the `untrustedModules` json schema is present and infer from that however.

I would propose that the most consistent way to handle things, while maintaining information on what the input data is and any special cases we have, would be to modify (3) to match the following example:

`schemas/telemetry/untrustedModules/untrustedModules.4.schema.json`
`schemas/telemetry/untrustedModules/untrusted_modules.4.bq`

Which is to say:

1. All namespaces in directories match 1:1 the input data i.e. kebab-case (we have thankfully no special cases here)
2. All document types in directories match 1:1 the input data (no normalization)
3. All doctypes in JSON schemas filenames match precisely the input data (i.e. no normalization)
4. All doctypes in BQ schemas filenames match precisely the output table name convention (2 i.e. snake_case)

With the above structure, (A) and (B) need no changes. (C) would need to use a combination of the non-normalized namespace and doctype and normalized (snake_cased) doctype to look up the the schema and table, and normalize (snake_case) all attributes to resolve its destination table. (D) would be able to determine the destination table for a schema by snake_casing namespace and portion of the BQ filename path, while also having information on what is a special case and what is not by whether the snake_casing transformation was a noop or the conversion resulted in a change besides s/-/_/. (D) is able to relatively easily determine whether the input data has a special case, and also where the output data is written to. For consistency there is some merit in normalizing for BQ output the namespace as well, such that the schemas for a table would live at:

`schemas/eng-workflow/specialCase/specialCase.4.schema.json`
`schemas/eng_workflow/special_case/special_case.4.bq`

but you effectively trade (D)'s need to snake_case namespaces with the increased complexity in determining from a BQ schema that there is a special case in the input data (not to mention a proliferation of directories).

Overall this is not a change I want to make before next week's production deploy. It would require relatively small changes to MSG and the BQ sink however, and not affect the AWS pipeline since it only involves changes to BQ schemas.

jklukas:
I think I generally agree with what you laid out, @whd. 

It looks like underscores were only allowed recently due to this change from June: https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/be37b4905323a622df752a749d6977b8d139281d

I think it would be worth considering whether we can change those two docTypes to kebab-case in this repo. I believe the schema generator and the GCP pipeline would both handle it fine.

As for the two existing camelCase docTypes, I'd like to look at moving those to fully compliant kebab-case directories and files, and then add symlinks for the historical cases. We could add these two historical names as special cases in the regex for testing but otherwise disallow any new names with uppercase letters. I believe the AWS pipeline should be compliant with change (we'd still generate the same files in `schemas/` but we'd end up with two additional schemas for the kebab-case variants). Then we can have the schema generator ignore docTypes with uppercase letters and remove the special casing for untrustedModules.

Basically, the existing special cases are controlled enough right now that we can probably patch over them. Again, this probably shouldn't be done in the next few days, and perhaps should wait for the AWS pipeline to be torn down.

jklukas:
The discussion here has been very useful. Closing this PR as we have generally agreed not to pursue this particular course.

I split out the discussion of how to restrict Glean doctype naming to https://github.com/mozilla-services/mozilla-pipeline-schemas/issues/390.

Solution for restricting camelCase proposed in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/391

PR #391, opened 2019-08-19
Disallow camelCase names for document types

jklukas:
As part of this change, we provide symlinks for compliant kebab-case names
for the two existing docTypes which are camelCase. Both variants produce
output in the schemas/ directory and mozilla-schema-generator will be updated
to ignore the old names when creating BQ schemas.

My expectation is that this change to the schema deployment pipeline will
be effectively invisible to the AWS and GCP pipelines. They should both
continue functioning correctly without changes.

See https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/387 for an extended discussion on naming standards in this repository.


jklukas:
See https://github.com/mozilla/mozilla-schema-generator/pull/64 for the related change in MSG.

PR #395, opened 2019-08-20
Add new profile scalars

mreid-moz:
This is effectively the same as #371 but targets the "master" branch (that one accidentally targetted `dev`). Note that there's a terrible hack here to exclude the `processes` chunk from the BigQuery schema for now.

Checklist for reviewer:

### DIFF
--- a/templates/telemetry/new-profile/new-profile.4.schema.json
+++ b/templates/telemetry/new-profile/new-profile.4.schema.json
@@ -17,6 +17,12 @@
             "shutdown",
             "startup"
             ]
+        },
+        "processes": {
+          "type": "object",
+          "properties": {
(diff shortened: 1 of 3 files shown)


mreid-moz:
cc @badboy

templates/telemetry/new-profile/new-profile.4.schema.json
```diff
@@ -17,6 +17,15 @@
             "shutdown",
             "startup"
             ]
+        },
+        "allOf": {
```
acmiyaguchi: `allOf` should actually be a  list of JSON objects. I would place the `$comment` in the `processes` object.

acmiyaguchi: https://json-schema.org/understanding-json-schema/reference/combining.html#allof

PR #405, opened 2019-09-11
Update untrusted modules ping to v2 format

dblohm7:
NOTE:
Since this ping is currently only generated on Nightly, and its results are currently of limited utility, I made no effort to support backward compatibility with the v1 ping format.

Checklist for reviewer:

For glean changes:


acmiyaguchi:
Hello @dblohm7. First, could you rebase you changes on top of the `master` branch? The `dev` branch is no longer being updated.

Secondly, these changes are definitely not backwards compatible with the current schema, which is currently populating the `telemetry.untrusted_modules_v4` dataset. The migration path here is to create a new schema with an updated version number e.g. `untrustedModules.<VERSION>.schema.json`. The client-side will need to be modified to send payloads with the new schema version. Is this something that is being discussed or has been implemented? Otherwise the pings with this new format will be rejected at ingestion.

Additionally, the parquet schema is no longer required. The JSON schema is used to inform the ingestion system and BigQuery of the structure of the incoming data.

mreid-moz:
Re: Changing versions - it turns out [it's not easy](https://bugzilla.mozilla.org/show_bug.cgi?id=1453377) to change the version of a ping using the ["Common Ping" structure](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/common-ping.html), so we may be stuck with v4 for now. 

A couple of options:
1. Looks like you don't care about data in the old form - in this case we can change the schema, leave the version at 4, and the pipeline will reject old-format submissions as validation errors. This may require intervention from data ops (cc @whd) to drop and re-create incompatible tables in BigQuery.
2. Give the new version of the ping a new name

dblohm7:
Closing in favour of a new ping name.

PR #411, opened 2019-09-16
Require all client ids be uuids

relud:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/eng-workflow/build/build.1.schema.json
+++ b/schemas/eng-workflow/build/build.1.schema.json
@@ -64,6 +64,7 @@
     },
     "client_id": {
       "description": "A UUID to uniquely identify a client",
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
       "type": "string"
     },
     "command": {
--- a/schemas/metadata/decoded/decoded.1.schema.json
+++ b/schemas/metadata/decoded/decoded.1.schema.json
@@ -3,6 +3,7 @@
   "id": "http://jsonschema.net",
   "properties": {
     "client_id": {
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
       "type": "string"
     },
     "document_id": {
--- a/templates/eng-workflow/build/build.1.schema.json
+++ b/templates/eng-workflow/build/build.1.schema.json
@@ -64,6 +64,7 @@
     },
     "client_id": {
       "description": "A UUID to uniquely identify a client",
+      @COMMON_PATTERN_UUID_1_JSON@,
       "type": "string"
     },
     "command": {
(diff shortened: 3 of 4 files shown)


schemas/eng-workflow/build/build.1.schema.json
```diff
@@ -64,6 +64,7 @@
     },
     "client_id": {
       "description": "A UUID to uniquely identify a client",
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
```
relud: r? @cgsheeh as the original author of this field

schemas/metadata/decoded/decoded.1.schema.json
```diff
@@ -3,6 +3,7 @@
   "id": "http://jsonschema.net",
   "properties": {
     "client_id": {
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
```
relud: r? @jklukas

schemas/metadata/error/error.1.schema.json
```diff
@@ -6,6 +6,7 @@
       "type": "string"
     },
     "client_id": {
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
```
relud: r? @jklukas

relud:
@jklukas the metadata schemas won't be validated, but this still defines an expectation that client ids are uuids, which is currently enforced for all schemas except `eng-workflow.build`, for which all client ids since sept 1st are of this format.

schemas/metadata/error/error.1.schema.json
jklukas: For the error schema, it seems wrong to imply that values must follow this pattern. Imagine if a record comes to the decoder and fails validation _because_ the client_id was in an incompatible format. Currently, I believe the validation failure would cause the message to go to error output before the client_id is added as an attribute, but I could imagine wanting to try to get the errant client_id into attributes for the error message to help with debugging.

At the least, we'd expect client_id to not be populated in many cases for error messages.

schemas/metadata/decoded/decoded.1.schema.json
jklukas: This seems fine. We shouldn't have any bad client_ids make it through the decoder.

relud: oooh, good point, will remove

schemas/metadata/error/error.1.schema.json
relud: oooh, good point, will remove

schemas/eng-workflow/build/build.1.schema.json
cgsheeh: This field is generated using Python's `uuid.uuid4()` function. Testing with `re.match('^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$', str(uuid.uuid4()))` came up with a match, so I think this should be fine. r+

PR #416, opened 2019-09-17
1531556: Allow any string -- not just snake case -- for experiment names

mdboom:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1581556 for context and analysis.

Experiment names were being rejected for not being in kebab case, causing all pings for the experiment-enrolled cohort to be rejected.  This just removes the regex from experiment and branch names to avoid this sort of problem in the future.  Unless there are other, unrelated, reasons to restrict experiment names, this might be the best solution.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/glean/glean.1.schema.json
+++ b/templates/include/glean/glean.1.schema.json
@@ -54,11 +54,17 @@
         "end_time": @GLEAN_DATETIME_1_JSON@,
         "experiments": {
           "type": "object",
-          "propertyNames": @GLEAN_SHORT_ID_1_JSON@,
+          "propertyNames": {
+            "type": "string",
+            "maxLength": 30
+          },
           "additionalProperties": {
             "type": "object",
             "properties": {
-              "branch": @GLEAN_SHORT_ID_1_JSON@,
+              "branch": {
+                "type": "string",
+                "maxLength": 30
+              },
               "extra": {
                 "type": "object",
                 "properties": {
--- a/schemas/glean/baseline/baseline.1.schema.json
+++ b/schemas/glean/baseline/baseline.1.schema.json
@@ -727,7 +727,6 @@
             "properties": {
               "branch": {
                 "maxLength": 30,
-                "pattern": "^[a-z_][a-z0-9_]*$",
                 "type": "string"
               },
               "extra": {
@@ -746,7 +745,6 @@
           },
           "propertyNames": {
             "maxLength": 30,
-            "pattern": "^[a-z_][a-z0-9_]*$",
             "type": "string"
           },
           "type": "object"
--- a/schemas/glean/events/events.1.schema.json
+++ b/schemas/glean/events/events.1.schema.json
@@ -727,7 +727,6 @@
             "properties": {
               "branch": {
                 "maxLength": 30,
-                "pattern": "^[a-z_][a-z0-9_]*$",
                 "type": "string"
               },
               "extra": {
@@ -746,7 +745,6 @@
           },
           "propertyNames": {
             "maxLength": 30,
-            "pattern": "^[a-z_][a-z0-9_]*$",
             "type": "string"
           },
           "type": "object"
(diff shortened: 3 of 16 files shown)


templates/include/glean/glean.1.schema.json
```diff
@@ -54,11 +54,17 @@
         "end_time": @GLEAN_DATETIME_1_JSON@,
         "experiments": {
           "type": "object",
-          "propertyNames": @GLEAN_SHORT_ID_1_JSON@,
+          "propertyNames": {
```
Dexterp37: This also suggests we were being too permissive in the Glean SDK implementation, as we were not erroring out on these. Right? The change sounds good to me, unless it has deep implications for pipeline. @fbertsch will tell.

mdboom: That's correct -- we don't to the regex check at runtime.  But it would be hard to raise an error because, like most Glean SDK calls, it's asynchronous and returns nothing.  The best we could probably do is drop the experiment from the `experiments` object.

@fbertsch would know for sure, but I doubt it has deep implications for the pipeline -- the Presto ingestion is *already* behaving this way.  It would only change the behavior on BQ.

jklukas: I assume this ends up as a pseudo map type in BQ, in which case the experiment name ends up in a string field and can be anything. So I don't see any technical problem with the pipeline by relaxing this.

But there _is_ a data curation cost to having different experiments use different conventions. There will likely be bugs introduced in analyses where we assume one convention is being used and miss some experiments data because of it. It would be great to be able to enforce a strong convention at the SDK level, but assuming that's not possible, this may be a fight we simply have to let ourselves lose.

mdboom: I'm open to discussing whether this *should* restrict things and what those restrictions should be.  In part this perhaps my overreaction to realizing we lost 2/3 of our data for nearly a month :)

With [rejection monitoring](https://bugzilla.mozilla.org/show_bug.cgi?id=1581554) in place, client-side validation (if possible), and validation of the data that goes into an experiment, we could more safely restrict this.

I'm going to mark this as WIP for now while we figure all that out...

jklukas: I don't know what previous conventions we have in place for experiment names, so I don't have an opinion as to what the convention here should be, just that it will avoid some problems down the line if experiment names are all consistent.

That said, I would agree that losing wide swaths of pings is a bigger problem than ensuring we have clean experiment names, so I totally understand if being more permissive in validation here seems like the most pragmatic option.

fbertsch: Removing this is only removing the regex, so we're good to go on that end.

We need to have a broader discussion on what we're trying to do when we put field limitations in a schema here. are we:
1. Trying to catch pathological clients
2. Trying to prevent non schema-compatible pings from entering the pipeline downstream
3. Trying to make analysis easier by limiting field sets

IMO we are mostly trying to do (1.) and (2.). Doing (3.) is perhaps a job for downstream datasets, however I'm open to alternatives.

tl;dr: this is fine to deploy

mdboom:
A couple details: The schema used to validate experiment configuration data sets the experiment name and branch name at `maxLength` of 100.  ([See here, private repo](https://github.com/mozilla-services/remote-settings-permissions/blob/master/kinto.prod.yaml#L613))  We should at least make these match by changing one or the other length.

Dexterp37:
> 
> 
> A couple details: The schema used to validate experiment configuration data sets the experiment name and branch name at `maxLength` of 100. ([See here, private repo](https://github.com/mozilla-services/remote-settings-permissions/blob/master/kinto.prod.yaml#L613)) We should at least make these match by changing one or the other length.

Changing ours would require a change to the SDK code as well, as we truncate. We might want to leave this limit in place-for now, as it won't result in data loss. We should then file a bug to discuss that with the experiments team.

Issue #386, opened 2019-08-15
GCP: Decide what to do with tables when they are removed from this repository


fbertsch:
BQ tables are auto-generated and updated as the schema changes. Once the schema is removed from this repository, the table is dropped. We shouldn't be dropping data when a schema is removed, instead we should retain the historical data for however long the retention period is (cc @mreid-moz).

Option 1: We keep the table in the same location, allowing for the small possibility that a new schema will be written to that location (we could add automatic checking for these, it would be especially bad if the schemas weren't compatible).

Option 2: We move that data to a historical location, such that we know it's not being updated and new data is not flowing in, and a new ping can replace it; however it will remain queryable (for the duration of the retention period).

I'm leaning towards (2.), but the downside is we either need to manually change queries to point to the new location, or move views to point there (and version views for the new data).

whd:
I'm also vaguely pro (2), as it would be nice to more generally have a concept of "deprecated" or "historical" data. There are many ping types we have collected over the years that we probably don't care about processing (e.g. appusage). In https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/334 we decided which pings we care about and don't care about for the purposes of schemas. I do like the idea of being able to make an active decision to no longer process a ping type by removing its schemas.

I don't like (1) because it means the state of production and the state of generated-schemas would not be precisely the same. In this case I would prefer that we never drop schemas (hitherto the standard practice). This goes back to developing a notion of "deprecated" data, which doesn't exist for ingestion currently.

fbertsch:
You make a good point about (1.) not matching the `generated-schemas` branch. A tentative plan for (2.) could be:

0. Schema is removed from this repository
1. Deploy notices the schema is now missing
  a. Copy the data to a historical location (TBD)
  b. Update the view that points to the historic data (related, [auto-deployed views](https://github.com/mozilla/bigquery-etl/issues/291). If the view definition is not auto-deployed manual intervention could be required)
  c. Drop the prod table
  d. Deploy gcp-ingest with no json schema

I do believe that makes https://github.com/mozilla/bigquery-etl/issues/291 a dependency here.

jklukas:
I'm not convinced that we gain much by actually moving the table to a historical location. I'd like to see a way of marking a docType as deprecated, perhaps via metadata in the JSON schema file itself.

[...]

PR #422, opened 2019-09-24
Remove telemetry_sync_v5

acmiyaguchi:
`sync` v5 was a mistake, as per [bug 1410145](https://bugzilla.mozilla.org/show_bug.cgi?id=1410145). There is a surprising number of pings that are arriving on this doc version though (https://sql.telemetry.mozilla.org/queries/64998/source).

We should remove this table from the BigQuery dataset too, since it's causing `telemetry.sync` to alias to `telemetry.sync_v5`.


whd:
CI is still failing on the third-party-modules change, for some reason.

Given the relatively high volume, I want somebody (@mreid-moz?) to make a policy decision that we can/should delete this table. We haven't sorted out https://github.com/mozilla-services/mozilla-pipeline-schemas/issues/386  yet and probably won't for some time. In the current model, we will remove these tables in a mostly non-reversible fashion.

An alternative solution would be to special-case the sync ping unversioned view to point at v4. Absent any compelling reasons I prefer this approach, since it appears we're using a [hard-coded view](https://github.com/mozilla/bigquery-etl/blob/master/templates/telemetry/sync/view.sql#L4) at the moment anyway.

mreid-moz:
Either way, we need to fix the view to point to v4, so I sent https://github.com/mozilla/bigquery-etl/pull/401.

I've asked for confirmation on deleting the data (as well as adding my understanding of the implications) in [Bug 1572744#c4](https://bugzilla.mozilla.org/show_bug.cgi?id=1572744#c4).

acmiyaguchi:
Closing per https://bugzilla.mozilla.org/show_bug.cgi?id=1572744#c9

PR #430, opened 2019-10-03
Bug 1585147 - Add schema for messaging system

ncloudioj:
Checklist for reviewer:

For glean changes:

This patch migrates the `assa_router_events_daily` table from Tiles. I've splitted it into three new tables, `cfr`, `onboarding`, and `snippets`, hence the `source` column from the original scheam has been dropped.

N.B. I didn't introduce the composite structure here to support the messaging batching as most of existing pings are triggered at different occations, also it requires some changes on the client side in order to support batching. I think we can revisit this feature as needed in the future. 

@jklukas r?
@crankycoder mind taking the CFR schema?
@jgmize you can find the schmea for Snippets here :)


validation/messaging-system/cfr/cfr.1.prerelease.pass.json
```diff
@@ -0,0 +1,13 @@
+{
+  "client_id": "94642acb-4996-034b-916c-147da723cc41",
+  "impression_id": "n/a",
```
jklukas: I think these new test cases are being skipped due to having an extra level of directories. The expected pathing is different here from `templates` and there should be no intermediate `cfr` directory.

I think these test will fail because we have regexes on impression_id and client_id that require UUIDs and should fail on "n/a".

The value of "n/a" may cause problems in the pipeline, too, since we specially look for `client_id` if it exists in the payload in order to calculate sample_id. We may make an assumption that if it's present it's a valid UUID.

ncloudioj: Nice catch. Since `impression_id` and `client_id` form a `oneOf` relation here, we should just omit the other one if one is present in the payload.

schemas/messaging-system/cfr/cfr.1.schema.json
```diff
@@ -0,0 +1,69 @@
...
+      ]
+    },
+    {
+      "required": [
+        "bucket_id"
+      ]
+    }
+  ],
+  "properties": {
+    "addon_version": {
```
crankycoder: Is this addon_version required anymore?  I think this is a legacy field from when this was a TestPilot project?

schemas/messaging-system/cfr/cfr.1.schema.json
```diff
@@ -0,0 +1,69 @@
...
+      "type": "integer"
+    },
+    "release_channel": {
+      "type": "string"
+    },
+    "shield_id": {
+      "description": "A semicolon separated string to store a list of Shield study IDs",
+      "type": "string"
+    },
+    "value": {
```
crankycoder: Can we use a better name like `event_context`?

schemas/messaging-system/cfr/cfr.1.schema.json
ncloudioj: Since we could export multiple versions in a single release cycle, I think this is still useful for us to investigate the potential issues.

schemas/messaging-system/cfr/cfr.1.schema.json
ncloudioj: Yep, we can do that.

schemas/messaging-system/cfr/cfr.1.schema.json
```diff
@@ -0,0 +1,69 @@
...
+      "description": "A string value that describes this event",
+      "type": "string"
+    },
+    "version": {
+      "type": "string"
+    }
+  },
+  "required": [
+    "client_id",
```
crankycoder: `client_id` is only going to be populated in pre-release versions.  I don't think it should be in the `required` list.

ncloudioj: This is now fixed!

ncloudioj:
@jklukas all review comments have been resolved. Mind taking another scan?

validation/messaging-system/cfr.1.both_ids.fail.json
```diff
@@ -0,0 +1,13 @@
+{
+  "impression_id": "{94642acb-4996-034b-916c-147da723cc41}",
+  "client_id": "94642acb-4996-034b-916c-147da723cc41",
```
jklukas: Thank you for adding the test case for failing if both IDs are present 👍

Issue #413, opened 2019-09-16
Add checks to CI for potential gotchas


sunahsuh:
- Add special rule for pinned telemetry schema version (currently 4)
- Enforce directory structure for all schemas

acmiyaguchi:
These conditions are being asserted.

Version numbers in testing:
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/698fdb13e58338def105ac8ac34145571f65e676/tests/test_telemetry_version.py#L20-L26

Directory path depth
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/698fdb13e58338def105ac8ac34145571f65e676/tests/conftest.py#L15-L21

Validation document depth
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/698fdb13e58338def105ac8ac34145571f65e676/tests/conftest.py#L37-L43

PR #433, opened 2019-10-04
Add checks for telemetry version number to CI

acmiyaguchi:
This tackles 1/2 items in issue #413, which includes checking that the version number is set up correctly. The goal of this change is to add a location where more checks can be added.

Checklist for reviewer:

For glean changes:


acmiyaguchi:
That should be fine, as long as the `core` ping doesn't explicitly set the `version` field to something that isn't `4`, it should be valid.

PR #435, opened 2019-10-11
Add some missing fields to the environment and main payload.

mreid-moz:
This adds some of the missing fields from https://github.com/mozilla/mozilla-schema-generator/issues/47

A couple of things to note - I allowed for nulls in some, but not all, of the boolean fields based on nearby fields, and I couldn't find a real example of lateWrites, so I'm going by what's in the [main ping docs](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/main-ping.html#latewrites).

Checklist for reviewer:

### DIFF
--- a/templates/include/telemetry/mainPayload.1.schema.json
+++ b/templates/include/telemetry/mainPayload.1.schema.json
@@ -127,38 +127,20 @@
           "description": "entries in the format [module name, breakpad identifier]",
           "items": {
             "type": "array",
-            "items": [
-              {
-                "type": "string",
-                "description": "module name"
-              },
-              {
-                "type": "string",
-                "description": "breakpad identifier"
-              }
-            ],
+            "items": { "type": "string" },
             "minItems": 2,
             "maxItems": 2
           }
         },
         "stacks": {
           "type": "array",
-          "description": "a list of stacks",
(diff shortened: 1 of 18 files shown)


templates/include/telemetry/mainPayload.1.schema.json
```diff
@@ -109,7 +120,51 @@
...
       "type": "object",
-      "properties": { }
+      "description": "This sections reports writes to the file system that happen during shutdown. The reported data contains the stack and the file names of the loaded libraries at the time the writes happened.",
+      "properties": {
+        "memoryMap": {
+          "type": "array",
+          "description": "entries in the format [module name, breakpad identifier]",
+          "items": {
+            "type": "array",
+            "items": [
```
fbertsch: This representation will cause these to be anonymous structs in BQ. If we set it simply to `"items": {"type": "string"}`, it will be a string array. Does that seem reasonable since the types are all the same?

fbertsch: For this and `stacks` below.

mreid-moz: Yeah, that's cleaner. Good suggestion!

PR #436, opened 2019-10-14
Add all allowed process types to main ping

sunahsuh:
Doesn't look like I can actually add this to Mark's PR since it's from a personal fork. Fingers crossed we avoid merge conflicts!

Checklist for reviewer:

### DIFF
--- a/schemas/telemetry/first-shutdown/first-shutdown.4.schema.json
+++ b/schemas/telemetry/first-shutdown/first-shutdown.4.schema.json
@@ -1590,6 +1590,590 @@
               },
               "type": "object"
             },
+            "dynamic": {
+              "properties": {
+                "events": {
+                  "items": {
+                    "items": [
+                      {
+                        "minimum": 0,
+                        "type": "integer"
+                      },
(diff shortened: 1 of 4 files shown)


jklukas:
Is it possible/easy to see what changes will be made to the BQ schema as a result of this. It looks like there are ~400 fields associated with the `parent` process; if the number is similar for other processes, we're adding more than 1000 fields here. So I have some concern about how much room we have before hitting the 10,000 column limit for main ping.

sunahsuh:
@fbertsch would know in more detail but – all histograms.json and scalars.yaml have a "recorded in processes" property that specify which ones they might possible appear under. From a cursory search, we're looking at three additional metrics (plus the events tuples for the three new processes)

sunahsuh:
Huh, what's the bulk of the new probes? I only saw a few recorded in `socket`, basically

fbertsch:
The killers are `all`, `all_children`, and `all_childs`. Who knows if they ever _actually_ record in those but we have to assume that is the case.

PR #445, opened 2019-11-07
Add new Firefox Voice telemetry fields

ianb:
Voice bug: https://github.com/mozilla/firefox-voice/issues/546

Adds to ping:
* Date of first installation
* Number of tabs open
* Local time of day (hour)
* Country
* Server timing for DeepSpeech
* Audio recording time
* Performance measure (time) post-processing

~Also removes some fields related to a component (intent parser) that we decided not to implement separately.~

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/voice/voice.4.schema.json
+++ b/schemas/telemetry/voice/voice.4.schema.json
@@ -1130,9 +1130,33 @@
     },
     "payload": {
       "properties": {
+        "audioTime": {
+          "description": "Time in milliseconds of the recording",
+          "title": "Amount of spoken time",
+          "type": "integer"
+        },
+        "country": {
+          "title": "User country code",
+          "type": "string"
+        },
+        "deepSpeechConfidence": {
+          "maximum": 1,
+          "minimum": 0,
+          "title": "Confidence of DeepSpeech text-to-speech transcription",
+          "type": "number"
+        },
         "deepSpeechMatches": {
           "title": "Did the DeepSpeech transcription match with Google's?",
           "type": "boolean"
+        },
+        "deepSpeechServerTime": {
+          "description": "Milliseconds for the DeepSpeech transcription to complete and respond",
+          "title": "Response time of DeepSpeech server",
+          "type": "integer"
+        },
+        "extensionInstallDate": {
+          "title": "Timestamp when the extension was installed",
+          "type": "integer"
         },
         "extensionInstallationChannel": {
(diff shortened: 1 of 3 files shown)


schemas/telemetry/voice/voice.4.schema.json
```diff
@@ -1218,23 +1236,29 @@
...
+        "intentTime": {
+          "description": "Milliseconds between getting the utterance and execution the intent",
+          "title": "Time to execute the intent",
+          "type": "number"
+        },
         "internalError": {
           "title": "If there was an unexpected error, the name of that error",
           "type": "string"
         },
-        "serverErrorIntentParser": {
```
acmiyaguchi: Unfortunately, schemas are append/update only. Removing columns from a table is not well defined, and requires dropping all of the existing data or updating the version number of the schema.

templates/telemetry/voice/voice.4.schema.json
```diff
@@ -141,6 +140,21 @@
           "type": "number",
           "minimum": 0
         },
+        "localHour": {
+          "title": "Local hour of action",
+          "description": "The hour, 0-23, in the local timezone when the action was run",
+          "type": "number"
```
acmiyaguchi: Somewhat of an implementation detail of how the schemas are transformed into table schemas, but you can use `integer` here so the columns are treated as `INT64` instead of `FLOAT64`.

schemas/telemetry/voice/voice.4.schema.json
ianb: Fixed. FYI, serverTimeIntentParser moved during the rebuilds, so it appears deleted but is inserted elsewhere.

ianb:
I put the fields back in, ready for re-review.

ianb:
I was mostly copying from other examples where I saw numbers, so admittedly I haven't thought carefully about it. Are updates to column types OK, or is that problematic like deleting columns is?

acmiyaguchi:
There isn't much documentation about this at the moment, so what you have is very reasonable. Existing columns are immutable, so once the column type has been chosen, it's very difficult to change. But to reiterate again, I don't think there will be any problems caused by the choice floats/ints here (but this is the time to change it).

ianb:
I set some of the new fields to be integers, and added one more field (deepSpeechConfidence) that I had forgotten about before. I think it's ready to land.

ianb:
If someone could merge this I would appreciate that.

acmiyaguchi:
@ianb I apologize for the delay, merging now.

PR #450, opened 2019-11-20
Add events to Activity-Stream

ncloudioj:
Checklist for reviewer:

For glean changes:

This ports `Tiles.assa_events_daily` to here.

@jklukas r?

### DIFF
--- a/schemas/activity-stream/events/events.1.schema.json
+++ b/schemas/activity-stream/events/events.1.schema.json
@@ -1,0 +1,73 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "properties": {
+    "action_position": {
+      "description": "A zero based integer indicating the position of this event",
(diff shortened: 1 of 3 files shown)


jklukas:
Before merging, one question: are events emitted discretely, or do they happen in groups? In particular, the `action_position` field makes it sound like we have groups of events. If so, I'd much prefer to emit nested payloads like we do for `impression-stats` which will be substantially more efficient in the pipeline and in storage.

With the ETL harness we have in place now, it would be trivial to provide an deploy a view on top of this table that does the unnesting; we could make that available perhaps at `activity_stream.event` in BQ.

ncloudioj:
Events are emitted discretely, and I don't think the client would produce multiple events in a short period of time. The `action_position` field merely records the position information, e.g. one clicked on the first card in Topsites, which allows us to understand the user engagement for different placements.

PR #460, opened 2019-12-11
Remove all application-specific glean schemas

jklukas:
The mozilla-schema-generator now handles ingesting the single glean baseline
schema and generating all the relevant per-application schemas for validation,
so it is no longer necessary for this repository to house glean application
schemas on the master branch.

The glean/baseline schema is [referenced directly in m-g-s](https://github.com/mozilla/mozilla-schema-generator/blob/83d1292474a158652c582255ff8780c2ca247015/mozilla_schema_generator/glean_ping.py#L23)
so that is the only schema we keep on master here.

This also allows us to enforce that ping names may not contain underscores,
which requires also removing the `templates/heka` folder which should no
longer be relevant after teardown of the AWS pipeline.

Checklist for reviewer:

For glean changes:


jklukas:
@fbertsch Did I correctly understand we can do this now?

fbertsch:
@jklukas indeed we can. We need on version of it around to use in the schema-generator. [Currently that is the baseline ping](https://github.com/mozilla/mozilla-schema-generator/blob/master/mozilla_schema_generator/glean_ping.py#L23), but it seems maybe a bit confusing to call it that - we could instead move it to `glean/glean`

jklukas:
> @jklukas indeed we can. We need on version of it around to use in the schema-generator. [Currently that is the baseline ping](https://github.com/mozilla/mozilla-schema-generator/blob/master/mozilla_schema_generator/glean_ping.py#L23), but it seems maybe a bit confusing to call it that - we could instead move it to `glean/glean`

I agree on glean/glean. I'll add that as well here, so for the time being we'll have both glean/baseline and glean/glean. Then we can update the schema generator to use glean/glean, and then finally remove baseline here.

jklukas:
Removed glean parquet schemas, added glean/glean ping, updated tests. Tests are now passing and this is ready for final review.

jklukas:
> r+, but the sync ping tests were changed to having `-` rather than `_`, but e.g. `invalid_use_counter` was not. Was that purposeful?

The casing within test case names does not have any technical effect. I chose to move the doctype into the "description" portion of the test case name for test cases that used to be named simply with the source doc type, and in that case I updated to use the hyphenated "correct" version of the doctype. For the tests that had descriptions involving underscores, I left that alone.

PR #465, opened 2019-12-17
Bug 1604666 - Add `payload.crashId` in telemetry.crash_v4

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1604666)

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/telemetry/crash/crash.4.schema.json
+++ b/templates/telemetry/crash/crash.4.schema.json
@@ -17,7 +17,6 @@
         },
         "crashId": {
           "type": "string",
-          @COMMON_PATTERN_UUID_1_JSON@,
           "description": "Optional, ID of the associated crash. Added to schema in bug 1604666."
         },
         "hasCrashEnvironment": {
--- a/templates/telemetry/crash/crash.4.schema.json
+++ b/templates/telemetry/crash/crash.4.schema.json
@@ -16,7 +16,9 @@
           "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
         },
         "crashId": {
-          "type": "string"
+          "type": "string",
+          @COMMON_PATTERN_UUID_1_JSON@,
+          "description": "Optional, ID of the associated crash. Added to schema in bug 1604666."
         },
         "hasCrashEnvironment": {
           "type": "boolean"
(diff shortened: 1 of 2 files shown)


templates/telemetry/crash/crash.4.schema.json
```diff
@@ -15,6 +15,9 @@
           "type": "string",
           "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
         },
+        "crashId": {
+          "type": "string"
```
jklukas: Is crashId expected to be a UUID? Should we enforce that with a pattern?

jklukas: Is crashId always present, or is it potentially null?

fbertsch: Tests for these cases would be good.

acmiyaguchi: It does appear to be a UUID, but I'm not sure if we would start to reject any pings if we enforced the validation. I'll add this and check the integration report. 

The documentation for thing ping marks the field as optional: https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/crash-ping.html

The field is already included as part of the passing document here: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/e8226ee26926df54fb641cee02c4e91b0d346a5b/validation/telemetry/crash.4.sample.pass.json#L188

acmiyaguchi: It looks like there's a small increase in the number of rejected pings, based on the document sample:

```diff
--- 
+++ 
@@ -168,7 +168,7 @@
         "error_rate": 0.0
     },
     "telemetry.crash.4": {
-        "error_rate": 1.9
+        "error_rate": 2.0
     },
     "telemetry.deletion.4": {
         "error_rate": 0.0
```

acmiyaguchi:
Yes, it also seems like the `crashId` is not the primary way to correlate crashes with socorro. Instead, the sha256 hash associated with the crash dump subsumes this id. This means the `crashId` field might not actually be useful for anything, so I don't think there's necessarily a case to be strict about the values in this field.

PR #478, opened 2020-01-08
Move devices directly under payload in sync pings

### LINKED ISSUE #473: update sync ping schema more :) `devices` array is now directly under `payload`
In the sync ping, [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1582263) moved the `devices` field (nullable array of structs/maps) to be directly under `payload` rather than nested within each member of the `syncs` array. we should update the pipeline schema to reflect this.

it will help with not having to write ugly queries [like this](https://sql.telemetry.mozilla.org/queries/67386/source)

jklukas:
Closes #473

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/sync/sync.4.schema.json
+++ b/schemas/telemetry/sync/sync.4.schema.json
@@ -64,6 +64,28 @@
             "null"
           ]
         },
+        "devices": {
+          "items": {
+            "additionalProperties": false,
+            "properties": {
+              "id": {
+                "pattern": "^[0-9a-f]{64}$",
+                "type": "string"
+              },
+              "os": {
+                "type": [
+                  "string",
+                  "null"
+                ]
+              },
+              "version": {
+                "type": "string"
+              }
+            },
+            "type": "object"
+          },
+          "type": "array"
+        },
         "discarded": {
           "minimum": 1,
           "type": "integer"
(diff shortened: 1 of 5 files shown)


jklukas:
cc @irrationalagent

irrationalagent:
@thomcc ref [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1582263) I'm 90% sure Jeff is doing the right thing here, but can you just confirm that that we should be moving the sync ping `devices` field to be directly under `payload` instead of under `syncs`? 

asking because the [.rst file](https://searchfox.org/mozilla-central/source/toolkit/components/telemetry/docs/data/sync-ping.rst) still has it under `syncs` but the [json schema](https://searchfox.org/mozilla-central/source/services/sync/tests/unit/sync_ping_schema.json) seems to have it moved up, as per the bug.

jklukas:
I can confirm that there are real pings with devices directly under payload, since that is appearing under additional_properties in some pings (looks like it's a minority of pings):

```
SELECT
  COUNT(*)
FROM
  `moz-fx-data-shared-prod.telemetry_stable.sync_v4`
WHERE
  DATE(submission_timestamp) = "2020-01-06"
  AND json_EXTRACT(additional_properties,
    '$.payload.devices') IS NOT NULL
```

And to be clear, the old location for `devices` remains in the schema, so we'll be ingesting it whether it occurs in either place.

Looks to me like the `.rst` docs need to be updated to reflect current reality.

irrationalagent:
Its probably the case that clients who are still on versions older than the change are still submitting the devices list in the old location. Thanks for looking Jeff

PR #481, opened 2020-01-13
bug 1604312 - Scalars can now appear in 'deletion-request' pings.

chutten:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/validation/telemetry/deletion-request.4.payload.pass.json
+++ b/validation/telemetry/deletion-request.4.payload.pass.json
@@ -15,12 +15,10 @@
     "channel": "default"
   },
   "clientId": "c4582ba1-79fc-1f47-ae2a-671118dccd8b",
-  "payload": { 
-    "processes": {
+  "payload": {
+    "scalars": {
       "parent": {
-        "scalars": {
-          "browser.engagement.unfiltered_uri_count": 782
-        }
+        "browser.engagement.unfiltered_uri_count": 782
       }
     }
   }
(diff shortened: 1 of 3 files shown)


validation/telemetry/deletion-request.4.payload.pass.json
```diff
@@ -15,13 +15,9 @@
...
-  "payload": { 
-    "processes": {
-      "parent": {
-        "scalars": {
-          "browser.engagement.unfiltered_uri_count": 782
-        }
-      }
+  "payload": {
+    "scalars": {
+      "browser.engagement.unfiltered_uri_count": 782
```
relud: Based on the schema I would expect this scalar to be located at `payload.scalars.parent`, and this is only valid because `"additionalProperties": false` isn't set on `scalars`.

chutten: Good point! This is a bug in this test file.

chutten:
> 
> 
> Is the location of scalars in deletion request pings intentionally different between from main pings? e.g. `payload.scalars.parent.*` instead of `payload.processes.parent.scalars.*`?

I asked Frank if it mattered and he said [it didn't](https://bugzilla.mozilla.org/show_bug.cgi?id=1604312#c2). So I gave it a think.

Since there shouldn't ever be non-scalar (and specifically non-string-scalar) values in this ping there is no value in making this look like the "main" ping. This is a little more glean-esque (which is the bright future we're heading towards), too.

(( And as for the real reason, this is the format the snapshotting function outputs, so it saves me code on the client-side : D ))

chutten:
RFAL

jklukas:
Is anything holding up merging this change?

chutten:
Not from my end. @mreid-moz ?

PR #491, opened 2020-02-03
1612940: Allow "null" extras in experiment

mdboom:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/glean/glean/glean.1.schema.json
+++ b/schemas/glean/glean/glean.1.schema.json
@@ -752,12 +752,19 @@
                 "type": "string"
               },
               "extra": {
-                "properties": {
-                  "type": {
-                    "type": "string"
+                "anyOf": [
+                  {
+                    "type": "null"
+                  },
+                  {
+                    "properties": {
+                      "type": {
+                        "type": "string"
+                      }
+                    },
+                    "type": "object"
                   }
-                },
-                "type": "object"
+                ]
               }
             },
             "required": [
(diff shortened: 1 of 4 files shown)


mdboom:
Importantly, I don't know if this has implications for the database schema.

schemas/glean/glean/glean.1.schema.json
```diff
@@ -752,12 +752,19 @@
                 "type": "string"
               },
               "extra": {
-                "properties": {
-                  "type": {
-                    "type": "string"
+                "anyOf": [
```
jklukas: I don't know if the jsonschema-transpiler supports anyOf. In testing locally, it looks like this ended up turning `extra` into a STRING field rather than a RECORD. @acmiyaguchi should be able to comment on whether we have a good way of expressing an object that can be null.

jklukas: It looks like `oneOf` is supported and produces identical BQ schema output. Does `oneOf` make sense in this situation?

acmiyaguchi: `"type": ["null", "object"]` is also a valid construct that produces the same schema. `"oneOf": [...]` would also make sense here, as these two fragments are treated equivalently.

https://gist.github.com/acmiyaguchi/a273dfbd8438769f6047566021cfa72d

mdboom: `oneOf` works for me.  I'll update this.

jklukas:
> I'm not sure if schema deploys will be done before 2020-02-06, though.

We may be able to ask other data ops folks if they can push a prod schema deploy while whd is out. As long as there's nothing weird that shows up, I think there's a process in place.

PR #498, opened 2020-02-07
Bug 1613225 Add descriptions to payload_bytes_* fields

jklukas:
In particular, we draw attention to the fact that IP address fields
in error tables will always be null.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/metadata/error/error.1.schema.json
+++ b/schemas/metadata/error/error.1.schema.json
@@ -3,18 +3,22 @@
   "id": "http://jsonschema.net",
   "properties": {
     "args": {
+      "description": "query parameter part of the submission URI, i.e. \"v=4\"",
       "type": "string"
     },
     "client_id": {
       "type": "string"
     },
     "content_length": {
+      "description": "Content-Length header",
       "type": "string"
     },
     "date": {
+      "description": "Date header",
       "type": "string"
     },
     "dnt": {
+      "description": "DNT header",
       "type": "string"
     },
     "document_namespace": {
@@ -48,6 +52,7 @@
       "type": "string"
     },
     "host": {
+      "description": "Hostname part of the submission URI, i.e. \"incoming.telemetry.mozilla.org\"",
       "type": "string"
     },
     "input": {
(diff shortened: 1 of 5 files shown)


schemas/metadata/raw/raw.1.schema.json
```diff
@@ -3,53 +3,69 @@
...
     "x_forwarded_for": {
+      "description": "X-Forwarded-For header containing intermediate IP addresses",
       "type": "string"
     },
     "x_pingsender_version": {
+      "description": "X-Pingsender-Version header",
       "type": "string"
     },
     "x_pipeline_proxy": {
+      "description": "Time that the AWS to GCP tee service received the message",
```
whd: Or any non-null value (usually ["data-noclip"](https://github.com/mozilla-services/cloudops-deployment/blob/master/projects/data-noclip/puppet/modules/noclip/templates/http_app.erb#L22)) to indicate this data was proxied from AWS. Previously it was populated with the heka timestamp in the same timestamp format as submission_timestamp but there's no more heka. I grepped gcp-ingestion and it doesn't appear we document this anywhere.

templates/metadata/raw/raw.1.schema.json
```diff
@@ -3,7 +3,19 @@
...
+      "type": "string",
+      "description": "Source IP address, usually a load balancer"
+    },
+    "x_forwarded_for": {
+      "type": "string",
+      "description": "X-Forwarded-For header containing intermediate IP addresses"
+    },
+    "x_pipeline_proxy": {
+      "type": "string",
+      "description": "Time that the AWS to GCP tee service received the message"
```
whd: ditto above

schemas/metadata/error/error.1.schema.json
```diff
@@ -48,6 +52,7 @@
       "type": "string"
     },
     "host": {
+      "description": "Hostname part of the submission URI, i.e. \"incoming.telemetry.mozilla.org\"",
```
whd: I'm not sure if this is technically accurate in all cases (I've seen Host header and URI mismatches before) but it probably doesn't matter.

jklukas:
> Nit: you should replace all your "i.e."s (that is) with "e.g."s (for example).

I thought through this but apparently my mnemonic failed me. I care about these things. Such shame I feel.

PR #499, opened 2020-02-08
Replace validation testing with JSON Schema v7 compatible libraries 

### LINKED ISSUE #332: Use everit-org/json-schema for schema validation tests
The GCP pipeline is using the [everit-org/json-schema](https://github.com/everit-org/json-schema) Java library for schema validation and we should move testing of schemas to rely on that same library.

Currently, we test using hindsight (which I believe is using rapidjson for schema validation). We could run both testing implementations while we have both pipelines running concurrently.

[Rapidjson's schema validator](http://rapidjson.org/md_doc_schema.html) implements only v4 of the JSON schema specification and it ignores the "format" keyword. Moving completely to everit-org/json-schema would allow us to consider upgrading to the current v7 draft and taking better advantage of the "format" keyword for ensuring correct date and time types.

acmiyaguchi:
This will fix #332, and also introduces the `jsonschema` python package in case testing is done locally. There are 3 failing tests, 1 in python and 2 in java. 

I've tested this on my local machine (without java configured, because [`pyjnius`](https://github.com/kivy/pyjnius) was not playing well with my macOS instance) which will skip the java tests, and in the docker image which can run either. When running in docker, I recommend the `-n` flags. 

command | stats
-|-
local, `pytest` | 1 failed, 166 passed, 167 skipped in 8.25s
`scripts/mps-test -k python` |  1 failed, 166 passed, 167 deselected in 157.60s (0:02:37)
`scripts/mps-test -k java` | 2 failed, 165 passed, 167 deselected in 141.69s (0:02:21)

The `scripts/mps-build` and `scripts/mps-test` are helpers for this workflow. 

Current errors:

```
test_validation_pass_python[glean/glean.1.baseline.pass.json]
E           jsonschema.exceptions.ValidationError: 'examples.labeled_timing_distribution_example' does not match '^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})+$'
E           
E           Failed validating 'pattern' in schema['properties']['metrics']['properties']['labeled_timing_distribution']['propertyNames']:
E               {'maxLength': 61,
E                'pattern': '^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})+$',
E                'type': 'string'}
E           
E           On instance['metrics']['labeled_timing_distribution']:
E               'examples.labeled_timing_distribution_example'
```

```
test_validation_pass_java[glean/glean.1.baseline.pass.json]
E   jnius.JavaException: JVM exception occurred: #/metrics/labeled_timing_distribution/examples.labeled_timing_distribution_example: string [examples.labeled_timing_distribution_example] does not match pattern ^[a-z_][a-z0-9_]{0,29}(\.[a-z_][a-z0-9_]{0,29})+$
```

```
test_validation_pass_java[telemetry/untrusted-modules.4.unknownmodule.pass.json]
E   jnius.JavaException: JVM exception occurred: #/payload/combinedStacks/stacks/0/2/1: expected type: Number, found: String
```

There are still a few more things to do, but this should be in a reviewable state. 

Checklist for reviewer:

For glean changes:


tests/test_validation.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def test_validation_fail_python(schemas, qualifier, failing_example):
+    assert qualifier in schemas, f"{qualifier} missing from schemas"
+    with pytest.raises(ValidationError):
+        validate(failing_example, schemas[qualifier])
+
+
+@runif_java_configured
+def test_validation_pass_java(schemas, qualifier, passing_example):
```
acmiyaguchi: This should get split out into a separate file to avoid conflating dependencies, i.e. `pytest tests/test_validation_python.py` should run without java dependencies.

tests/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+        with path.open() as fp:
+            schemas[qualifier] = json.load(fp)
+    return schemas
+
+
+# TODO: test the following cases
+# - example in wrong place
+# - example with the wrong name
+# - schema that doesn't exist
+def pytest_generate_tests(metafunc):
```
acmiyaguchi: One interesting idea here is to generate tests to make sure that every schema has an example, but that's for another PR.

Dockerfile
```diff
@@ -30,14 +33,16 @@ RUN yum -y install \
...
 
-RUN rm -fr release/; mkdir release/
-
+RUN rm -fr /app/release; mkdir /app/release
 WORKDIR /app/release
+RUN cmake ..; make
 
-RUN ln -s /usr/bin/cmake3 /usr/bin/cmake; cmake ..; make
-
-CMD ctest3 -V -C hindsight
```
jklukas: Is this PR removing all use of hindsight? If so, we should remove the dependencies in this file too so that we no longer pull in the hindsight libraries.

pom.xml
```diff
@@ -0,0 +1,19 @@
...
+    <groupId>com.mozilla.telemetry</groupId>
+    <artifactId>mozilla-pipeline-schemas</artifactId>
+    <version>0.1-SNAPSHOT</version>
+    <dependencies>
+        <dependency>
+            <groupId>com.github.everit-org.json-schema</groupId>
+            <artifactId>org.everit.json.schema</artifactId>
+            <version>1.12.1</version>
+        </dependency>
+    </dependencies>
```
jklukas: We should probably also specify the same explicit version of org.json that we're using in ingestion-beam.

Dockerfile
acmiyaguchi: I think removing hindsight + parquet removal should come in a separate PR that also introduces a GRAVEYARD.md.

pom.xml
acmiyaguchi: This looks to be the case. I've added to a comment pointing to pom.xml in gcp-ingestion to be clear about this.

acmiyaguchi:
@relud r?, in particular with respect to the pytest configuration for generating tests.

tests/test_validation_java.py
```diff
@@ -0,0 +1,58 @@
...
+        func
+    )
+
+
+@runif_java_configured
+def test_validation_pass_java(schemas, qualifier, passing_example):
+    assert qualifier in schemas, f"{qualifier} missing from schemas"
+
+    JSONObject = autoclass("org.json.JSONObject")
+    SchemaLoader = autoclass("org.everit.json.schema.loader.SchemaLoader")
```
relud: why are these classes being loaded in tests, rather than using the result from lines 18 and 19?

README.md
```diff
@@ -70,18 +73,42 @@ To run the tests:
...
+# run java tests only (if Java is configured)
+pytest -k java
+```
+
+If you would like to run validation against
+[`everit-org/json-schema`](https://github.com/everit-org/json-schema) used in
+[mozilla/ingestion-beam](https://mozilla.github.io/gcp-ingestion/ingestion-beam/),
+either run the docker container or install the java dependencies.
+
+```pytest
```
relud: `s/pytest/bash/`?

tests/test_validation_java.py
acmiyaguchi: Good catch. The reason is that I wrote the tests first, and then added the tests for ensuring the libraries existed.

pom.xml
acmiyaguchi: Understood, I misread and took `org.json` to be the same thing as `org.everit.json.schema`.

README.md
```diff
@@ -70,18 +73,42 @@ To run the tests:
...
+export JAVA_HOME=...
+
+# resolves and copies jars into `target/dependency`
+mvn dependency:copy-dependencies
+
+# check that tests are not skipped
+pytest -k java -n 8
```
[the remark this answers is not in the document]
acmiyaguchi: Thanks for verifying this on a local machine :+1:

tests/test_validation_java.py
```diff
@@ -0,0 +1,52 @@
...
+    [p.resolve().as_posix() for p in (ROOT / "target").glob("**/*.jar")]
+)
+
+# try importing java libraries
+ENABLE_JAVA_TESTS = False
+JAVA_CONFIGURED_REASON = "failed to configure java"
+try:
+    from jnius import JavaException, autoclass
+
+    autoclass("org.json.JSONObject")
```
relud:
```suggestion
    JSONObject = autoclass("org.json.JSONObject")
```

tests/test_validation_java.py
```diff
@@ -0,0 +1,52 @@
...
+    autoclass("org.everit.json.schema.loader.SchemaLoader")
```
relud:
```suggestion
    SchemaLoader = autoclass("org.everit.json.schema.loader.SchemaLoader")
```

PR #508, opened 2020-03-02
Bug 1603536 Add descriptions for environment from probe dictionary

scholtzan:
I wrote a small script to extract descriptions from [probe-dictionary](https://github.com/mozilla/probe-dictionary/blob/master/environment.json) and add them to corresponding fields that were missing the description. 

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/telemetry/environment.1.schema.json
+++ b/templates/include/telemetry/environment.1.schema.json
@@ -10,25 +10,43 @@
             "type": "object",
             "properties": {
               "blocklisted": {
-                "type": "boolean"
+                "type": "boolean",
+                "description": "Whether or not the add-on appears in the blocklist. This field is only available after the 'sessionstore-windows-restored' topic is notified."
               },
               "description": {
-                "type": [ "string", "null" ]
+                "type": [
+                  "string",
+                  "null"
+                ]
               },
               "name": {
-                "type": "string"
+                "type": "string",
+                "description": "The add-on name, limited to 100 characters. This field is only available after the 'sessionstore-windows-restored' topic is notified."
               },
               "userDisabled": {
-                "type": [ "boolean", "integer" ]
+                "type": [
+                  "boolean",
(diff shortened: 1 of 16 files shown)


templates/include/telemetry/environment.1.schema.json
```diff
@@ -10,25 +10,43 @@
...
+                  "number"
+                ],
+                "description": "The version of the add-on. This field is available at startup."
               },
               "scope": {
-                "type": [ "integer", "null" ]
+                "type": [
+                  "integer",
+                  "null"
+                ],
```
scholtzan: Writing the JSON with descriptions to a new file changed the formatting for some fields. However, it seems there is no real formatting standard right now. so it's probably fine?

jklukas:
Is this intended to be a one-off, or is there any plan to pull new probe descriptions in the future? Would this potentially be a better fit for the schema generator where we already have access to scraped probe information?

scholtzan:
Ah, I wasn't aware of the schema generator having access to that information. In that case it's probably better to extend that one instead of having a script that needs to be run manually

templates/include/telemetry/environment.1.schema.json
jklukas: This seems fine to me. Indeed, there's no style constraints on templates and includes, though the generated schemas have consistent formatting enforced. It's pretty easy to see from the generated schemas here that this PR is correctly just adding descriptions.

acmiyaguchi: Seconded -- it's actually kind of nice to have the templates formatted too, but it doesn't need to be a sweeping change since the generated schemas are formatted.

scholtzan:
I added the script that I used for extracting the descriptions. I only tested it on [environment.json](https://github.com/mozilla/probe-dictionary/blob/master/environment.json)

PR #510, opened 2020-03-05
Bug 1612935 - Missing enhancedSearch fields in telemetry.core_v10 pings

acmiyaguchi:
[buglink](https://bugzilla.mozilla.org/show_bug.cgi?id=1612935)

https://sql.telemetry.mozilla.org/queries/68956/source
https://sql.telemetry.mozilla.org/queries/68955/source

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/telemetry/core/core.10.schema.json
+++ b/templates/telemetry/core/core.10.schema.json
@@ -8,16 +8,7 @@
       "enum": [ 10 ]
     },
     @TELEMETRY_CLIENTID_1_JSON@,
-    @TELEMETRY_CORECOMMON_2_JSON@,
-    "enhancedSearchUsed": {
-      "type": "boolean"
-    },
-    "enhancedSearchReady": {
-      "type": "boolean"
-    },
-    "enhancedSearchVersion": {
-      "type": "string"
-    }
+    @TELEMETRY_CORECOMMON_2_JSON@
   },
   "required": [ "v", "clientId", "seq", "arch", "device", "locale", "os", "osversion" ]
 }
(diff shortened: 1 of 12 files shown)


templates/telemetry/core/core.10.schema.json
```diff
@@ -8,7 +8,16 @@
       "enum": [ 10 ]
     },
     @TELEMETRY_CLIENTID_1_JSON@,
-    @TELEMETRY_CORECOMMON_2_JSON@
+    @TELEMETRY_CORECOMMON_2_JSON@,
```
Ben-Wu: What's the reason for putting these here rather than in coreCommon?  Is it because these are possibly temporary fields?

jklukas: I share Ben's concern here. I think these should be in coreCommon so they're applied to all versions of the core schema. The `telemetry.core` view unions v2 through v10 together and relies on them all being identical.

acmiyaguchi: I put them here because they seemed specific to v10 of the core ping. This would certainly break the union between v2 to v10 though.

Ben-Wu: Oh that's true, I didn't consider the union

jklukas: My understanding of the history here is that there never should have been multiple versions. The multiple versions only exist because of a misunderstanding that version has to be bumped for adding fields. We got to v10 and cleared that up, and made the decision to basically ignore the version and have all new fields appear in all versions of the table.

acmiyaguchi:
I've moved the `enhancedSearch` fields into the common core, and also added a description with the bug number.

PR #511, opened 2020-03-10
Add missing environement column and types

scholtzan:
Add missing column `environment.system.gfx.adapters.adapterDriver` to environment and add missing types to schemas.

Fixes part of https://bugzilla.mozilla.org/show_bug.cgi?id=1614416

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/include/telemetry/environment.1.schema.json
+++ b/templates/include/telemetry/environment.1.schema.json
@@ -820,6 +820,12 @@
                   },
                   "GPUActive": {
                     "type": "boolean"
+                  },
+                  "adapterDriver": {
+                    "type": [
+                      "string",
+                      "null"
+                    ]
                   }
                 }
               }
--- a/templates/include/telemetry/environment.1.schema.json
+++ b/templates/include/telemetry/environment.1.schema.json
@@ -821,7 +821,7 @@
                   "GPUActive": {
                     "type": "boolean"
                   },
-                  "adapterDriver": {
+                  "driverVendor": {
                     "type": [
                       "string",
                       "null"
--- a/templates/telemetry/new-profile/new-profile.4.schema.json
+++ b/templates/telemetry/new-profile/new-profile.4.schema.json
@@ -19,6 +19,7 @@
             ]
         },
         "processes": {
+          "type": "object",
           "$comment": "This use of 'allOf' is a terrible hack to force this section not to appear in BigQuery",
           "allOf": [{
             "type": "object",
--- a/schemas/telemetry/crash/crash.4.schema.json
+++ b/schemas/telemetry/crash/crash.4.schema.json
@@ -832,6 +832,12 @@
                           "null"
                         ]
                       },
+                      "adapterDriver": {
+                        "type": [
+                          "string",
+                          "null"
+                        ]
+                      },
                       "description": {
                         "type": [
                           "string",
--- a/schemas/telemetry/crash/crash.4.schema.json
+++ b/schemas/telemetry/crash/crash.4.schema.json
@@ -832,12 +832,6 @@
                           "null"
                         ]
                       },
-                      "adapterDriver": {
-                        "type": [
-                          "string",
-                          "null"
-                        ]
-                      },
                       "description": {
                         "type": [
                           "string",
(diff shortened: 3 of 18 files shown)


templates/telemetry/new-profile/new-profile.4.schema.json
```diff
@@ -19,6 +19,7 @@
             ]
         },
         "processes": {
+          "type": "object",
```
acmiyaguchi: What's the reason behind adding this type to the new-profile ping?

templates/include/telemetry/environment.1.schema.json
```diff
@@ -820,6 +820,12 @@
                   },
                   "GPUActive": {
                     "type": "boolean"
+                  },
+                  "adapterDriver": {
+                    "type": [
+                      "string",
+                      "null"
+                    ]
```
acmiyaguchi: Do you have an example query showing instances of `adapterDriver`? I tried the `main` and `event` ping (https://sql.telemetry.mozilla.org/queries/69068/source). Additionally, I don't see the field listed in the source docs: https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/environment.html#environment

scholtzan: This would be one example query: 
https://console.cloud.google.com/bigquery?sq=1018637472215:73ba37ca3a3445519443811af49da76d

Right, it doesn't seem to be documented anywhere. But it gets send when looking into the code: https://searchfox.org/mozilla-central/source/toolkit/components/telemetry/app/TelemetryEnvironment.jsm#509

scholtzan:
As part of https://github.com/mozilla/mozilla-schema-generator/pull/120 mozilla-schema-generator will process schemas that follow the common ping format. The schema generator checks that every field has a type, so when running it on the schemas it threw errors due to missing types.

templates/include/telemetry/environment.1.schema.json
acmiyaguchi: This may be a typo then, `adapterDriver` should be renamed into `driverVendor`, which is a documented field.

acmiyaguchi:
> As part of [mozilla/mozilla-schema-generator#120](https://github.com/mozilla/mozilla-schema-generator/pull/120) mozilla-schema-generator will process schemas that follow the common ping format. The schema generator checks that every field has a type, so when running it on the schemas it threw errors due to missing types.

Thanks for clarifying. This was an issue thrown by the jsonschema-transpiler presumably? Declaring the type outside of the "allOf" block seems like a fine way to get around the check.

templates/include/telemetry/environment.1.schema.json
scholtzan: Mh, it seems like both fields are getting used. But right, if `driverVendor` is set then `adapterDriver` is null and vice versa.

scholtzan:
It's also checked in schema generator (https://github.com/mozilla/mozilla-schema-generator/blob/30e470f55c652deb9bc108fa1a127ec9a91a37ea/mozilla_schema_generator/schema.py#L107)

templates/include/telemetry/environment.1.schema.json
acmiyaguchi: I'm a bit unclear on the usage of `adapterDr. In the cited `TelemetryEnvironment::getGfxAdapter` method, it looks like the function is returning a single adapter element. Internally, there is a property `adapterDriver` appended with a suffix. It seems like this is translated into the fields observed in in this section of the schema by removing the adapter prefix (e.g. `adapterDriver` seems to map to `environment.system.gfx.adapters.[].driver`). I [modified the query to search for `adapterDriver`](
https://console.cloud.google.com/bigquery?sq=1018637472215:73ba37ca3a3445519443811af49da76d)
in `additional_properties`, which returns no rows. Is there an `adapterDriver` field that you've found in one of the other pings? 

The query example you provided extracts `$.environment.system.gfx.adapters[0].driverVendor` from `additional_properties`, which does seems to be missing. 

It may be useful for review to update a validation document with a value observed from `additional_properties` for the missing field(s).

scholtzan: I don't see a `adapterDriver` field in any other ping. I would also think that `adapterDriver` maps to `environment.system.gfx.adapters.[].driver`.

I'll add this case to one of the validation documents 👍

scholtzan:
It seems I got confused with the naming. `driverVendor` is the field that was missing.

PR #516, opened 2020-03-23
Bug 1623626 - Add crash stacks to crash schema

wlach:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/telemetry/crash/crash.4.schema.json
+++ b/templates/telemetry/crash/crash.4.schema.json
@@ -162,6 +162,72 @@
         "processType": {
           "type": "string"
         },
+        "stackTraces": {
+          "type": ["object", "null"],
+          "properties": {
+            "crash_info": {
+              "type": "object",
+              "properties": {
+                "address": @TELEMETRY_HEXADDRESS_1_JSON@,
+                "crashing_thread": {
+                  "type": "integer",
+                  "minimum": 0
+                },
+                "type": {
+                  "type": ["string", "null"]
+                }
+              }
+            },
+            "main_module": {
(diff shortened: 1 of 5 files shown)


wlach:
Still a work in progress, per error rate above. Plz hold off on review for now

wlach:
Ok, ready for review now. In addition to the the integration tests, I also validated this works ok with all the crash stacks submitted with crash stacks on March 1st.

@jklukas -- could you validate that these schema additions should be interpreted ok by our bigquery machinery? it's rather larger and more complex than usual.

templates/telemetry/crash/crash.4.schema.json
```diff
@@ -162,6 +162,72 @@
         "processType": {
           "type": "string"
         },
+        "stackTraces": {
+          "type": ["object", "null"],
+          "properties": {
+            "crash_info": {
+              "type": "object",
+              "properties": {
+                "address": @TELEMETRY_HEXADDRESS_1_JSON@,
```
jklukas: Do we really want to throw out the ping if one of these hex fields isn't in the expected format? Should we just let these be strings without validation?

wlach: On the fence about this -- the fact that all the pings we're seeing in the field are formatting this correctly make me think we should be strict about this so we notice if people start submitting garbage. The main point of this telemetry is to allow us to create a signature of crashes that are being submitted, I'd like to be notified directly if that seems to be impossible.

wlach:
Good idea re: the test cases, I'll make sure to add some for the permutations I found.

PR #519, opened 2020-03-24
Bug 1623439 Replace shield_id with experiments for activity-stream etc

ncloudioj:
…system

Checklist for reviewer:

For glean changes:

r? @jklukas

### DIFF
--- a/schemas/activity-stream/events/events.1.schema.json
+++ b/schemas/activity-stream/events/events.1.schema.json
@@ -20,6 +20,18 @@
       "description": "An event identifier",
       "type": "string"
     },
+    "experiments": {
+      "additionalProperties": {
+        "properties": {
+          "branch": {
+            "type": "string"
+          }
+        },
+        "type": "object"
+      },
+      "description": "An object to record all active experiments, experiments IDs are stored as keys, and the value object stores the branch information. Example: {\"experiment_1\": {\"branch\": \"control\"}, \"experiment_2\": {\"branch\": \"treatment\"}}. This deprecates the \"shield_id\" used in activity-stream and messaging-system.",
+      "type": "object"
+    },
     "locale": {
       "type": "string"
     },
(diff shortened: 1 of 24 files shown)


ncloudioj:
This fixes [bug 1623439](https://bugzilla.mozilla.org/show_bug.cgi?id=1623439), which has been causing lots of pain while conducting experiments for Activity Stream and Messaging System.

We used to do string matching (`activity-stream`) to filter out the non-ActivityStream experiments in the telemetry, this has become super error-prone for experiment designers. Now that BigQuery allows us to easily store the whole experiment information, we'd like to lift this limitation.

Specifically, we are looking to deprecate the `shield_id` filed, and to add a new filed `experiments` of the type "REPEATED RECORD". It follows the same structure as the `experiments` defined in [telemetry.environments](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/include/telemetry/environment.1.schema.json#L316) without including the `type` and `enrollmentID` fields.

jklukas:
Note that it may be a few days before changes are deployed due to pipeline instability issues we're currently working through.

ncloudioj:
@jklukas Thanks!

> Note that it may be a few days before changes are deployed due to pipeline instability issues we're currently working through.

Copy that, will put the client-side patch on hold.

jklukas:
> Copy that, will put the client-side patch on hold.

If you release the client-side patch early, the new info will go to additional_properties. But it won't cause any operational trouble for the pipeline.

ncloudioj:
On a side note, if there is no experiment, shall we send an empty object, like `"experiments": {}`, or just omit this field in the payload?

jklukas:
> On a side note, if there is no experiment, shall we send an empty object, like `"experiments": {}`, or just omit this field in the payload?

Either is fine. Omitting the field from the payload will lead to the field being NULL in BQ. Note, though, that the document would fail to validate if you set `"experiments": null`; you would need to specify `"type": ["null", "object"]` to support that.

ncloudioj:
OK, will always send an object then.

Thanks for confirming this!

PR #520, opened 2020-03-25
Bug 1624986 - Fix specification of sessionId, crashTime fields in crash schema

wlach:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/telemetry/crash/crash.4.schema.json
+++ b/templates/telemetry/crash/crash.4.schema.json
@@ -19,6 +19,14 @@
           "type": "string",
           "description": "Optional, ID of the associated crash. Added to schema in bug 1604666."
         },
+        "crashTime": {
+          "description": "<ISO Date>, time of crash. Per-hour resolution",
+          "type": "string"
+        },
+        "sessionId": {
+          "description": "<uuid>, Telemetry ID of crashing session. May be missing for crashes that happen early in startup",
+          "type": ["string", "null"]
+        },
         "hasCrashEnvironment": {
           "type": "boolean"
         },
@@ -58,7 +66,7 @@
               "type": "string"
             },
             "CrashTime": {
-              "description": "<time>, Seconds since the Epoch",
+              "description": "<time>, Seconds since the Epoch (see also: `payload.crashTime`)",
               "type": "string"
             },
             "EventLoopNestingLevel": {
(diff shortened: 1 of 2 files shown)


wlach:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1624986 for context

templates/telemetry/crash/crash.4.schema.json
```diff
@@ -121,10 +125,6 @@
               "description": "<percentage>, Windows-only, percent of memory in use",
               "type": "string"
             },
-            "TelemetrySessionId": {
-              "description": "<id>, Active telemetry session ID when the crash was recorded",
-              "type": "string"
-            },
```
acmiyaguchi: This has already made it into `generated-schemas` so this field can't be removed. If this appears in additional properties but is camelCased instead, it would be fine to change it here, otherwise I would update the description to mention that this is an obsolete field as of bug 1624986.

templates/telemetry/crash/crash.4.schema.json
```diff
@@ -57,10 +65,6 @@
               "description": "1, Optional, if set indicates that the crash had a memory report attached",
               "type": "string"
             },
-            "CrashTime": {
```
acmiyaguchi: Fortunately `CrashTime` and `crashTime` map to the same field in the generated schema.

jklukas:
Good catch on removing a field, @acmiyaguchi

templates/telemetry/crash/crash.4.schema.json
wlach: I don't think the problem is that it's obsolete so much as it was never correct in the first place. But I have no problem leaving it in!

wlach:
These fields were already in the validation documents (sessionId as null):

https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/86956f9bf25a520c30da27b8a77b058f79c8a9f8/validation/telemetry/crash.4.sample.pass.json#L170

I could add another validation document for non-null sessionId, but I don't think it would add a lot of value.

templates/telemetry/crash/crash.4.schema.json
acmiyaguchi: Actually, this field is completely separate from the other crashTime field.

https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/86956f9bf25a520c30da27b8a77b058f79c8a9f8/validation/telemetry/crash.4.sample.pass.json#L173-L183

Looking at the validation document, this is `payload.metadata.crash_time`, where-as the field you are adding is `payload.crash_time`. This field is [also mentioned in the firefox source docs](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/crash-ping.html). This should also be left since this would be a schema incompatible change.

wlach: Oh right! Good catch! 💯

PR #530, opened 2020-04-24
Add minimal Account Ecosystem Telemetry document types

jklukas:
These are not yet useful, but they demonstrate the proposed structure
detailed in https://github.com/mozilla/gcp-ingestion/pull/1264

Please poke lots of holes in this, suggest better naming, etc.


templates/telemetry/account-ecosystem/account-ecosystem.4.schema.json
```diff
@@ -0,0 +1,78 @@
...
+    "type": {
+      "type": "string",
+      "enum": [ "account-ecosystem" ]
+    },
+    "version": {
+      "type": "number",
+      "minimum": 4,
+      "maximum": 4
+    },
+    "payload": {
```
jklukas, line 18: @chutten - Can you take a look at this `telemetry/account-ecosystem` docType?

I've refactored it to look more like existing telemetry pings; in particular:

- Made ecosystemUserId camelCase
- Moved ecosystemUserId underneath a `payload` structure
- Added placeholders for the various metric types: scalars, keyedHistograms, etc.
- Added the standard top-level application, creationDate, and id fields

This is based on a copy of the pre-account schema. Notably, I've removed `environment` assuming that might give too much identifying information about the user, but I may be wrong there.

Does this structure look about like what you'd expect, @chutten? Will it be convenient for implementation on the desktop client side? Anything in here look dangerous?

chutten: It's looking okay to me. In the original impl we included a subset of the Environment (see [docs](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/ecosystem-telemetry.html#environment)) that were specifically requested for analysis purposes and passed muster as being insufficiently fingerprinting for comparison with client_id-having pings. If no one's asking for them this time around, don't add them. Relevant fields could be added later as Scalars instead and have better description, expiry, and ownership semantics.

I think the name `duration` could be something that mentions its units and the span of time it's measuring (applicationSessionSeconds?), but maybe that'll be better covered in the in-tree documentation. Beware that durations don't compare between Windows and Mac because our timers treat computer sleep differently (see [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1204823)).

I'm a little worried that adding the fields that SHALL NOT BE POPULATED to this schema is asking for them to be populated in pings being sent by clients. Seems like a trap we're setting for our future selves somehow. I don't know what to do about that beyond the descriptions you already included and the code we're about to write to drop payloads with these forbidden fields... but it makes my shoulders itch.

This'll be no trouble on the client side... though your Scalars and Histograms make no mention of processes. What wil you do if the data to be included isn't in the parent process? Is that a limitation we're comfortable with? If not you might want to structure things a little like [deletion-request does](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/telemetry/deletion-request/deletion-request.4.schema.json#L13)

jklukas: > I think the name `duration` could be something that mentions its units and the span of time it's measuring (applicationSessionSeconds?), but maybe that'll be better covered in the in-tree documentation. Beware that durations don't compare between Windows and Mac because our timers treat computer sleep differently (see [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1204823)).

This was copied from the pre-account schema. I'll remove `duration` for now and we can add back in later if we need it and have thought more about it.

> I'm a little worried that adding the fields that SHALL NOT BE POPULATED to this schema is asking for them to be populated in pings being sent by clients. Seems like a trap we're setting for our future selves somehow. I don't know what to do about that beyond the descriptions you already included and the code we're about to write to drop payloads with these forbidden fields... but it makes my shoulders itch.

Fair enough. My hope is that having good descriptions in the schema that make this point clear will help a lot, and then we'll have to rely on metrics on the pipeline side to keep track of how many pings we throw out as invalid, with enough granularity in the metrics to understand why payloads are being thrown out and debug what clients are doing.

> This'll be no trouble on the client side... though your Scalars and Histograms make no mention of processes. What will you do if the data to be included isn't in the parent process? Is that a limitation we're comfortable with? 

At this point, I'm mostly worried about making sure we don't do anything that's going to cause us grief in the future. The deletion-request structure looks reasonable and flexible for future needs, so I'll plan to switch to that.

templates/telemetry/account-ecosystem/account-ecosystem.4.schema.json
```diff
@@ -0,0 +1,59 @@
...
+          "items": {
+            @ACCOUNT-ECOSYSTEM_USER_ID_TYPE_1_JSON@
+          }
+        },
+        "scalars": {
+          "type": "object",
+          "properties": {
+            "parent": @TELEMETRY_SCALARS_1_JSON@
+          }
+        }
```
jklukas, line 43: Including only scalars and only "parent" process for now as placeholders. We can add in other metric types and processes as needed.

schemas/telemetry/account-ecosystem/account-ecosystem.4.schema.json
```diff
@@ -0,0 +1,121 @@
...
+          "type": "string"
+        },
+        "ecosystemUserId": {
+          "description": "Account Ecosystem Telemetry user identifier; this value is not present in the original payload sent by clients, but is instead inserted by the pipeline after decrypting and removing ecosystemAnonId",
+          "pattern": "[a-zA-z0-9]{32}",
+          "type": "string"
+        },
+        "previousEcosystemUserIds": {
+          "description": "Previous Account Ecosystem Telemetry identifiers associated with this user; this value is not present in the original payload sent by clients, but is instead inserted by the pipeline after decrypting and removing previousEcosystemAnonIds",
+          "items": {},
```
jklukas, line 74: We should specify type as string for these items

PR #531, opened 2020-04-24
Allow running bigquery_schema_diff for same base and head refs

acmiyaguchi:
This fixes broken CI on master, which is caused by the `bigquery_schema_diff` being run for the same revision in the base and head references. This also adds an option to the script to specify a separate directory to run tests in via the `--root-directory` option.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/scripts/bigquery_schema_diff.py
+++ b/scripts/bigquery_schema_diff.py
@@ -426,7 +426,8 @@ def test_checkout_transpile_schemas(tmp_git: Path, tmp_path):
 
 
 def test_main(tmp_git):
-    main(["--root-directory", str(tmp_git)])
+    # choose a base ref relative to HEAD, since the head ref may be master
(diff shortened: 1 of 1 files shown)


scripts/bigquery_schema_diff.py
```diff
@@ -287,6 +298,8 @@ def tmp_git(tmp_path: Path) -> Path:
 
     run(f"git clone {origin} {workdir}")
     os.chdir(workdir)
+    # make branches available by checking them out, but ensure state ends up on HEAD
```
Ben-Wu, line 301: What happens if master isn't checked out?

acmiyaguchi: `git rev-parse` will run into `fatal: ambiguous argument 'master': unknown revision or path not in the working tree.` Running `git fetch --all` doesn't solve this issue for some reason.

PR #532, opened 2020-04-24
 Bug 1631849 - Remove pioneer-study pings and update pioneer-study envelope

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1631849)

This removes all of the pioneer-study pings under the respective namespaces. This also updates the telemetry/pioneer-study envelope schema to include a `schemaNamespace` for pioneer v2. This field is required, so it will start rejecting existing pioneer v1 pings into the error stream as a validation error.

Checklist for reviewer:

For glean changes:


acmiyaguchi:
Moving a schema sounds involved, with the condition that the table is deleted before the next deploy. I'm in favor of reusing this since it requires no incompatible state changes in BigQuery and little to no change in the client code for submitting to the correct route. 

As for the long tail, adding the new field will already allow us to reject the ping via a validation exception. Nothing stops us from marking this data as unwanted in the decoder since the study pings should get rerouted before they reach the telemetry decoder in the near future.

PR #541, opened 2020-05-05
Bug 1635212 - Handle frames property of threads in crash pings

wlach:
This also makes the schema slightly more strict to make sure that
we're actually seeing what we expect to see.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/crash/crash.4.schema.json
+++ b/schemas/telemetry/crash/crash.4.schema.json
@@ -1470,6 +1470,25 @@
             "threads": {
               "items": {
                 "properties": {
+                  "frames": {
+                    "items": {
+                      "properties": {
+                        "ip": {
+                          "pattern": "^(0x[a-fA-F0-9]+|0)$",
+                          "type": "string"
+                        },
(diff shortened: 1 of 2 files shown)


schemas/telemetry/crash/crash.4.schema.json
```diff
@@ -1470,18 +1470,33 @@
...
+                          "type": "integer"
+                        },
+                        "trust": {
+                          "type": "string"
+                        }
+                      },
+                      "required": [
+                        "ip",
+                        "trust"
+                      ],
```
wlach, line 1491: I'm not sure if this is the right thing-- it looks like it has no change on things as they stand (all crash pings with stack traces have these values, contained inside a frames object) but may artificially restrict us from modifying these pings in the future.

wlach: let's not do this after all

wlach:
I'm going to go ahead and land this, leaving (2) and (3) to followups.

PR #557, opened 2020-06-04
Document default browser agent's client id field as deprecated

bytesized:
[Bug 1621703](https://bugzilla.mozilla.org/show_bug.cgi?id=1621703#c10)

Originally, we planned to have a client id in the default browser agent's ping. However, we eventually decided not to include it. We never merged the code that added the client id, so no pings have ever been sent with this field.

### DIFF
--- a/schemas/default-browser-agent/default-browser/default-browser.1.schema.json
+++ b/schemas/default-browser-agent/default-browser/default-browser.1.schema.json
@@ -7,6 +7,11 @@
     "build_version": {
       "type": "string"
     },
+    "client_id": {
+      "description": "Deprecated; This field will never be sent",
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
+      "type": "string"
+    },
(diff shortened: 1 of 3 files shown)


bytesized:
I hope I did this right. I'm less experienced with git and I'm not sure why this PR ended up with 3 commits instead of 1.

jklukas:
We're currently lacking any documentation about when changes to schemas are breaking for BigQuery. Removing fields, though, is one of those case. BigQuery does not support dropping fields, so removing a field from the schema requires recreating the table from scratch, which is operationally dangerous.

I would recommend changing this PR to update the description of the field to indicate that it's deprecated. It will still exist in BigQuery (in the `default_browser_agent_stable.default_browser_agent_v1` table), but we can hide the field in the user-facing view (`default_browser_agent.default_browser_agent`).

It's also possible to bump the version on the schema, adding a new `v2` schema that removes the field. Clients would have to be updated to use the new version. We could alter the user-facing view to union the v1 and v2 tables.

bytesized:
I'll mark it as deprecated, then. I should have the PR updated by the end of the day.

jklukas:
See https://github.com/mozilla/bigquery-etl/pull/1036 for removing this field from the view

PR #562, opened 2020-06-22
Pin Centos image

jklukas:
We're seeing all test runs currently fail (see [example failure](https://app.circleci.com/pipelines/github/mozilla-services/mozilla-pipeline-schemas/1666/workflows/2af886d5-3764-4da0-9d51-bab9ff994611/jobs/3706/steps)).

This looks to be due to "diff" not being available in the centos8 image
published last week.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:centos8
+FROM centos:centos8.1.1911
 LABEL maintainer="Firefox Data Platform"
 
 # Install the appropriate software


jklukas:
> I would actually add diff to the list of dependencies, but I'm okay with pinning too. Strange that they would remove diff from the base install.

Feel free to issue a follow-up PR for that approach.

PR #564, opened 2020-06-23
Make sync migration comprehensible for bigquery.

### LINKED ISSUE #563: Bigquery schema for sync ping only contains "type" field and not the rest of the data
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/generated-schemas/schemas/telemetry/sync/sync.4.bq#L353-L362

Rest of the data is https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/telemetry/sync/sync.4.schema.json#L235-L258.

I guess it doesn't understand the `anyOf`? Hrm, I didn't really realize that this processing of the json-schema into bigquery schema was happening, or I'd have been more careful here in #560.

thomcc:
Fixes #563, hopefully.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/sync/sync.4.schema.json
+++ b/schemas/telemetry/sync/sync.4.schema.json
@@ -230,35 +230,26 @@
         },
         "migrations": {
           "items": {
-            "anyOf": [
-              {
-                "properties": {
-                  "entries": {
-                    "minimum": 0,
-                    "type": "integer"
-                  },
-                  "entriesSuccessful": {
-                    "minimum": 0,
-                    "type": "integer"
-                  },
-                  "extensions": {
-                    "minimum": 0,
-                    "type": "integer"
-                  },
-                  "extensionsSuccessful": {
-                    "minimum": 0,
-                    "type": "integer"
-                  },
-                  "openFailure": {
-                    "type": "boolean"
-                  },
-                  "type": {
-                    "const": "webext-storage"
-                  }
-                }
-              }
-            ],
             "properties": {
+              "entries": {
+                "minimum": 0,
+                "type": "integer"
+              },
+              "entriesSuccessful": {
+                "minimum": 0,
+                "type": "integer"
+              },
+              "extensions": {
+                "minimum": 0,
+                "type": "integer"
+              },
+              "extensionsSuccessful": {
+                "minimum": 0,
+                "type": "integer"
+              },
+              "openFailure": {
+                "type": "boolean"
+              },
               "type": {
                 "type": "string"
               }
--- a/schemas/telemetry/sync/sync.5.schema.json
+++ b/schemas/telemetry/sync/sync.5.schema.json
@@ -244,35 +244,26 @@
         },
         "migrations": {
           "items": {
-            "anyOf": [
-              {
-                "properties": {
-                  "entries": {
-                    "minimum": 0,
-                    "type": "integer"
-                  },
-                  "entriesSuccessful": {
-                    "minimum": 0,
-                    "type": "integer"
-                  },
-                  "extensions": {
-                    "minimum": 0,
-                    "type": "integer"
-                  },
-                  "extensionsSuccessful": {
-                    "minimum": 0,
-                    "type": "integer"
-                  },
-                  "openFailure": {
-                    "type": "boolean"
-                  },
-                  "type": {
-                    "const": "webext-storage"
-                  }
-                }
-              }
-            ],
             "properties": {
+              "entries": {
+                "minimum": 0,
+                "type": "integer"
+              },
+              "entriesSuccessful": {
+                "minimum": 0,
+                "type": "integer"
+              },
+              "extensions": {
+                "minimum": 0,
+                "type": "integer"
+              },
+              "extensionsSuccessful": {
+                "minimum": 0,
+                "type": "integer"
+              },
+              "openFailure": {
+                "type": "boolean"
+              },
               "type": {
                 "type": "string"
               }
--- a/templates/include/telemetry/syncMigration.1.schema.json
+++ b/templates/include/telemetry/syncMigration.1.schema.json
@@ -2,18 +2,11 @@
   "type": "object",
   "required": ["type"],
   "properties": {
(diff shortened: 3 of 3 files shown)


thomcc:
> This will work, however I'd like to clarify what was the intent to use `anyOf` before and if that's still the case?
> Did you want to require at least one of these fields?

The idea was: if the `type` fields is `"webext-storage"`, then these fields exist. When we add other migration types, it might have other fields. This is discussed a bit here: https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/sync-ping.html#the-migrations-array.

It's possible that this should be revisited given the nature of how bigquery schemas work, since I think it won't work well for that.

In the meantime, there's some time sensitivity to getting this data, as once the migration code hits release channel, we have no ability to fix it (barring something truly catastrophic, we're not going to ship a second migration for this code), so we'd rather not spend the time sorting out this logical inconsistency prior to that.

akkomar:
OK, I'm going to merge this since generated BQ schema matches what's described in the docs.

> It's possible that this should be revisited
I'd suggest to add some validation examples if we revisit this.

@acmiyaguchi FYI: it seems like nested schemas combined via `anyOf` aren't transpiled into BQ schemas - is that by design?

thomcc:
> I'd suggest to add some validation examples if we revisit this.

Ah, I meant the design of our migration telemetry should be revisited to not be so idiosyncratic on a per-type basis. (That said, it seems tricky to avoid, given that the data we want to report is based on the differences between the old and new schemas... Oh well)

But yes, I agree that if we do anything tricky we should have more validation examples (we have a few in there now, but they aren't effected by this, since both this and the old way the schema was formulated are basically equivalent as far as json-schema is concerned, just not equivalent after bigquery translation).

acmiyaguchi:
> OK, I'm going to merge this since generated BQ schema matches what's described in the docs.
> 
> > It's possible that this should be revisited
> > I'd suggest to add some validation examples if we revisit this.
> 
> @acmiyaguchi FYI: it seems like nested schemas combined via `anyOf` aren't transpiled into BQ schemas - is that by design?

There are a few keywords that aren't supported, but should be. With anyOf, it's possible to generate an object in bigquery that would be incompatible e.g. `{"anyOf": [{"type": "string"}, {"type": "integer"}]}`. It's something worth revisiting.

thomcc:
Not sure who to ask about this (I asked in #telemetry on c.m.o without any luck), but when will this be in a release / usable on redash?

Is my best option in the meantime trying to analyze with databricks?

> There are a few keywords that aren't supported, but should be

An error or warning here would have been really helpful, FWIW.

acmiyaguchi:
@thomcc Our deploys typically start around midnight UTC, finishing approximately at 2 am UTC. Once this is done, you should have access to the new schemas in Redash/BigQuery.

thomcc:
And old data will be available? Or will this only apply to new data going forward.

acmiyaguchi:
We currently don't have a mechanism to rewrite history in tables to accommodate schema evolution. The schema changes will only apply to new data going forward. Data in these new columns can be found in `additional_properties` for existing data, so you can extract the information out using [`JSON_EXTRACT` functions](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions) if you have a need.

thomcc:
> Data in these new columns can be found in additional_properties for existing data, so you can extract the information out using JSON_EXTRACT functions if you have a need.

The nature of this migration data (it's submitted once ever, after migration) means that it's somewhat unlikely that we'll get this information for the nightly population again, now that stuff has landed and many users have had migration run. (Again, this is to attempt to detect errors before it gets to release, ideally before it gets far in beta as well).

Given that sync doesn't have additional_properties in this way: https://github.com/mozilla-services/mozilla-pipeline-schemas/search?q=additional_properties&unscoped_q=additional_properties, and I can't find any additional_properties, err, property in redash... any chance you can elaborate? Or would we need to have added an `additional_properties` declaration to migrations there for this to work.

(If so, that's fine, it would be nice to do this analysis in redash, but using databricks seems like a potential fallback, maybe).

acmiyaguchi:
Additional properties is a field that is added by our ingestion system. The schema is used to traverse the document tree and place it into a BigQuery column. If an element in the document doesn't exist in the schema, it's put into a column named `additional_properties`. So for example, if you were to use our structured ingestion endpoint for data ingestion and you made an empty schema, all of the columns would go into `additional_properties`. This section on [schemas during decoding](https://docs.telemetry.mozilla.org/concepts/pipeline/schemas.html?highlight=additional,properties#decoding) and [the table layout](https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html?highlight=additional,properties#structure-of-ping-tables-in-bigquery) might be useful references.

To be more concrete, here is the data in the "live" table. It takes a full day before the data is accessible in "telemetry.sync" since those tables only contain full days. This shows data since the last deploy, where the changes that you've made have landed.

```sql
select
    payload.migrations
from `moz-fx-data-shared-prod`.telemetry_live.sync_v4
where submission_timestamp > timestamp_sub(current_timestamp, interval 1 hour)
and array_length(payload.migrations) > 0
limit 10
```

https://sql.telemetry.mozilla.org/queries/72220/source

This query shows how you would access historical data. The column can be found in the `additional_properties` column. 

```sql
select
    json_extract(additional_properties, "$.payload.migrations") as migrations
from `moz-fx-data-shared-prod`.telemetry.sync
where date(submission_timestamp) = date "2020-06-22"
and json_extract(additional_properties, "$.payload.migrations") is not null
limit 10
```
https://sql.telemetry.mozilla.org/queries/72221/source

You can merge these two approaches if you need to go across the date boundaries. Be aware of the query size becoming large since additional properties is a catch-all.


```
-- NOTE: you would need to define a udf to convert a json array into a bigquery array of structs
select
  coalesce(
    extract_json_migrations(json_extract(additional_properties, "$.payload.migrations")),
    payload.migrations
  ) as migrations
from telemetry.sync
```

thomcc:
Ah! Thanks so much!

PR #565, opened 2020-06-24
Bug 1624908 Add sendFailure subfields for health ping

jklukas:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/health/health.4.schema.json
+++ b/schemas/telemetry/health/health.4.schema.json
@@ -109,10 +109,19 @@
           "type": "string"
         },
         "sendFailure": {
-          "minProperties": 1,
           "properties": {
             "additionalProperties": {
+              "description": "Never populated; this field was added unintentionally and is retained only for compatibility",
               "type": "number"
+            },
+            "eTerminated": {
+              "type": "integer"
+            },
+            "eUnreachable": {
+              "type": "integer"
+            },
+            "timeout": {
+              "type": "integer"
             }
           },
           "type": "object"
--- a/schemas/telemetry/health/health.4.schema.json
+++ b/schemas/telemetry/health/health.4.schema.json
@@ -91,11 +91,25 @@
           "type": "object"
         },
         "pingDiscardedForSize": {
-          "maxProperties": 10,
-          "minProperties": 1,
           "properties": {
+            "<unknown>": {
+              "type": "integer"
+            },
             "additionalProperties": {
+              "description": "Never populated; this field was added unintentionally and is retained only for compatibility",
               "type": "number"
+            },
+            "crash": {
+              "type": "integer"
+            },
+            "main": {
+              "type": "integer"
+            },
+            "prio": {
+              "type": "integer"
+            },
+            "sync": {
+              "type": "integer"
             }
           },
           "type": "object"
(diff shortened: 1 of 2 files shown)


schemas/telemetry/health/health.4.schema.json
```diff
@@ -91,11 +91,25 @@
           "type": "object"
         },
         "pingDiscardedForSize": {
-          "maxProperties": 10,
-          "minProperties": 1,
           "properties": {
+            "<unknown>": {
```
jklukas, line 95: @acmiyaguchi From the bq schema diff, it looks like this field with name "<unknown>" just gets dropped and is absent from the bq schema. Does that sound like intended behavior?

I realize that generating a sane bq-compatible field name for "<unknown>" is a bit of a stretch, but I was expecting that this might be normalized to "_unknown_".

schemas/telemetry/health/health.4.schema.json
```diff
@@ -109,10 +123,28 @@
           "type": "string"
         },
         "sendFailure": {
-          "minProperties": 1,
           "properties": {
+            "abort": {
```
jklukas, line 127: We include the list of observed keys as seen in query: https://sql.telemetry.mozilla.org/queries/72242/source

schemas/telemetry/health/health.4.schema.json
jklukas, line 95: We include the observed keys as seen in query: https://sql.telemetry.mozilla.org/queries/72243/source

acmiyaguchi: No, there is currently no behavior for normalizing an empty string as a key into BigQuery. This is certainly something that may be added, perhaps mapping it to `_` or `__unknown__`. However, this key would conflict with other JSON keys like `.`, `_`, or `__`. I think the right behavior would be to panic or drop the fields in the case where the transpiler detects these name clashes. 

Would this be handled deterministically in the decoder?

```javascript
>> x = {}; x[""]=1; x["."]=2; x["_"]=3;
3
>> x
Object { "": 1, ".": 2, _: 3 }
```

acmiyaguchi: I'm actually not sure why "<unknown>" is converted into "", that's unintended behavior for sure.

jklukas: Redash was converting "<unknown>" to "". The BQ console shows the value as "<unknown>". So I would expect the schema generator to be seeing "<unknown>" as the input name and potentially normalizing to "_unknown_".

jklukas: Yeesh, having this discussion is difficult due to literally typing `<unknown>` getting interpreted as an html tag and hidden.

acmiyaguchi: I was able to reproduce this locally in a shell, outside of any rendering done by javascript/browser. 

```
echo '{"properties": {"payload": {"properties": {"<unknown>": {"type": "string"}, "foo": {"type": "string"}}}}}' | jq                               
{
  "properties": {
    "payload": {
      "properties": {
        "<unknown>": {
          "type": "string"
        },
        "foo": {
          "type": "string"
        }
      }
    }
  }

python % echo '{"properties": {"payload": {"properties": {"<unknown>": {"type": "string"}, "foo": {"type": "string"}}}}}' | jsonschema-transpiler -t bigquery               
[
  {
    "fields": [
      {
        "mode": "NULLABLE",
        "name": "foo",
        "type": "STRING"
      }
    ],
    "mode": "NULLABLE",
    "name": "payload",
    "type": "RECORD"
  }
]
```

jklukas:
Closing in favor of https://github.com/mozilla/bigquery-etl/pull/1173

PR #572, opened 2020-07-14
AET update: 64-char user_ids and relaxing validation

### LINKED ISSUE #571: Please make `ecosystem_user_id` be 64 hex chars, rather than 32
This field is currently defined as `"pattern": "[a-zA-z0-9]{32}"`. We intend this value to be derived as an FxA [scoped key](https://docs.google.com/document/d/1IvQJFEBFz0PnL4uVlIvt8fBS_IPwSK-avK0BRIHucxQ/) which by default would make it 32 *bytes* rather than 32 hex chars. If there are no concerns about inflating the size of this identifier, could be please make it be 64 hex chars == 32 bytes in order to simplify handling on the FxA side of things?

jklukas:
Fixes #571

Also allows `firefox-accounts` pings without an `ecosystem_client_id` value as
long as `ecosystem_anon_id` is present. Server apps will need to send AET
pings with anon_id but without any relevant client_id to be associated with,
so we want to avoid having to introduce a static server-side client_id just to
conform to the schema.

Telemetry AET pings are still required to have an `ecosystem_client_id` as we
expect the client should always be able to provide that value.

Checklist for reviewer:

For glean changes:


jklukas:
cc @rfk @6a68

PR #579, opened 2020-07-16
Bug 1648208 Support pipeline metadata fields

jklukas:
This is a draft to show a concrete implementation of pipeline metadata fields as a companion to the [pipeline metadata proposal](https://docs.google.com/document/d/1oNi-eX_IPIZa8C0h5GZnWTM0WFivZ1hPMgMMX3c3HHs/edit#).

~~This PR may evolve into a final implementation or may be abandoned as a proof of concept.~~ This PR is evolving into a full implementation.

Checklist for reviewer:

For glean changes:


CMakeLists.txt
```diff
@@ -68,7 +73,15 @@ foreach(ITEM IN ITEMS ${SCHEMAS})
     # NOTE the artifacts are written back into the source repository (see the README)
     set(IN_FILE "${CMAKE_SOURCE_DIR}/templates/${ITEM}")
     set(OUT_FILE "${CMAKE_SOURCE_DIR}/schemas/${ITEM}")
-    if(ITEM MATCHES "\\.json$") # pretty print
+    if(ITEM MATCHES "account-ecosystem...schema.json$") # TODO: Make this the default handling for all JSON schemas
```
jklukas, line 76: For demonstration purposes, we are limiting the new python-based metadata processing to `account-ecosystem`-related schemas. This treatment would be applied to all schemas in a final implementation.

schemas/firefox-accounts/account-ecosystem/account-ecosystem.1.schema.json
```diff
@@ -12,6 +12,16 @@
       ]
     }
   ],
+  "mozPipelineMetadata": {
+    "bq_dataset": "firefox_accounts",
+    "bq_table": "account_ecosystem_v1",
```
jklukas, line 17: This template schema doesn't define these bq fields; they are default values injected by `inject_metadata.py` based on the directory structure.

schemas/telemetry/account-ecosystem/account-ecosystem.4.schema.json
```diff
@@ -1,5 +1,9 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
+  "mozPipelineMetadata": {
+    "bq_dataset": "account_ecosystem",
+    "bq_table": "desktop"
```
jklukas, line 5: In this case, the bq fields _were_ defined in the template, so those custom values appear here rather than the defaults (which would have been "telemetry" and "account_ecosystem_v4").

CMakeLists.txt
```diff
@@ -19,6 +19,11 @@ if(NOT JQ_EXE)
   message(FATAL_ERROR "jq is required for output formatting")
 endif()
 
+find_program(PYTHON_EXE python3)
+if(NOT PYTHON_EXE)
```
whd, line 23: There is probably a way to existing cmake templating  logic for metadata but I'm fine with however we want to manage this.

inject_metadata.py
```diff
@@ -0,0 +1,32 @@
...
+    # Prepare some default metadata if not overridden in the template schema.
+    bq_dataset = fileparts[0].replace('-', '_')
+    doctype = fileparts[1].replace('-', '_')
+    version = fileparts[2].split('.')[1]
+    bq_table = '{}_v{}'.format(doctype, version)
+
+    # Parse the schema and inject default metadata fields.
+    schema = json.load(sys.stdin)
+    meta = schema.setdefault('mozPipelineMetadata', {})
+    meta.setdefault('bq_dataset', bq_dataset)
```
whd, line 25: The concept is sound, but I would be careful about naming here, since `bq_dataset` isn't really accurate (there are multiple datasets associated with every `document_namespace`, and each ping is expected to minimally go in `_live` and `payload_bytes_decoded`). I think I'd prefer `document_namespace` or `bq_document_namespace`, which has the added benefit that normalization to bq tables can be done in downstream logic as per SOP.

schemas/telemetry/account-ecosystem/account-ecosystem.4.schema.json
```diff
@@ -1,5 +1,15 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
+  "mozPipelineMetadata": {
+    "bq_dataset": "account_ecosystem",
+    "bq_table": "desktop",
```
whd, line 5: Is the intent here to map telemetry account ecosystem pings to a non-telemetry dataset/table? Certainly that is a capability we're trying to support via custom metadata, but in this case we would probably have MSG write the .bq versions of these to 

`schemas/account-ecosystem/desktop/desktop.4.bq`

in `generated-schemas`, and downstream logic would remain the same. (It's my understanding that aet pings aren't sensitive once the identifier is decrypted so this is perhaps just for demonstration purposes).

schemas/telemetry/account-ecosystem/account-ecosystem.4.schema.json
```diff
@@ -1,5 +1,15 @@
...
+    "jwe_mappings": [
+      {
+        "decrypted_field_path": "payload.ecosystemUserId",
+        "source_field_path": "payload.ecosystemAnonId"
+      }
+    ]
```
whd, line 11: Just as a note, some additional metadata we're planning on leveraging are the following:

* retention rules, both partition retention (standard case) and fixed date retention (special case for pioneer), the former of which we could apply to the activity-stream namespace
* data categorization, which depending on how this evolves may not be used directly but I still think is worth having in metadata

I have a vague notion that including pipeline family eventually in metadata would help me automate some of the ops provisioning logic, which likely would only matter in the case we start to branch out into multiple bq projects or separate pipeline families for high-category data.

schemas/telemetry/account-ecosystem/account-ecosystem.4.schema.json
jklukas: > Is the intent here to map telemetry account ecosystem pings to a non-telemetry dataset/table? 

Yes, that was what I was intending to demonstrate here.

> Certainly that is a capability we're trying to support via custom metadata, but in this case we would probably have MSG write the .bq versions of these to
> 
> `schemas/account-ecosystem/desktop/desktop.4.bq`
> 
> in `generated-schemas`, and downstream logic would remain the same. (It's my understanding that aet pings aren't sensitive once the identifier is decrypted so this is perhaps just for demonstration purposes).

Yes, it's my intention that MSG would read this metadata and write out the BQ schema to a path determined by these fields.

inject_metadata.py
jklukas: I've updated the proposal to use `bq_dataset_family`.

> normalization to bq tables can be done in downstream logic as per SOP

I rather like the idea of defining the properly normalized bq names at this schema metadata level, which would relive the pipeline of the responsibility of normalizing the names. It certainly makes the connection between doctype and BQ table more explicit.

We do need to be careful not to spread responsibility over too many places for the same thing.

jklukas:
This is still in draft state, but I've now added all the features I intend to add. I'm asking @whd for review at this stage. Once we're settled on the content here, I'll make the final changes to `CMakeLists.txt` so that all doctypes will be processed to add metadata, and I'll rebase this PR so that we have one commit for the machinery and a separate commit for committing all the actual changes under `schemas/`.

CMakeLists.txt
```diff
@@ -84,8 +97,6 @@ foreach(ITEM IN ITEMS ${SCHEMAS})
         message(FATAL_ERROR "parquetfmt output error ${PFMT_RV} ${ITEM}")
       endif()
       file(REMOVE ${TMP_FILE}) # the tmp file is intentionally not removed on failure so it can be examined
-    elseif(ITEM MATCHES "/README\\.")
```
jklukas, line 87: Moved this clause into the outer `if (NOT ITEM MATCHES` conditional

schemas/metadata/metaschema/metaschema.1.schema.json
```diff
@@ -0,0 +1,62 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Metaschema for validating the structure of the per-doctype schemas in this repository",
+  "id": "http://jsonschema.net",
+  "properties": {
+    "mozPipelineMetadata": {
+      "additionalProperties": false,
+      "description": "Container for per-doctype metadata that can affect how pings are processed in the pipeline",
+      "properties": {
+        "bq_dataset_family": {
+          "description": "NOT YET IMPLEMENTED: The base name for the destination BigQuery dataset; if this is set to 'telemetry', the pipeline will write into 'telemetry_live'",
```
jklukas, line 11: All these fields will be informational only to begin with. The idea is that as we actually change the pipeline and ops logic to rely on these fields, we will remove these `NOT YET IMPLEMENTED` warnings.

templates/activity-stream/defaults.schema.json
```diff
@@ -0,0 +1,7 @@
+{
+  "mozPipelineMetadata": {
+    "expiration_policy": {
+      "delete_after_days": 180
+    }
+  }
```
jklukas, line 6: Namespace-level defaults were listed as a future extension in the proposal, but as I was implementing this, it really seems necessary sooner rather than later, so I've included here support for files called `defaults.schema.json` within each namespace directory. Like `README`s, these files are not copied into `schemas/`.

schemas/metadata/metaschema/metaschema.1.schema.json
```diff
@@ -0,0 +1,62 @@
...
+              },
+              "source_field_path": {
+                "type": "string"
+              }
+            },
+            "type": "object"
+          },
+          "type": "array"
+        },
+        "uri_scheme": {
```
jklukas, line 49: `uri_scheme` was not included in the proposal, but would in practice have been useful for `xfocsp-error-report` where we are using a telemetry URI scheme in a BQ dataset family other than `telemetry`.

schemas/activity-stream/impression-stats/impression-stats.1.schema.json
```diff
@@ -1,5 +1,13 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
+  "mozPipelineMetadata": {
+    "bq_dataset_family": "activity_stream",
```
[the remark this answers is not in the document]
jklukas: I agree with you that the ideal state here is the m-p-s and m-s-g are fully responsible for normalization, and that the ops logic and pipeline logic no longer have to do any normalization. It's worth spending some time figuring out if that can be the case.

From the pipeline perspective, the decoder only cares about JSON schemas and the BQ sinks only care about BQ schemas. These could be two separate artifacts where the directory structure for JSON schemas is `namespace/doctype` and the directory structure for BQ schemas is `bq_dataset_family/bq_table`.

The ops logic, however, needs to know about both JSON schemas and BQ schemas, and the relationship between them. Currently, we handle that by keeping JSON and BQ schemas in the same directory with the `s/schema.json/bq/` relationship in naming.

What if we take a middle ground between the above two approaches. We keep the current structure of the `generated-schemas` branch of m-p-s, but we add a top-level `bq` directory. During a transition period, we duplicate each `bq` schema, writing it into `schemas`, but also writing it into a path under `bq` that is determined by the `bq_*` metadata fields. So we have `schemas/telemetry/xfocsp-error-report/xfocsp-error-report.1.bq` and also `bq/xfocsp_error_report/xfocsp_error_report_v1.bq`.

We could transition tooling piece-by-piece to use the BQ schemas under `bq` and then eventually drop BQ schemas from from `schemas` or convert them to symlinks purely as an affordance for humans to understand the routing. We could also provide symlinks under `bq` to the relevant JSON schemas for human readability.

This assumes a 1:1 mapping between doctype and table. `xfocsp-error-report` currently breaks that, but I believe we can fix that up and enforce it as a requirement going forward.

> While gcp labels are constrained to under 64 characters, we could probably come up with labeling schemes that would make it easier to work backward from a BQ table to files in generated-schemas

I think it would be great to apply labels `document_namespace` (with value like `activity-stream`), `document_type` (with value like `impression-stats`), and `document_version` (with value like `4`) which could be used to construct the json schema path in m-p-s. We could also inject the entire URL into the table's `description` similar to how we inject m-p-s URLs into dataset descriptions.

schemas/activity-stream/impression-stats/impression-stats.1.schema.json
```diff
@@ -1,5 +1,13 @@
...
+    "bq_table": "impression_stats_v1",
+    "expiration_policy": {
+      "delete_after_days": 180
+    },
+    "uri_scheme": "structured"
```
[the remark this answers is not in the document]
jklukas: > pipeline_families would need to be a list

Why a list? Is multiple values just for the rerouting case? If we're maintaining a 1:1 relationship between schemas and BQ tables, then I'd like to see if we can resolve that ambiguity. `xfocsp-error-report` would be processed completely by the `telemetry` pipeline family, correct?

templates/activity-stream/defaults.schema.json
jklukas: If I'm reading this correctly, then I'm in agreement that we can defer proper namespace-level metadata to a future improvement.

schemas/activity-stream/impression-stats/impression-stats.1.schema.json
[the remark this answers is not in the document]
whd: Would that it were so simple. The reason for this is the firefox installer namespace https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/master/schemas/firefox-installer is populated by two pipeline families: `telemetry` (for the full installer ping) and `stub_installer` (DSMO / stub installer ping).

It's perhaps reasonable to revisit whether stub installer warrants its own pipeline family. Semantically I think it does, but doing so increases the complexity of the pipeline considerably and the ingestion decoder infrastructure doesn't rely on the pipeline family distinction. Operationally we would lose visibility into DSMO by merging it into telemetry, but it's such a small amount of data that in practice it probably doesn't matter. As another issue of semantics technically `uri_scheme` for this ping is a list in that source data URI schemes are the same as the pipeline families.

We don't need to block this PR on deciding what to do about pipeline families.

schemas/activity-stream/impression-stats/impression-stats.1.schema.json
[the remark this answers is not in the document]
whd: Ok, so it sounds like you're in favor of "change MPS filename conventions (breaking change)", but deploying it in a piecewise (non-breaking) fashion. I think the approach you outline is reasonable, but I would like some opinions about this structure from other people involved with MSG infrastructure such as @acmiyaguchi and @fbertsch. Generally, I'm in favor of keeping a single artifact (with two directory prefixes e.g. `bq` and `schemas`) i.e. no ops changes to the way `generated-schemas` is turned into a single versioned tarball for ops and pipeline consumers to use (the sink and decoder use the same artifact but read different paths).

From the ops pipeline perspective, only having to know about json schemas and using metadata to find the associated bq schema file path is pretty close to the ideal state I'd like to get to. The ops pipeline may still care about non-normalized values derived from json schemas directory structure for descriptions and labels, but I think it might be useful to eventually move away from any dependence on directory structure in favor of metadata. However, I also don't like the idea of un-normalizing or having both normalized and non-normalized values in metadata, but between these two options, I would prefer the latter.

> This assumes a 1:1 mapping between doctype and table. xfocsp-error-report currently breaks that, but I believe we can fix that up and enforce it as a requirement going forward.

I agree this is not a pressing concern and that it should be true in the future.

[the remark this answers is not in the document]
acmiyaguchi: I'm not against having a separate directory with the bigquery schemas, assuming a single tarball with the artifacts. 

> We could transition tooling piece-by-piece to use the BQ schemas under bq and then eventually drop BQ schemas from from schemas or convert them to symlinks purely as an affordance for humans to understand the routing. We could also provide symlinks under bq to the relevant JSON schemas for human readability.

How would we map backwards from the normalized name to the original naming? There would need to be an extra step as the schemas are being generated?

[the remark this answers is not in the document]
jklukas: > How would we map backwards from the normalized name to the original naming? There would need to be an extra step as the schemas are being generated?

If we want to provide this mapping, we would do so as the schemas are being generated. As we iterate through the JSON schemas and write out BQ schemas under `bq/`, we could add a symlink or a text file that contains the path to the relevant JSON schema (with non-normalized naming).

schemas/metadata/metaschema/metaschema.1.schema.json
acmiyaguchi: It's worth updating the proposal document to reflect the implementation here.

tests/test_schema_structure.py
```diff
@@ -0,0 +1,11 @@
...
+from jsonschema.exceptions import ValidationError
+
+
+def test_schema_structure(schemas):
+    metaschema = schemas.pop("metadata.metaschema.1")
+    for key, schema in schemas.items():
+        try:
+            validate(schema, schema=metaschema)
+        except ValidationError as e:
+            raise ValidationError("Schema {} failed validation".format(key)) from e
```
acmiyaguchi, line 11: It'd be useful to have a test for the defaults here, like checking whether the telemetry ping has the metadata field. This also implies that the tests require the schemas are generated (which we do in CI, but should eventually move into the test suites).

schemas/activity-stream/impression-stats/impression-stats.1.schema.json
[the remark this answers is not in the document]
whd: > How would we map backwards from the normalized name to the original naming? There would need to be an extra step as the schemas are being generated?

Yes the intent here is to avoid having to unnormalize at all, and what klukas describes is how we'd achieve that. We only want to normalize once.

tests/test_schema_structure.py
jklukas: > This also implies that the tests require the schemas are generated (which we do in CI, but should eventually move into the test suites).

We already make that assumption. `schemas` here is referencing files under `schemas`.

I will add a test for defaults being applied.

tests/test_metadata_defaults.py
```diff
@@ -0,0 +1,3 @@
+def test_metadata_defaults(schemas):
```
jklukas, line 1: Added test for application of namespace-level defaults.

schemas/metadata/metaschema/metaschema.1.schema.json
```diff
@@ -0,0 +1,62 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
fbertsch, line 2: I'm not sure that this schema belongs in `schemas/`, since it's defining what is valid in _other_ schemas in `schemas/`. This might belong in a separate `validation/` directory.

scripts/inject-metadata
```diff
@@ -0,0 +1,68 @@
...
+    schema_basename = fileparts[-1]
+    doctype = fileparts[-2]
+    namespace = fileparts[-3]
+
+    # Parse the schema and extract metadata.
+    schema = json.load(sys.stdin)
+    meta = schema.get(MOZ_PIPELINE_METADATA, {})
+
+    # Namespace-level defaults.
+    defaults_path = TEMPLATES_DIR / namespace / "defaults.schema.json"
```
fbertsch, line 45: I'm not sure that namespace-level defaults belong in `templates` either. Since these shouldn't need the capabilities of generating the schemas, we should move them to a separate directory.

schemas/activity-stream/impression-stats/impression-stats.1.schema.json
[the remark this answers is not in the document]
fbertsch: The above approach seems best to me. Approaching this piece-meal will make the implementation much easier 👍

jklukas:
@whd Do you want another chance to look this over? I'd like to go ahead and merge this today, and we can file changes as we start building use cases on top of the metadata.

whd:
> @whd Do you want another chance to look this over? I'd like to go ahead and merge this today, and we can file changes as we start building use cases on top of the metadata.

yep looking it over now

schemas/firefox-installer/install/install.1.schema.json
```diff
@@ -1,5 +1,10 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
+  "mozPipelineMetadata": {
+    "bq_dataset_family": "firefox_installer",
+    "bq_table": "install_v1",
+    "uri_scheme": "structured"
```
whd, line 6: I said before I didn't want to block on figuring out pipeline family vs. uri_scheme, but regardless of that this annotation is in fact ambiguous WRT firefox installer pings i.e. if we were to parse a DSMO stub installer ping based on this metadata it would fail. To correct my earlier statements, this namespace probably uses the `structured` uri scheme for full installer and `stub_installer` for DSMO pings. These happen to map to pipeline families, but uri scheme to pipeline family is not always 1:1 e.g. pioneer uses the `telemetry` uri scheme.

For this first iteration, as long as uri scheme is only used by MSG for BQ metadata generation (n.b. its documentation indicates this explicitly) this characterization is fine (side note, pioneer uses custom metadata even though its uri scheme is `telemetry`; it seems things are not quite 1:1 all over the place). We're going to want to define all of these carefully in further iterations and additions to metadata.

schemas/firefox-accounts/account-ecosystem/account-ecosystem.1.schema.json
```diff
@@ -12,6 +12,21 @@
...
+      {
+        "decrypted_field_path": "ecosystem_user_id",
+        "source_field_path": "ecosystem_anon_id"
+      },
+      {
+        "decrypted_field_path": "previous_ecosystem_user_ids",
+        "source_field_path": "previous_ecosystem_anon_ids"
+      }
+    ],
+    "uri_scheme": "structured"
```
whd, line 28: I'm a bit hung up on uri_scheme since semantically its a property of the input but this metadata as spec'd will only be used on output. For structured aet in particular I'd expect a new input uri scheme of `stackdriver`.

jklukas:
> Aside from prepping the exclusion in MSG for metadata before merging this, LGTM. I think we may need to rework or rename uri_scheme but I don't think that's worth blocking this PR on.

Thanks for the careful review. It'll likely be Wednesday before I have a chance to get the MSG change prepped, so I will delay merging this until then.

jklukas:
I've pushed a new commit that changes `uri_scheme` to `bq_metadata_format` to more precisely specify what it's expected to control. `metaschema` is now on the disallow list via https://github.com/mozilla/mozilla-schema-generator/pull/137 so I'm going to go ahead and merge.

PR #570, opened 2020-07-10
Bug 1647825 - Add a schema for the new ping 'xfocsp-error=report'.

artines1:
Checklist for reviewer:

- [x] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title)
- [x] Scan the PR and verify that no changes (particularly to `.circleci/config.yml`) will cause environment variables (particularly credentials) to be exposed in test logs
- [ ] If the PR comes from a fork, trigger the `integration` CI test by pushing this revision [as discussed in the README](https://github.com/mozilla-services/mozilla-pipeline-schemas#packaging-and-integration-tests-optional) and review the report posted in the comments.

For glean changes:
- [ ] Update `include/glean/CHANGELOG.md`

PR #581, opened 2020-07-22
Add xfocsp-error-report schema

jklukas:
Closes #570

This pulls in the commit from that PR and duplicates it in a new namespace.
We will be rerouting these pings in the pipeline so that they can have
stricter access controls.

There is a test case added here so that CI will fail if these two schemas
get out of sync.

Checklist for reviewer:

For glean changes:


jklukas:
> The schema itself looks fine. Why does the schema need to exist in the telemetry namespace if it will be rerouted to another pipeline family?

I had thought we would need to validate based on the telemetry schema, but then create the table based on the other schema, but it looks like we can do the rerouting immediately after parsing the URI.

For sanity, though, I think it makes sense to keep a copy under `telemetry` since that's what clients will be sending.

Assuming the [pipeline metadata schema proposal](https://docs.google.com/document/d/1oNi-eX_IPIZa8C0h5GZnWTM0WFivZ1hPMgMMX3c3HHs/edit?ts=5f169a98#) moves forward, we would refactor this in the future such that only the schema under `telemetry` exists, and the routing information is encoded in that schema.

PR #582, opened 2020-07-23
Remove client id from xfocsp sample ping

whd:
https://bugzilla.mozilla.org/show_bug.cgi?id=1654558#c11


validation/telemetry/xfocsp-error-report.4.sample.pass.json
```diff
@@ -1,6 +1,5 @@
 {
     "type": "xfocsp-error-report",
-    "id": "0cb9115c-471b-164b-9630-9fb75567eca8",
```
jklukas, line 3: This shouldn't be a client_id. The top-level `id` field in the common ping format is expected to contain the document_id. That said, if it's not in the schema, I'm fine not having it here.

PR #584, opened 2020-07-24
Bug 1654809 - Add JWE metric to Glean schema

brizental:
Checklist for reviewer:

For glean changes:


schemas/glean/glean/glean.1.schema.json
```diff
@@ -177,6 +177,18 @@
           },
           "type": "object"
         },
+        "jwe": {
+          "additionalProperties": {
+            "pattern": "^[a-zA-Z0-9\\-_]{1,1024}\\.[a-zA-Z0-9\\-_]{0,1024}\\.([a-zA-Z0-9\\-_]{16})?\\.[a-zA-Z0-9\\-_]{1,1024}\\.([a-zA-Z0-9\\-_]{22})?$",
```
jklukas, line 182: For AET, we perform decryption as early as possible in the pipeline, before we perform schema validation. I believe this would be the case for other JWE pipelines in the future as well.

So at the point we're using the glean JSON schema to validate documents, the JWE has already been replaced with the decrypted value under the decrypted_field_name. So this should just be a plain old string rather than having validating it as a JWE.

jklukas:
I also want @acmiyaguchi to take a look for Glean schema expertise.

jklukas:
@acmiyaguchi I assume we can ignore the BQ schema diff from the integration report, correct? It looks like it's directly translating the glean/glean schema into a BQ schema.

jklukas:
Adding an example for jwe metric in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/585

PR #592, opened 2020-08-17
Bug 1659514 Add reduced environment to telemetry/account-ecosystem

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1635659#c16

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/account-ecosystem/account-ecosystem.4.schema.json
+++ b/schemas/telemetry/account-ecosystem/account-ecosystem.4.schema.json
@@ -52,6 +52,102 @@
       "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]{3}Z$",
       "type": "string"
     },
+    "environment": {
+      "properties": {
+        "profile": {
+          "properties": {
+            "creationDate": {
+              "type": "number"
+            },
+            "firstUseDate": {
+              "type": "number"
+            }
+          },
+          "required": [
+            "creationDate"
+          ],
+          "type": "object"
+        },
+        "settings": {
+          "properties": {
+            "locale": {
+              "type": [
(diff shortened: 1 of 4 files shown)


schemas/telemetry/account-ecosystem/account-ecosystem.4.schema.json
```diff
@@ -74,6 +170,9 @@
           },
           "type": "array"
         },
+        "reason": {
+          "type": "string"
+        },
```
jklukas, line 175: I also add payload.reason here, which is currently getting sent (and stored in additional_properties) and is quite standard for common pings.

templates/telemetry/pre-account/pre-account.4.schema.json
```diff
@@ -15,84 +16,6 @@
       "minimum": 4,
       "maximum": 4
     },
-    "environment": {
```
jklukas, line 18: The environment fields are the same as for the old pre-account ping, so I've factored out that existing definition to an include that now appears both here and in the new account-ecosystem ping.

jklukas:
@acmiyaguchi I'm seeing in the generated BQ schema diff that many of the new fields are marked "REQUIRED", which doesn't seem right. Shouldn't our BQ schemas have only NULLABLE and REPEATED fields?

acmiyaguchi:
@jklukas The CI used in the repo is slightly different from how we actually deploy it. We don't force all values to be nullable here, in particular (https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/7ea6c19b448f4785b2b43ad40229cb24c5de91e1/scripts/bigquery_schema_diff.py#L39-L53 vs https://github.com/mozilla/mozilla-schema-generator/blob/bafad8868988ed54ff57a57c0f063fbe903ada50/bin/schema_generator.sh#L171-L177).

PR #598, opened 2020-08-27
Introduce "pioneer" value for bq_metadata_format

jklukas:
This should bring parity with the [regex logic on filepaths in m-s-g](https://github.com/mozilla/mozilla-schema-generator/blob/0bdc35a407b991c3cf6de591a4afee6a8d8ea61a/bin/schema_generator.sh#L76-L91) Once this and https://github.com/mozilla/mozilla-schema-generator/pull/139 are merged, it should be possible to update m-s-g's metadata_merge to look up the appropriate metadata schema based on bq_metadata_format and remove the regex logic.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/metadata/metaschema/metaschema.1.schema.json
+++ b/schemas/metadata/metaschema/metaschema.1.schema.json
@@ -15,7 +15,8 @@
           "description": "NOT YET IMPLEMENTED: The logical format for the metadata struct in the destination BigQuery table",
           "enum": [
             "structured",
-            "telemetry"
+            "telemetry",
+            "pioneer"
           ],
           "type": "string"
         },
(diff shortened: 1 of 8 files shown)


jklukas:
I'm going to do some more thinking about the implications of not yet having metadata generation fully automated such that new `pioneer-` namespaces would be included.

jklukas:
I added tests that should prevent us from accidentally:

1. Adding new pioneer- namespaces without proper metadata, and
2. Adding new telemetry doctypes with version other than 4

whd (approved):
The test seems reasonable for now and provides clear instructions on how to proceed. It's also not immediately obvious to me how we'd automate what I'm calling "namespace family" metadata for pioneer studies. "namespace family" does map 1:1 to pipeline family in the telemetry and pioneer cases (also to regular expressions) but it would be mixing purposes to use pipeline families this way.

PR #608, opened 2020-09-10
Bug 1658787 - Add Pocket on-save recs ping

kirill-demtchouk:
Checklist for reviewer:

For glean changes:

This creates a new telemetry ping to track impressions, clicks, and saves of on-save recommendations shown to users when they save a webpage via the Pocket button in the address bar.

I have included most standard information that we track on AS pings and entered this ping into the `activity-stream` namespace, as per guidance from @ncloudioj . This ping does not have the FX `client_id`, but instead uses the Pocket `impression_id` as a user identifier.

The only new parameters of note are:
-`model` = name of the machine learning model that generated these recommendations

### DIFF
--- a/templates/activity-stream/on-save-recs/on-save-recs.1.schema.json
+++ b/templates/activity-stream/on-save-recs/on-save-recs.1.schema.json
@@ -5,21 +5,33 @@
   "properties": {
     @ACTIVITY-STREAM_IMPRESSIONID_1_JSON@,
     @ACTIVITY-STREAM_EXPERIMENTS_1_JSON@,
-    "action_position": {
-      "description": "A zero based integer indicating the position of this event",
-      "type": "integer"
-    },
     "addon_version": {
       "type": "string"
     },
-    "event": {
-      "enum": [
-        "impression",
-        "click",
-        "save"
-      ],
-      "description": "An event identifier",
-      "type": "string"
+    "events": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "action": {
+            "enum": [
+              "impression",
+              "click",
+              "save"
+            ],
+            "description": "An event identifier",
+            "type": "string"
+          },
+          "position": {
+            "description": "A zero based integer indicating the position of this event",
+            "type": "integer"
+          }
+        },
+        "required": [
+          "action",
+          "position"
+        ],
+      }
     },
     "locale": {
       "type": "string"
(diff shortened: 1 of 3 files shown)


templates/activity-stream/on-save-recs/on-save-recs.1.schema.json
```diff
@@ -0,0 +1,51 @@
...
+    },
+    "event": {
+      "enum": [
+        "impression",
+        "click",
+        "save"
+      ],
+      "description": "An event identifier",
+      "type": "string"
+    },
```
jklukas, line 23: Have you considered replacing this with an `events` array with `action` and `position` fields to allow multiple impressions to be expressed in a single ping? I'm assuming that we generally present a series of impressions all at once, so we could make a single call. That would save some bandwidth for the client, reduce the raw storage size, and cut down on some per-ping overhead in the pipeline. We could provide a view that performs the `UNNEST` necessary to transform it into this flattened representation.

This has been discussed a bit for previous A-S pings and I believe was rejected because it would have required additional code changes, and in those cases we were replacing existing ping-centre calls with identical payloads to the telemetry pipeline. If this is going to be new code on the client side, though, allowing multiple events per ping might make sense.

jklukas (commented):
This looks all looks correct. I just have the one question about whether we should allow multiple events per ping.

templates/activity-stream/on-save-recs/on-save-recs.1.schema.json
kirill-demtchouk: Sounds reasonable, making this change now.

templates/activity-stream/on-save-recs/on-save-recs.1.schema.json
```diff
@@ -0,0 +1,63 @@
...
+          },
+          "position": {
+            "description": "A zero based integer indicating the position of this event",
+            "type": "integer"
+          }
+        },
+        "required": [
+          "action",
+          "position"
+        ],
```
jklukas, line 33: Extraneous comma here. I believe that's what's causing the test failure.

kirill-demtchouk: Good catch, the JSON validator I was using didn't catch this!

jklukas (approved):
Looks good. I just triggered integration tests and will merge when those finish.

Schema deploys happen approximately once per day, so the pipeline should be ready tomorrow to accept payloads with this new schema and destination BQ tables will be automatically created.

PR #610, opened 2020-09-10
Bug 1663857 - Add pioneer-enrollment schemas to debug and meta

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1663857)

Checklist for reviewer:

### DIFF
--- a/templates/pioneer-meta/pioneer-enrollment/pioneer-enrollment.1.schema.json
+++ b/templates/pioneer-meta/pioneer-enrollment/pioneer-enrollment.1.schema.json
@@ -1,8 +1,1 @@
-{
-  "$schema": "http://json-schema.org/draft-04/schema#",
-  "type": "object",
-  "title": "pioneer-enrollment",
-  "$comment": "Bug 1663857 - Add pioneer-enrollment pings for explict consent logging.",
-  "properties": {},
-  "additionalProperties": false
-}
+@PIONEER-STUDY_PIONEER-ENROLLMENT_1_JSON@
--- a/templates/pioneer-meta/pioneer-enrollment/pioneer-enrollment.1.schema.json
+++ b/templates/pioneer-meta/pioneer-enrollment/pioneer-enrollment.1.schema.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "pioneer-enrollment",
+  "$comment": "Bug 1663857 - Add pioneer-enrollment pings for explict consent logging.",
+  "properties": {},
+  "additionalProperties": false
+}
--- a/tests/test_validation_java.py
+++ b/tests/test_validation_java.py
@@ -21,6 +21,8 @@ try:
 except SystemError as ex:
     JAVA_CONFIGURED_REASON = f"{JAVA_CONFIGURED_REASON}: {str(ex)}"
 except Exception as ex:
+    # JavaException cannot be imported when javac is not found on the system.
+    # Just log the exception and move on with the rest of the tests instead.
     JAVA_CONFIGURED_REASON = f"{JAVA_CONFIGURED_REASON}: {str(ex)}"
 
 
--- a/tests/test_validation_java.py
+++ b/tests/test_validation_java.py
@@ -20,7 +20,7 @@ try:
     ENABLE_JAVA_TESTS = True
 except SystemError as ex:
     JAVA_CONFIGURED_REASON = f"{JAVA_CONFIGURED_REASON}: {str(ex)}"
-except JavaException as ex:
+except Exception as ex:
     JAVA_CONFIGURED_REASON = f"{JAVA_CONFIGURED_REASON}: {str(ex)}"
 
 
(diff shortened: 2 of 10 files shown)


acmiyaguchi:
cc @Dexterp37

tests/test_validation_java.py
```diff
@@ -20,7 +20,7 @@
     ENABLE_JAVA_TESTS = True
 except SystemError as ex:
     JAVA_CONFIGURED_REASON = f"{JAVA_CONFIGURED_REASON}: {str(ex)}"
-except JavaException as ex:
+except Exception as ex:
```
jklukas, line 23: Can you document why we need to catch raw `Exception` here? It leaves us open to unexpected behavior.

templates/pioneer-meta/pioneer-enrollment/pioneer-enrollment.1.schema.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "pioneer-enrollment",
+  "$comment": "Bug 1663857 - Add pioneer-enrollment pings for explict consent logging.",
+  "properties": {},
+  "additionalProperties": false
+}
```
jklukas, line 8: Given that this ping is duplicated in many namespaces, can we factor this whole structure out to `templates/include/pioneer` for consistency?

templates/pioneer-meta/pioneer-enrollment/pioneer-enrollment.1.schema.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "pioneer-enrollment",
+  "$comment": "Bug 1663857 - Add pioneer-enrollment pings for explict consent logging.",
+  "properties": {},
+  "additionalProperties": false
```
jklukas, line 7: Why aren't we giving these pings more explicit structure? It looks like we have a validation example with a well-defined format.

jklukas (commented):
I think I'm missing some context here, and have added a few questions inline.

I've scanned through the bug, but it's still not clear to me why we aren't defining concrete fields for these pings. The bug links to a document that I don't have access to, which may be relevant.

tests/test_validation_java.py
acmiyaguchi: I documented this in the commit message, but I'll put a comment in the code too.

templates/pioneer-meta/pioneer-enrollment/pioneer-enrollment.1.schema.json
acmiyaguchi: `"additionalProperties": false` with an empty object schema is the most explicit schema for an empty document. The validation document is the envelope, this is the actual message. It's being treated like the deletion ping. See https://github.com/mozilla/gcp-ingestion/pull/1394 for more context.

tests/test_validation_java.py
jklukas: What exception is raised if `javac` is not found? `ImportError`? Can we catch that explicitly?

If we stick with catching `Exception`, then we should remove the `SystemError` clause above, since it's a subclass of `Exception` anyway. If we catch a more specific exception, we can still collapse these into a single line like:

```
except (JavaException, ImportError) as ex:
```

acmiyaguchi: It throws a JavaException on `from jnius import JavaException` if `javac` doesn't exist, so there's a bit of a chicken and egg problem going on there.

jklukas (approved):
r+wc

tests/test_validation_java.py
acmiyaguchi: Actually, I recalled this wrong, it's actually a subprocess exception. 

```
tests/test_validation_java.py:4: in <module>
    from jnius import JavaException
venv/lib/python3.8/site-packages/jnius/__init__.py:42: in <module>
    from .reflect import *  # noqa
venv/lib/python3.8/site-packages/jnius/reflect.py:17: in <module>
    class Class(with_metaclass(MetaJavaClass, JavaClass)):
venv/lib/python3.8/site-packages/six.py:856: in __new__
    return meta(name, resolved_bases, d)
jnius/jnius_export_class.pxi:114: in jnius.MetaJavaClass.__new__
    ???
jnius/jnius_export_class.pxi:164: in jnius.MetaJavaClass.resolve_class
    ???
jnius/jnius_env.pxi:11: in jnius.get_jnienv
    ???
jnius/jnius_jvm_dlopen.pxi:118: in jnius.get_platform_jnienv
    ???
jnius/jnius_jvm_dlopen.pxi:70: in jnius.create_jnienv
    ???
jnius/jnius_jvm_dlopen.pxi:47: in jnius.find_java_home
    ???
/usr/lib/python3.8/subprocess.py:411: in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
/usr/lib/python3.8/subprocess.py:512: in run
    raise CalledProcessError(retcode, process.args,
```

I'm going to collapse this down to a single exception though, since the error is logged and can be shown with `pytest -rs`:

```
SKIPPED [150] tests/test_validation_java.py:36: failed to configure java: Command '['which', 'javac']' returned non-zero exit status 1.
SKIPPED [52] tests/test_validation_java.py:46: failed to configure java: Command '['which', 'javac']' returned non-zero exit status 1.
```

PR #612, opened 2020-09-11
Add glean client_info descriptions

wlach:
* Add glean client info descriptions to base glean schema
* Add a simple script to keep these up to date

Checklist for reviewer:

For glean changes:


templates/include/glean/client_info.1.schema.json
```diff
@@ -0,0 +1,67 @@
...
+    },
+    "os": {
+      "description": "The name of the operating system. Possible values: Android, iOS, Linux, Darwin, Windows, FreeBSD, NetBSD, OpenBSD, Solaris, unknown",
+      "type": "string"
+    },
+    "os_version": {
+      "description": "The user-visible version of the operating system (e.g. \"1.2.3\"). If the version detection fails, this metric gets set to `Unknown`.",
+      "type": "string"
+    },
+    "telemetry_sdk_build": {
```
wlach, line 53: This appears not to be "documented" in the Glean SDK. @Dexterp37 is this an oversight?

wlach:
After having gone to the trouble of writing this PR, I found out that the Glean SDK keeps its own copy of this file here for validation purposes:

https://github.com/mozilla/glean/blob/main/glean.1.schema.json

It is kept in sync with this script:

https://github.com/mozilla/glean/blob/main/bin/update-schema.sh

Longer term, I wonder if we should invert this relationship: keep/generate the schema inside glean itself, and then copy it over into this repository as needed.

fbertsch (commented):
Can we validate that the descriptions are up-to-date in ci, [like we do here](https://github.com/mozilla/bigquery-etl/blob/master/.circleci/config.yml#L109)?

wlach:
> Can we validate that the descriptions are up-to-date in ci, [like we do here](https://github.com/mozilla/bigquery-etl/blob/master/.circleci/config.yml#L109)?

I would feel a bit bad doing that, because it could cause PRs to fail for no fault of the person writing them. I think a better solution (long term) is what I outlined above: moving the schema generation (and the schema itself) to glean.

fbertsch (approved):
@wlach I'll leave it up to you then - for now this change seems appropriate.

templates/include/glean/client_info.1.schema.json
Dexterp37: You can find the docs for that [here](https://mozilla.github.io/glean/book/user/pings/index.html#the-client_info-section).

scripts/extract_glean_client_info_descriptions
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
```
Dexterp37, line 1: This file is missing the `.py` suffix and the license header . Any chance you could fix this before merging?

Dexterp37 (approved):

scripts/extract_glean_client_info_descriptions
wlach: I'm inclined to leave this as-is:

1. The other python scripts don't have .py extensions
2. We don't generally add MPL license headers for internal tooling like this, as we consider the top-level license sufficient context. See e.g. mozilla/gcp-ingestion#777

Dexterp37: Oh, TIL. Cool, go ahead then!

templates/include/glean/client_info.1.schema.json
wlach: Got it. Should this be added here:

https://github.com/mozilla/glean/blob/ccf91a74dc0d95e5b62df2452187157e882f1907/glean-core/metrics.yaml#L32

?

Dexterp37: Oh right, you're right.

PR #615, opened 2020-09-15
Remove android-anr-report schemas

jklukas:
I don't have much context on ANR reporting beyond the high-level overview
available in the [initial ANR reporting
bug](https://bugzilla.mozilla.org/show_bug.cgi?id=833990).

I have confirmed, however, that the relevant tables are completely empty in BQ:

```
SELECT
  _table_suffix AS tbl,
  COUNT(*) AS n
FROM
  `moz-fx-data-shared-prod.telemetry_stable.android_anr_report_*`
WHERE
  DATE(submission_timestamp) >= "2010-01-01"
GROUP BY
  1
ORDER BY
  2 DESC
```

So it would appear ANR reporting no longer exists or at least does not function.
I recommend that we remove these schemas and all the associated tables in BQ
to reduce overall noise in the infrastructure.

Checklist for reviewer:

For glean changes:


jklukas:
Tagging @whd here for review. Does this seem like reasonable cleanup? There's no data to archive here, so I don't see that we lose anything by deleting the tables.

whd (approved):
I'm fine with doing this but again, hold off on merging until the current probe scraper run completes and is deployed.

I'm curious though why we created this table if we didn't have any data in AWS, but not curious enough to look into it any further.

PR #617, opened 2020-09-15
Bug 1658787 - Drop region field from on-sav-recs ping

kirill-demtchouk:
Checklist for reviewer:

For glean changes:

It turns out the data pipeline adds region information based on IP address to all events by default, so no need for the `region` parameter I had added to this ping [earlier](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/608).

### DIFF
--- a/schemas/activity-stream/on-save-recs/on-save-recs.1.schema.json
+++ b/schemas/activity-stream/on-save-recs/on-save-recs.1.schema.json
@@ -64,9 +64,6 @@
     "profile_creation_date": {
       "type": "integer"
     },
-    "region": {
-      "type": "string"
-    },
     "release_channel": {
       "type": "string"
     },
--- a/templates/activity-stream/on-save-recs/on-save-recs.1.schema.json
+++ b/templates/activity-stream/on-save-recs/on-save-recs.1.schema.json
@@ -43,9 +43,6 @@
     "profile_creation_date": {
       "type": "integer"
     },
-    "region": {
-      "type": "string"
-    },
     "release_channel": {
       "type": "string"
     },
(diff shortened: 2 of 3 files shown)


jklukas (approved):
I discussed this change with @kirill-demtchouk on Slack. Normally, deleting a field would be an incompatible change for BQ, but we're able to get around that in this case because the tables have not yet been deployed.

acmiyaguchi (commented):
This schema has already been deployed to BigQuery, so the field cannot be dropped due to schema evolution rules. https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/a4be85b9bc2b9904a23cc1808963acfb04a71df2/schemas/activity-stream/on-save-recs/on-save-recs.1.bq#L245-L249

I would suggest a comment about deprecation.

whd:
The latest schemas deploy to production was 202009081456_5e5c475 due to ongoing schemas pipeline issues. We may have deployed a version of this to stage, but this can likely be deleted and re-created as I'm now made aware of this change.

jklukas:
> We may have deployed a version of this to stage, but this can likely be deleted and re-created as I'm now made aware of this change.

I checked nonprod in the BQ console before merging to verify that the live table didn't exist, so I thought this was safe. If this is indeed partially deployed, then it's definitely on me for not verifying with @whd directly.

whd:
> I checked nonprod in the BQ console before merging to verify that the live table didn't exist, so I thought this was safe. If this is indeed partially deployed, then it's definitely on me for not verifying with @whd directly.

Ok that was sufficient then and you didn't need to ping me. I qualified my statement with "may" because I've been rolling out ACL changes to stage (and prod) and the ops machinery proposed to delete some tables in stage due to the latest schemas being bad. I didn't check whether these tables were among those, so I'm now assuming the latest schemas change will introduce more tables than these.

Basically I'm just being very conservative until things are working correctly.

PR #621, opened 2020-09-18
Bug 1665966: Reject pings with invalid histogram keys

mdboom:
We've found exactly one ping with an invalid character in a histogram key.  (Who knows why?  Bad disk?  Cosmic rays?)

But in any event we shouldn't ingest these into the tables.  The regex used to enforce that the keys are numeric was incorrect, because it would match as long as any character was numeric, not only when the entire string was numeric.

Checklist for reviewer:

For glean changes:


schemas/glean/glean/glean.1.schema.json
```diff
@@ -136,7 +136,7 @@
                   "type": "integer"
                 },
                 "propertyNames": {
-                  "pattern": "[0-9]+"
+                  "pattern": "^[0-9]+$"
```
badboy, line 139: Do we know which regex engine is used for this check? Because `^` and `$` usually mean `line-start` and `line-end`, thus if the data contains a newline it could still be invalid (plus the fact that a newline-containing value is wrong anyway in our case)

mdboom: JSON Schema specifies the [ECMA 262 regex syntax](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.3.3).  In that case, `^` and `$` are beginning and end of input, respectively, unless the [multiline flag is set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Assertions), but I can only assume that it isn't for JSON Schema (and experimentation confirms that)...

mdboom:
@fbertsch: Any thoughts on why the [test is failing?](https://circleci.com/gh/mozilla-services/mozilla-pipeline-schemas/4381?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

fbertsch (approved):
r+ once we get the test results. I'm looking into that failure now. Offhand it looks like the `diff` function is missing from the container

fbertsch:
I cannot repro this failure locally. @acmiyaguchi any ideas?

acmiyaguchi:
@fbertsch Sorry for the delay in response, the issue is that diffutils isn't installed on the container. I reproduced this by running the CircleCI job with SSH and starting the docker instance that runs the BigQuery schema diffing. After installing diff (`dnf install diffutils`), this completes successfully:

```
[root@fdb25781a0aa app]# ls -alh integration/ 
total 20K
drwx------  3 root root 4.0K Sep 23 17:10 .
drwxr-xr-x 15 root root 4.0K Sep 23 17:10 ..
drwxr-xr-x  2 root root  12K Sep 23 17:10 b8db7f86
-rw-r--r--  1 root root    0 Sep 23 17:10 bq_schema_b8db7f86-b8db7f86.diff
```

This is an empty diff. I think in this case it should be okay to push the schema through with this failing test.

acmiyaguchi (approved):

PR #630, opened 2020-10-09
Note special handling of xfocsp-error-report

jklukas:
Resolves https://github.com/mozilla/mozilla-schema-generator/pull/155#discussion_r501368859

Checklist for reviewer:

For glean changes:


whd (approved):
Not sure how I feel about introducing a new top-level comment field instead of nesting `comment` in `mozPipelineMetadata`, but either is better than no comment at all. I guess we have a metaschema for `mozPipelineMetadata` that this would need to be added to, which is not the case for the top level.

jklukas:
Moved comment under mozPipelineMetadata and added to the metaschema.

PR #636, opened 2020-11-05
Bug 1675103 - Define the bhr ping schema.

mikeconley:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/telemetry/bhr/bhr.4.schema.json
+++ b/templates/telemetry/bhr/bhr.4.schema.json
@@ -1,1 +1,84 @@
-@COMMON_PLACEHOLDER_1_JSON@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$comment": "Bug 1675103 - Define the bhr ping schema.",
+  "type": "object",
+  "title": "bhr",
+  "properties": {
+    @TELEMETRY_APPLICATION_1_JSON@,
+    @TELEMETRY_CLIENTID_1_JSON@,
+    @TELEMETRY_CREATIONDATE_1_JSON@,
+    @TELEMETRY_ENVIRONMENT_1_JSON@,
+    @TELEMETRY_ID_1_JSON@,
+    "type": {
+      "type": "string",
+      "enum": [ "bhr" ]
+    },
+    "version": {
+      "type": "number",
+      "minimum": 4,
+      "maximum": 4
+    },
+    "payload": {
+      "type": "object",
+      "properties": {
+        "modules": {
+          "type": "array",
+          "items": {
+            "type": "array",
+            "items": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "string"
+              }
+            ],
+            "additionalItems": false
+          }
+        },
+        "hangs": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "duration": {
+                "type": "number",
+                "minimum": 0.0
+              },
+              "thread": {
+                "type": "string"
+              },
+              "runnableName": {
+                "type": "string"
+              },
+              "process": {
+                "type": "string"
+              },
+              "remoteType": {
+                "type": ["string", "null"]
+              },
+              "annotations": {
+                "type": "array",
+                "items": {
+                  "type": "array",
+                  "items": [
+                    {
+                      "type": "string"
+                    },
+                    {
+                      "type": "string"
+                    }
+                  ],
+                  "additionalItems": false
+                }
+              }
+            }
+          }
+        },
+        "timeSinceLastPing": {
+          "type": "integer"
+        }
+      }
+    }
+  }
+}
--- a/schemas/telemetry/bhr/bhr.4.schema.json
+++ b/schemas/telemetry/bhr/bhr.4.schema.json
@@ -1,12 +1,1410 @@
 {
-  "$id": "moz://mozilla.org/schemas/placeholder/schema/1",
+  "$comment": "Bug 1675103 - Define the bhr ping schema.",
   "$schema": "http://json-schema.org/draft-04/schema#",
   "mozPipelineMetadata": {
     "bq_dataset_family": "telemetry",
     "bq_metadata_format": "telemetry",
     "bq_table": "bhr_v4"
   },
-  "properties": {},
-  "title": "placeholder_schema",
+  "properties": {
+    "application": {
+      "additionalProperties": false,
+      "properties": {
+        "architecture": {
+          "type": "string"
+        },
+        "buildId": {
+          "pattern": "^[0-9]{10}",
+          "type": "string"
+        },
+        "channel": {
+          "type": "string"
+        },
+        "displayVersion": {
+          "pattern": "^[0-9]{2,3}\\.",
+          "type": "string"
+        },
+        "name": {
+          "type": "string"
+        },
+        "platformVersion": {
+          "pattern": "^[0-9]{2,3}\\.",
+          "type": "string"
+        },
+        "vendor": {
+          "type": "string"
+        },
+        "version": {
+          "pattern": "^[0-9]{2,3}\\.",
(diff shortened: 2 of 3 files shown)


acmiyaguchi (commented):
Unfortunately we don't accept versions other than 4 for the telemetry pipeline due to schemas being tightly coupled to the telemetry module. You should be able to replace the placeholder schema with this schema, though, since it's currently a placeholder. We only version schemas when the underlying data is incompatible (e.g. types for properties change, or columns are removed).

templates/telemetry/bhr/bhr.4.schema.json
```diff
@@ -1 +1,95 @@
...
+          "items": {
+            "type": "array",
+            "items": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "string"
+              }
+            ]
```
acmiyaguchi, line 34: Please specify `"additionalItems": false`. If this is a variable length tuple, then you can set `additionalItems` to a single type and specify a `maxItems` property.

templates/telemetry/bhr/bhr.4.schema.json
```diff
@@ -1 +1,95 @@
...
+                  "type": "array",
+                  "items": [
+                    {
+                      "type": "string"
+                    },
+                    {
+                      "type": "string"
+                    }
+                  ]
+                }
```
acmiyaguchi, line 70: `annotations` needs to have `additionalItems` set to be treated as an array of anonymous tuples.

acmiyaguchi (changes requested):
Thanks for the schema, there are a couple of changes that will need to be modified so it passes our transpilation process. This document is particularly tricky because of the nested arrays.

I made a few changes to the CI tool locally, and this is the result that I get when dropping the stack traces and adding `"additionalItems": false` in the appropriate places. 

```
mps bigquery transpile schemas/telemetry/bhr/bhr.4.schema.json | mps bigquery columns /dev/stdin | grep payload
root.payload.hangs.[].annotations.[].f0_ STRING
root.payload.hangs.[].annotations.[].f1_ STRING
root.payload.hangs.[].duration FLOAT64
root.payload.hangs.[].process STRING
root.payload.hangs.[].remote_type STRING
root.payload.hangs.[].runnable_name STRING
root.payload.hangs.[].thread STRING
root.payload.modules.[].f0_ STRING
root.payload.modules.[].f1_ STRING
root.payload.time_since_last_ping INT64
```

mikeconley:
I _think_ I've correctly applied the review feedback. Let me know if I've missed anything. Thanks!

acmiyaguchi:
Could I ask you to rebase onto the current master? There are a couple of CI changes in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/643 that will be handy during review.

mikeconley:
Sure, done!

mikeconley:
Oops, looks like I'm failing CI. Back to the drawing board.

acmiyaguchi:
Okay, this looks good to me, the types in the `root.payload` columns look correct. I'm going to leave it for Monday to merge in so that we can check on the schema error rate on deploy.

I did a quick check on another branch I have for looking at the transformed validation document for more insight into what's going to happen. 

In the hangs column will have this repeated structure:
```
% mps bigquery transform validation/telemetry/bhr.4.sample.pass.json | jq '.payload.hangs[0]'
{
  "duration": 157.073998,
  "thread": "Gecko_Child",
  "process": "tab",
  "annotations": [
    {
      "f0_": "PendingInput",
      "f1_": "709"
    }
  ],
  "runnable_name": "TimeoutExecutor Runnable",
  "remote_type": "webIsolated"
}
```

The additional_properties field is going to have an JSON blob that contains a the stack object that doesn't make it into the structure portion of the schema, which should have a 1:1 array mapping:

```
% mps bigquery transform validation/telemetry/bhr.4.sample.pass.json | \
jq '.additional_properties | fromjson | .payload.hangs[0]' 
{
  "stack": [
    "XRE_InitChildProcess",
    "nsTimerImpl::Fire",
    "",
    "(content script)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)",
    "(unresolved)"
  ]
}
```

Do you know what you want to do with the stacks columns long term?

templates/telemetry/bhr/bhr.4.schema.json
```diff
@@ -1 +1,83 @@
-@COMMON_PLACEHOLDER_1_JSON@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "event",
```
acmiyaguchi, line 4: This is a good place to add a description field linking out to the bug where this is first implemented. The title should probably be `bhr`?

acmiyaguchi (approved):

mikeconley:
> Do you know what you want to do with the stacks columns long term?

Similar to crash stacks, we're probably going to want a way of symbolicating them and generating signatures for them. Something like what's happening in https://bugzilla.mozilla.org/show_bug.cgi?id=1631563 for crash pings.

acmiyaguchi:
Does this require context from the full hangs array? It might spare a few headaches if `stack` got moved into the repeated array as a JSON string instead of being housed in the additional_properties field. This is something I can do with the heavy lifting of the schema definition done here.

mikeconley:
> Does this require context from the full hangs array?

I'm honestly not sure what you mean. But I certainly trust you more to know about the best way to order this stuff in the databases than I do myself. :)

acmiyaguchi:
To reframe my question, flattening out the data will look something like this:

```sql
SELECT
  id,
  index,
  hangs.*
  JSON_EXTRACT(additional_properties, CONCAT("$.payload.hangs[", index, "].stack")) AS stack
FROM `moz-fx-data-shared-prod`.telemetry.bhr,
UNNEST(payload.hangs) AS hang
WITH OFFSET AS index
```

In the result from this query (which flattens out the `hangs` column), is symbolicating the stack just `symbolicate(stack) AS stack_symbol` for some `symbolicate` function?

And I'm suggesting a change to https://github.com/mozilla/mozilla-schema-generator to get around some of the hacky stuff required to avoid the `JSON_EXTRACT` function. I'm convinced it's not a bad idea though, so I'll probably go ahead with it.

mikeconley:
Yeah, that looks like the right kind of query to me!

The symbolication step is still being worked out, I think. @willkg or @wlach might have details on how it could work, but effectively it's a service call-out to transform the memory addresses into readable symbols. I'm not sure whether or not it can be a function like that, or if it'd be a periodic post-processing task.

willkg:
I don't know how crash ping symbolication and signature generation is going to work, yet. I'm hoping to get to that in early 2021.

mikeconley:
Thanks for the review, @acmiyaguchi! Having never contributed to this repository before, I'm not 100% sure what the procedure is now. Am I still waiting on approval from @mreid-moz? Or do I trigger a landing somehow? Or is there something else I should wait for?

acmiyaguchi:
I'm going to merge this now, it will get triggered automatically at UTC+2. I didn't merge it on Friday because of the weekend. Check https://protosaur.dev/mps-deploys/ for the actual deploy time. The new data should be available in the new format in `telemetry_live.bhr_v4` 10-15 minutes after deploy, while `telemetry.bhr` will be available once a full day has elapsed.

acmiyaguchi:
Some follow-up, it looks like things deployed correctly.

Data is being inserted into the table correctly: https://sql.telemetry.mozilla.org/queries/76767/source
We're also not seeing any schema validation errors: https://sql.telemetry.mozilla.org/queries/76768/source

If you want to access data before this time, you will need to work with the `additional_properties` field.

PR #638, opened 2020-11-10
Bug 1672763 - Add pocket-button ping.

kirill-demtchouk:
Checklist for reviewer:

For glean changes:
- [x] Update `templates/include/glean/CHANGELOG.md`

This PR updates a ping introduced in an earlier [PR](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/608) and generalizes it to track all types of interactions with the Pocket button in the nav bar. This ping does not have the FX client_id, but instead uses the Pocket impression_id as a user identifier.

The possible events it can be used to track, as well as test events that I've included:
- [impressions](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/638/files#diff-f8b701b9cdee099f01bfb9798ef670e013a8ab9bd5fa47d55ac7e339675116d7) of on-save recommendations
- [clicks](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/638/files#diff-7a72508d2dbb9e53f6025e0645f87a4ee5473c07916860fb4fa58018e475dbcc) of on-save recommendations
- [saves](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/638/files#diff-925ad8a1425502fdb980d9559bcfcf65036370931af31319080482a57de6e81a) of on-save recommendations
- [clicks](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/638/files#diff-af4dc0c2f2f2c098459ee30bae91cf8d69f47c3a670402e09155fd2e92a621b3) of the Pocket button itself
- [removal](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/638/files#diff-fb6c0f86274f3b3e03b75534cd3815f22cbad5ebc8217a0273bc2b44716a978a) of the Pocket button from the nav bar
- [clicks](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/638/files#diff-6347021187cf7301911f188fe23d0d8ad3efaa3f862c71535f31e2390451c0c6) on any of the buttons/links in the door hanger menu shown when a logged-out user clicks the Pocket button

### DIFF
--- a/templates/activity-stream/pocket-button/pocket-button.1.schema.json
+++ b/templates/activity-stream/pocket-button/pocket-button.1.schema.json
@@ -15,7 +15,7 @@
               "impression",
               "click",
               "save",
-              "remove"
+              "unpin"
             ],
             "description": "An event identifier.",
             "type": "string"
--- a/templates/activity-stream/pocket-button/pocket-button.1.schema.json
+++ b/templates/activity-stream/pocket-button/pocket-button.1.schema.json
@@ -1,0 +1,66 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "pocket-button",
+  "properties": {
+    @ACTIVITY-STREAM_IMPRESSIONID_1_JSON@,
+    @ACTIVITY-STREAM_EXPERIMENTS_1_JSON@,
+    "events": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "action": {
+            "enum": [
+              "impression",
+              "click",
+              "save",
+              "remove"
+            ],
+            "description": "An event identifier.",
+            "type": "string"
+          },
+          "source": {
+            "description": "An identifier for where the event occurred (e.g. what part of the Pocket button UI a click was performed on).",
+            "type": "string"
+          },
+          "position": {
+            "description": "A zero based integer indicating the position of this event.",
+            "type": "integer"
+          }
+        },
+        "required": [
+          "action",
+          "source"
+        ]
+      }
+    },
+    "locale": {
+      "type": "string"
+    },
+    "model": {
+      "description": "(In the case of actions related to on-save recs) An identifier for the machine learning model used to generate the recommendations.",
+      "type": "string"
+    },
+    "pocket_logged_in_status": {
+      "description": "A boolean for whether the user was logged-in to the Pocket Firefox integration when they performed the indicated action(s). True = logged in.",
+      "type": "boolean"
+    },
+    "profile_creation_date": {
+      "type": "integer"
+    },
+    "release_channel": {
+      "type": "string"
+    },
+    "version": {
+      "type": "string"
+    }
+  },
+  "required": [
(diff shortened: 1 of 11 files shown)


validation/activity-stream/pocket-button.1.door_hanger_click.pass.json
```diff
@@ -0,0 +1,14 @@
+{
+  "locale": "en-US",
+  "impression_id": "{94642acb-4996-034b-916c-147da723cc41}",
+  "version": "72.0a1",
+  "release_channel": "nightly",
+  "events": [
+    {
+      "action": "click",
+      "source": "learn_more"
```
kirill-demtchouk, line 9: Possible `source` values for links on the door-hanger menu are:
-`learn_more`
-`log_in`
-`sign_up_1`
-`sign_up_2`

I am not making this an enum, as we may introduce new values in the future.

templates/activity-stream/pocket-button/pocket-button.1.schema.json
```diff
@@ -0,0 +1,66 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "pocket-button",
```
jklukas, line 4: Can you add a top-level "$comment" field to the existing `on-save-recs` schema to indicate that it's deprecated in favor of `pocket-button`? Do we expect we'll continue to receive `on-save-recs` submissions from older Fx installations for a long time? Is there an expected date on which we could completely remove `on-save-recs` considering we only store 180 days of history?

validation/activity-stream/pocket-button.1.door_hanger_click.pass.json
jklukas: Note that string fields and enums are both represented as STRING in BQ, so you're free to update this detail in the schema going forward. If you want to leave this loose here, that's fine. But it's also an option to make it an enum with known existing values, and make future PRs here to add new values before you start sending pings with those values. If this is likely to evolve frequently, then the enum is likely to cause more pain than benefit.

jklukas (approved):
This looks good. I only ask that we add some documentation about what to expect for the future of `on-save-recs` as part of this PR.

validation/activity-stream/pocket-button.1.door_hanger_click.pass.json
kirill-demtchouk: I'd like to leave it as an open string for now, thanks for the detail!

templates/activity-stream/pocket-button/pocket-button.1.schema.json
kirill-demtchouk: Will do on the comment. The legacy data won't be essential once there is a critical mass of users on Firefox releases sending the new ping. I think we're targeting to add this ping into FX84, so probably 2 release cycles after that, we can get rid of it. That would put us at FX86 on 2/23/21.

jklukas:
This looks good to me. Let me know if you're ready to have it merged.

PR #640, opened 2020-11-12
Encourage people to add a descriptions by adding to checklist

wlach:
This came up again in a recent PR, and I feel like it might be worth calling out, since adding descriptions to new fields will help us generate comprehensible views in BigQuery. @jklukas any thoughts?


jklukas (approved):
Good idea!

PR #646, opened 2020-12-08
Add bq_access workgroups metadata

whd:
The intent of this PR is to externalize `tables_access` metadata defined in the ops infrastructure [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery/namespaces.tfvars.json#L92), [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery/namespaces.tfvars.json#L210),
and [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery/namespaces.tfvars.json#L303).

I don't have a corresponding ops infra PR as I'm looking for feedback on the format first. Some safety checks need to be implemented as part of this externalization to gate workgroups ACL changes defined in MPS on operator input. The new metadata is thus marked as NOT IMPLEMENTED. This change will also need a corresponding MSG PR and I'll file a bigquery-etl PR for access metadata as well once we've settled on a format.

The format is mostly the same as what you'd specify in terraform/bq api, but we're using the `workgroup:` prefix for the access list where GCP natively supports the `user:` `serviceAccount:` and `group:` schemes. We could even suffix our workgroups with `@mozilla.com` or something but right now workgroup identifiers aren't valid email addresses while the others probably are.

The `workgroups` nesting under `access` is to allow for some kind of authorized views metadata in future, or at least compatibility with dataset-level authorized views metadata. I'm not sure what this will look like but maybe we can omit the nesting from tables metadata.


jklukas:
> The workgroups nesting under access is to allow for some kind of authorized views metadata in future

Perhaps we can change the key from `bq_access` to `bq_workgroup_access` such that we could add another separate key for authorized view access in the future.

I like that this is very close to the IAM structure, even if it's rather verbose.

templates/telemetry/account-ecosystem/account-ecosystem.4.schema.json
```diff
@@ -3,6 +3,14 @@
   "type": "object",
   "title": "account-ecosystem",
   "mozPipelineMetadata": {
+    "bq_access": {
+      "workgroups": [
+        {
+          "members": [],
```
jklukas, line 9: This empty list seems strange. Should we just have `"workgroups": []` in this case where we're allowing only administrator-level access?

whd:
I've simplified the metadata structure by keying on `bq_workgroup_access` and replaced the verbose empty access lists with `[]`.

jklukas (approved):
Changes look good.

whd:
I'm closing this as stale but filed https://mozilla-hub.atlassian.net/browse/DSRE-371 to track it for any metadata initiatives in coming years.

PR #576, opened 2020-07-14
Bug 1652842 - Add a blocklist field so the environment schema works with pre- and post-bug-1647225 data

staktrace:
Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title)
- [ ] Scan the PR and verify that no changes (particularly to `.circleci/config.yml`) will cause environment variables (particularly credentials) to be exposed in test logs
- [ ] If the PR comes from a fork, trigger the `integration` CI test by pushing this revision [as discussed in the README](https://github.com/mozilla-services/mozilla-pipeline-schemas#packaging-and-integration-tests-optional) and review the report posted in the comments.

For glean changes:
- [ ] Update `include/glean/CHANGELOG.md`

PR #656, opened 2021-02-10
Bug 1688698 - Add initial schemas for Contextual Services

ncloudioj:
Checklist for reviewer:

For glean changes:

@sunahsuh @BenWu r?

This is the initial commit for Contextual Services schemas, expect more future adjustments :)

A few notes for this PR:
* Unlike the schema specified in the proposal, the impressions and clicks are defined as separate doctypes so that we can apply different processing strategies (e.g. aggregation latency) to them within the pipeline
* `action_url` is not included in the schemas yet as we might want to introduce a unique identifier for each sponsored TopSites tile / QuickSuggest link
* The query parameters `ua` and `ip` of the `reporting_url` are currently set as `%UA%` and `%IP%` (I made them up ;)), we might want to use some other ways to interpolate the reporting URL

Let me know what do you think?


jklukas:
If these are all related to single impressions or clicks, it may be more descriptive to have the document type names be singular, so `quicksuggest-clicks` would be `quicksuggest-click`.

validation/contextual-services/topsites-click.1.event.pass.json
```diff
@@ -0,0 +1,16 @@
+{
+  "context_id": "{94642acb-4996-034b-916c-147da723cc41}",
+  "advertiser": "amazon",
+  "reporting_url": "https://bridge.sfo1.ap01.net/ctp?version=16.0.0&key=1598991514900100001.1&ci=15 98991514792.12747&ctag=1598291839300100006&aespFlag=altinst&ip=%IP%&ua=%UA%",
```
BenWu, line 4: Are these real urls and if so do they reveal anything sensitive?  Would it be better to use placeholder values?

validation/contextual-services/quicksuggest-impression.1.event.pass.json
```diff
@@ -0,0 +1,20 @@
+{
+  "context_id": "{94642acb-4996-034b-916c-147da723cc41}",
+  "search_query": "Puma running shoes",
+  "matched_keywords": "Puma shoes",
+  "is_clicked": false,
+  "advertiser": "amazon",
```
BenWu, line 6: I don't think it really matters but I would prefer not to use real brands for test values

templates/contextual-services/quicksuggest-impression/quicksuggest-impression.1.schema.json
```diff
@@ -0,0 +1,58 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "quicksuggest-impression",
+  "properties": {
+    @ACTIVITY-STREAM_EXPERIMENTS_1_JSON@,
+    "context_id": {
+      "description": "A UUID representing this user. Note that it's not client_id, nor can it be used to link to a client_id",
+      "type": "string",
+      "pattern": "^\\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\\}$"
```
BenWu, line 10: What's the reason for the braces rather than just using standard uuid?  I see that activity stream pings use this too

BenWu (commented):
This looks fine to me as the initial schema.  We can revisit when the rest of the fields are finalized

whd:
Unless this is meant to be accessible to `workgroup:mozilla-confidential`, this data needs to be categorized before this PR is merged. Is there data review somewhere? Do we have a workgroup definition for the group that should have access to this data?

ACL metadata currently lives in the cloudops-infra internal repo, but will eventually live in this repo when https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/646 is fully implemented. Currently an entry similar to  https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery/namespaces.tfvars.json#L340-L410 will need to be added before this PR lands to insure the dataset is properly restricted from the get-go.

validation/contextual-services/topsites-click.1.event.pass.json
ncloudioj: Yes, they are. Though there are a few query parameters to be populated by the data processor before sharing with the partner.

validation/contextual-services/quicksuggest-impression.1.event.pass.json
ncloudioj: Good point, we should do that.

templates/contextual-services/quicksuggest-impression/quicksuggest-impression.1.schema.json
ncloudioj: No particular reason, this is what Firefox UUID generator produces, and I believe it's still a legit v4 UUID.

ncloudioj:
> Unless this is meant to be accessible to `workgroup:mozilla-confidential`, this data needs to be categorized before this PR is merged. Is there data review somewhere? Do we have a workgroup definition for the group that should have access to this data?

Good point, I believe this dataset would require some access control at some point. Is it possible to add ACL later as I don't have a clear answer right now?

Yes, this collection will require the legal & data review, we will do that when we land the client implementation to the tree.

whd:
> Good point, I believe this dataset would require some access control at some point. Is it possible to add ACL later as I don't have a clear answer right now?

I've filed https://github.com/mozilla-services/cloudops-infra/pull/2876 which should make sure that these datasets are restricted when they are first deployed. We can update the `workgroup_access` blocks when we've defined some more concrete access.

As a matter of policy, I don't think we should deploy datasets with the intent to eventually increase restrictions on them. They should start out restricted and we grant access later using workgroups as needed.

ncloudioj:
> I've filed mozilla-services/cloudops-infra#2876 which should make sure that these datasets are restricted when they are first deployed. We can update the workgroup_access blocks when we've defined some more concrete access.

@whd Thanks for setting that up! Shall we file PR to that repo to grant access to individuals for now until #646 lands?

> As a matter of policy, I don't think we should deploy datasets with the intent to eventually increase restrictions on them. They should start out restricted and we grant access later using workgroups as needed.

Makes sense. Let's do this.

whd:
> Shall we file PR to that repo to grant access to individuals for now until #646 lands?

https://github.com/mozilla-services/cloudops-infra/pull/2876 was the PR to restrict access to these datasets when they are provisioned. Data Engineering will have access to these datasets, but client engineering and any other individuals (e.g. you, as a member of `mozilla-confidential`) will not.

We don't grant access to individuals directly, only via workgroups. We can create a workgroup (https://bugzilla.mozilla.org/show_bug.cgi?id=1687669 is an example where we created one) for this initiative but I think that should probably be tied to legal or data review.

If there is an intermediate period between when data is being sent and when workgroup-based access has been codified, you will only be able to access the data via someone in Data Engineering. If you think the process of getting access to data directly should happen before legal or data review, file a bug similar to the aforementioned one for Data SRE to prioritize.

templates/contextual-services/quicksuggest-impression/quicksuggest-impression.1.schema.json
jklukas: We already have some [special handling in the pipeline for validating `impression_id` with the curly braces](https://github.com/mozilla/gcp-ingestion/blob/500cc2d88ad3bbfbb588325b70705121e86cb7ac/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java#L320-L334). Note that we will need some follow-ups to this to support context_id as a user identifier. We'll want to [enable this as a source for sample_id](https://github.com/mozilla/gcp-ingestion/blob/500cc2d88ad3bbfbb588325b70705121e86cb7ac/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java#L251-L260) and we'll need to get this added to deletion-request pings and get configuration added to Shredder.

ncloudioj: Yeah, the "deletion-request" is considered as well in this data collection proposal. We will consult legal to see if it's OK to re-use the "impression_id" here to simplify that work.

jklukas:
We can add a `defaults.schema.json` here similar to the [one for activity-stream](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/b518c65b694324aebb5cd2df56b10364dd18db0d/templates/activity-stream/defaults.schema.json) to set the expiration policy on all these new tables.

ncloudioj:
Finalized the schemas with the following changes:
* Added a filed `tile_id` to all TopSites pings. (QuickSuggest pings already have the counterpart `block_id`)
* Dropped the `action_url` to favor the `tile_id`
* Set `reporting_url` as an optional property as we might want to not send it in the future
* Added a `position` property for all the pings

jklukas:
This looks good to me at this point. I will defer to @sunahsuh for final +1 as she still has the most context at this point.

ncloudioj:
I filed this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1692598) for the workgroup creation as well as for the access permissions.

sunahsuh (approved):
Look good to me, I'll merge this now for deployment tonight

PR #657, opened 2021-02-17
Bug 1688698 - Add 'source' to TopSites schemas in Contextual Services

ncloudioj:
Checklist for reviewer:

For glean changes:

This is a follow up for #656 

r? @sunahsuh @BenWu

### DIFF
--- a/schemas/contextual-services/topsites-click/topsites-click.1.schema.json
+++ b/schemas/contextual-services/topsites-click/topsites-click.1.schema.json
@@ -46,6 +46,14 @@
       "description": "The reporting URL of the ad, normally pointing to the ad partner's reporting endpoint. It's URL encoded",
       "type": "string"
     },
+    "source": {
+      "description": "The browser location where the click was triggered",
+      "enum": [
+        "newtab",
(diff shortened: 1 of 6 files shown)


BenWu (approved):
Looks good to me.  Can you also add this to the proposal?

sunahsuh (approved):
Looks fine to me – only comment is, will we expand the options for this field in the future? If so, how will we remember to keep the list updated here? Just wondering if being more permissive default would make sense, or if filtering out pings with sources outside this list is very important.

ncloudioj:
> Will we expand the options for this field in the future?  If so, how will we remember to keep the list updated here? Just wondering if being more permissive default would make sense, or if filtering out pings with sources outside this list is very important.

Good point, though I don't think we will expand this list for the foreseeable future. In case that we do so at some point and we forget to update it here, we should be able to capture it by the validation error monitors (we'd like to have this as one of the system health monitors).

sunahsuh:
Okay – I think @fbertsch can help point you to doctype-specific validation error monitors

PR #670, opened 2021-04-09
 Bug 1703615 - Add schema for rally study 01

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1703615)

Source schema: https://github.com/mozilla-rally/rally-study-01/blob/master/schemas/measurements.1.schema.json

@hamilton do you have an example document for validation that could be used for this study, like the new disinformation studies in https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/main/validation/pioneer-citp-news-disinfo

Checklist for reviewer:


akkomar (approved):
LGTM

schemas/rally-zero-one/measurements/measurements.1.schema.json
```diff
@@ -0,0 +1,104 @@
...
+        "title": {
+          "description": "the contents of the title element in the head of the page",
+          "type": "string"
+        }
+      },
+      "required": [
+        "eventType",
+        "pageId",
+        "url",
+        "referrer",
```
hamilton, line 89: hmm, I think this is wrong. This should be `referrerOrigin` and `url` should be `origin`.

hamilton:
@amiyaguchi I found an error in this schema that must not have been caught in the review on our side. I'm going to double-check the entire thing again to be 100% sure this is right.

hamilton:
@amiyaguchi – I've reviewed the rest of the schema. Those two required fields noted in the comments are the only things that need to change in this schema. Thanks for your patience and sorry for the last-minute issue!

hamilton:
Thanks @jklukas! Yeesh

schemas/rally-zero-one/measurements/measurements.1.schema.json
acmiyaguchi: I believe this would be something that would have been caught with an example document :) One would be much appreciated, otherwise I'll throw something together before I merge this in.

hamilton: Yes, I didn't realize this was a requirement but agreed that this would be helpful. I'll generate two different ones to test against. Thank you for the suggestion.

acmiyaguchi:
@hamilton Thank you very much.

PR #672, opened 2021-04-12
Bug 1703615 - Add correct schema for rally study 01

acmiyaguchi:
[Bug 1703615](https://bugzilla.mozilla.org/show_bug.cgi?id=1703615)

Followup to #670 and #671. This unnests the schema so that it properly matches the payloads that the addon is creating. @hamilton please confirm that this schema looks correct.

Checklist for reviewer:

For glean changes:


hamilton:
This looks correct to me – thanks @acmiyaguchi.

validation/rally-zero-one/measurements.1.attention-example.pass.json
```diff
@@ -0,0 +1,18 @@
...
+  "referrerOrigin": "",
+  "pageVisitStartTime": 1618243280228,
+  "pageVisitStopTime": 1618243283936,
+  "duration": 3460,
+  "maxRelativeScrollDepth": 0.5968342644320298,
+  "maxPixelScrollDepth": 641,
+  "scrollHeight": 1074,
+  "eventTerminationReason": "page-visit-stop",
+  "title": "Wikipedia",
+  "ogType": "",
```
fbertsch, line 13: A few of these are empty but present. Are we sure they will always be present? Do we want to un-require them?

templates/rally-zero-one/measurements/measurements.1.schema.json
```diff
@@ -0,0 +1,94 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$comment": "Please do not edit this schema by hand. Generate it instead by running npm run build:schema.",
```
fbertsch, line 3: Is this generated from rally?

fbertsch (approved):
r+ with comments

templates/rally-zero-one/measurements/measurements.1.schema.json
hamilton: yes, this was from our end.

fbertsch: Maybe it would be good to document more clearly here where this came from. Just a link to the repo should be fine (unless it can be installed via npm?)

validation/rally-zero-one/measurements.1.attention-example.pass.json
acmiyaguchi: Not sure, @hamilton what are your thoughts here?

hamilton: They will always be present. The data collection library we use sometimes reports empty values.

templates/rally-zero-one/measurements/measurements.1.schema.json
acmiyaguchi: Added to the title

PR #688, opened 2021-07-06
Bug 1719353 Add x_foxsec_ip_reputation metadata field

jklukas:
Per the [Contextual Services Anti-Fraud System plan](https://docs.google.com/document/d/1Db3zxNNeEmJGUtuOYBWwk-DJvBknQDxA0AlfBcYp3RU/edit)

It's unfortunate that this will add a metadata field to all schemas, even though
we only expect this to be set for contextual-services documents.
We could consider folding this value into `x_source_tags` instead within the
Dataflow job to avoid making this schema change at all. We would lose the data
for `payload_bytes_raw` output, though.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/metadata/decoded/decoded.1.schema.json
+++ b/schemas/metadata/decoded/decoded.1.schema.json
@@ -61,6 +61,10 @@
               "description": "X-Debug-Id HTTP header",
               "type": "string"
             },
+            "x_foxsec_ip_reputation": {
+              "description": "X-Foxsec-IP-Reputation header",
+              "type": "string"
+            },
             "x_pingsender_version": {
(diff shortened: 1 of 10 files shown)


whd (approved):
We do technically support other pipeline metadata formats via `bq_metadata_format` but `x_foxsec_ip_reputation` is not necessarily specific to ctxsvc so I think it's reasonable to apply everywhere.

(`bq_metadata_format` currently supports `telemetry`, `structured`, and `pioneer`, which is 1:1 with pipeline families right now, and using multiple metadata formats within a single family is probably going to have some implications somewhere that aren't immediately obvious.)

PR #691, opened 2021-07-19
Bug 1721391 - Creating Beyond the Paywall Rally study.

magorlick:
Checklist for reviewer:

For glean changes:


templates/rally-beyond-the-paywall/deletion-request/deletion-request.1.schema.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$comment": "Bug 1646151 - A deletion request in the Pioneer v2 pipeline.",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "mozPipelineMetadata": {
```
acmiyaguchi, line 4: This metadata information is unnecessary and will be compiled in later.

validation/rally-beyond-the-paywall/article-content-example.pass.json
```diff
@@ -0,0 +1,8 @@
+{
```
acmiyaguchi, line 1: I'd encourage following the convention for validation names, this will affect whether the tests get run.

```
<ping type>.<version>.<test name>.pass.json
```

https://github.com/mozilla-services/mozilla-pipeline-schemas#running-tests-via-docker-optional

templates/rally-beyond-the-paywall/measurements/advertisement.1.schema.json
```diff
@@ -0,0 +1,89 @@
...
+    },
+    "visitId": {
+      "type": "string",
+      "description": "a unique ID associated with a page visit."
+    },
+    "body": {
+      "type": "object",
+      "description": "This object contains two properties listed below about the <body> tag of the page.  This is needed to normalize the size of advertisments against full page size.",
+      "properties": {
+        "body.clientHeight": {
```
acmiyaguchi, line 34: Based on the validation document, it looks like this should be `clientHeight`, otherwise you'd access this field like `document["body"]["body.clientHeight"]`.

```suggestion
        "clientHeight": {
```

templates/rally-beyond-the-paywall/measurements/advertisement.1.schema.json
```diff
@@ -0,0 +1,89 @@
...
+          "type": "integer",
+          "description": "The calculated height in pixels of the <body> tag"
+        },
+        "body.clientWidth": {
```
acmiyaguchi, line 38: Ditto

```suggestion
        "clientWidth": {
```

templates/rally-beyond-the-paywall/measurements/advertisement.1.schema.json
```diff
@@ -0,0 +1,89 @@
...
+          "type": "integer",
+          "description": "The calculated width in pixels of the <body> tag"
+        }
+      }
+    },
+    "ads": {
+      "type": "array",
+      "description": "This array contains an object per advertisement found",
+      "items": {
+        "ads.id": {
```
acmiyaguchi, line 48: Each of these properties includes an `ads.`-prefix that should be removed.

```suggestion
        "id": {
```

templates/rally-beyond-the-paywall/measurements/advertisement.1.schema.json
```diff
@@ -0,0 +1,89 @@
...
+        "body.clientWidth": {
+          "type": "integer",
+          "description": "The calculated width in pixels of the <body> tag"
+        }
+      }
+    },
+    "ads": {
+      "type": "array",
+      "description": "This array contains an object per advertisement found",
+      "items": {
```
acmiyaguchi, line 47: This is malformed, you want an array of objects containing these properties.

```
"items": {
  "type": "object",
  "properties": {
     ...
   }
}

Comment (acmiyaguchi) [reply templates/rally-beyond-the-paywall/measurements/advertisement.1.schema.json:47]: You might also consider adding `required` to the fields the fields are actually all required.

File: templates/rally-beyond-the-paywall/measurements/advertisement.1.schema.json [hunk 7]
```diff
@@ -0,0 +1,89 @@
...
+        "WebScience.articleContents",
+        "WebScience.pageNav",
+        "WebScience.pageNavSensitive",
+        "WebScience.totalTiming"
+      ]
+    },
+    "userId": {
+      "type": "string",
+      "description": "a unique ID associated with a participant."
+    },
```
acmiyaguchi, line 25: Is this ping being sent via the glean.js or legacy pipeline. If the legacy pipeline, how does this differ from the rally/pioneer id?

templates/rally-beyond-the-paywall/pioneer-enrollment/pioneer-enrollment.1.schema.json
```diff
@@ -0,0 +1,12 @@
+{
```
acmiyaguchi, line 1: Please configure this ping like this: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/main/templates/pioneer-citp-news-disinfo-two/pioneer-enrollment/pioneer-enrollment.1.schema.json

validation/rally-beyond-the-paywall/1.i.advertisements.pass.json
```diff
@@ -0,0 +1,34 @@
...
+          "clientHeight": 90,
+          "clientWidth": 1265
+        },
+        {
+          "id": "google_ads_iframe_/29390238/nyt/us/politics_0",
+          "tag": "IFRAME",
+          "clientHeight": 250,
+          "clientWidth": 970,
+          "height": "250",
+          "width": "970"
```
acmiyaguchi, line 23: What prompted the collection for this string height/width field?

validation/rally-beyond-the-paywall/1.i.article-contents.pass.json
```diff
@@ -0,0 +1,10 @@
...
+    }
+  }
```
acmiyaguchi, line 10: I believe this document should fail validation.

acmiyaguchi (changes requested):
Could you also run the compilation process to go from templates/ to schemas/ and check in the results? (See https://github.com/mozilla-services/mozilla-pipeline-schemas#cmake-build-instructions). 

There are some inconsistencies with the test names, if it's possible I'd suggest running the tests locally and making sure that the unit tests run as expected (although the tests should also run here in CI). A few minor issues in the json-schema that should be straightforward to fix. Also, please remove the .DS_Store files from the PR.

acmiyaguchi (commented):
It also looks like you're missing a deletion-request schema here, it can be added like so: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/main/templates/pioneer-citp-news-disinfo-two/deletion-request/deletion-request.1.schema.json

templates/rally-beyond-the-paywall/measurements/advertisement.1.schema.json
magorlick: This should be the legacy pipeline.  This identifier is being used to join the Qualtrics survey data to study data and has been approved by data stewards.

validation/rally-beyond-the-paywall/1.i.advertisements.pass.json
magorlick: The value is a strings sometimes.  Example:
"height": "350px",
"width": "300px"

https://code.stanford.edu/gjmartin/beyond_the_paywall/-/blob/master/schemas/advertisement-example.json

whd:
I want to add here that using the `rally-` prefix for a study that uses the `pioneer`/firefox telemetry transport has caused a lot of confusion when we did it for `zero-one`. While most of that confusion has already occurred, it would simplify matters if firefox telemetry-based studies continue to use the `pioneer-` namespace prefix. See https://bugzilla.mozilla.org/show_bug.cgi?id=1693305 for an example of a "Rally" (from a product perspective) study that does this. For the ingestion pipeline, `rally-` means glean-based, and `pioneer-` means firefox telemetry-based.

Otherwise we can write more special cases akin to what was done for `rally-zero-one`, but it will mean that deployment logic will also need to change to support this study. Please don't merge this as-is until we've decided to either switch to `pioneer-` or added the special case deployment logic.

whd:
I realize now that the convention for external studies is to embed the organization name into the namespace, e.g. `pioneer-citp-news-disinfo`. In this case it should probably be `pioneer-stanford-beyond-the-paywall` or some organization more specific.

We've only done this once so if we want to drop the organization from the namespace we can probably do that, but it will mean that additional information is required in the request bug to name the study project associated with the namespace. This information could perhaps live in metadata eventually.

magorlick:
Thanks for your feedback @whd .  I am not sure that including the partner name is strong pattern yet as we did not do this for the rally study either (rally-rally-zero-one would comply).  We also need to consider the possibility of partnering with multiple individuals within the same institution (CITP or Stanford).  What do we use this information for in this context? I want to find a good long-run solution that serves the need.  Study names will always be unique and that is what we are using here with the pioneer- prefix.

whd:
>  I am not sure that including the partner name is strong pattern yet as we did not do this for the rally study either (rally-rally-zero-one would comply). 

We've hitherto made a distinction between studies internal to Mozilla and those that involve a separate organization. I believe `rally-zero-one` and `rally-study-zero-one` are internal studies but perhaps they were meant to be shared with an external organization eventually.

> We also need to consider the possibility of partnering with multiple individuals within the same institution (CITP > or Stanford). What do we use this information for in this context? 

We use it to apply a naming convention to the GCP Project ID associated with a study. This is arbitrary to a degree and we could change it. The current pattern is motivated by restrictions on the Project ID length (it can't be 1:1 with the namespace since it is limited to 30 characters and there is a substantial prefix consumed by ops naming convention). I mention in https://bugzilla.mozilla.org/show_bug.cgi?id=1721391#c2 what the project id might look like for this study.

I'm ok with making the project id convention something like `moz-fx-data-pioneer-f757` instead and removing any organization or namespace information from the project id. Sufficient information should be available in the project name at any rate.

acmiyaguchi (changes requested):
Please remove the .DS_Store value from the commit too.

Outside of this and the naming convention, the schemas look good to me.

acmiyaguchi (approved):
The final name for the study is `pioneer-sgsb-beyond-the-paywall`, it sounds like everyone in this thread is okay for this to be the final study name. 

The schemas look good, so I'll merge this in.

PR #698, opened 2021-09-08
Bump java version to 11 (fixes #696)

### LINKED ISSUE #696: CI `test` step currently failing on all PRs
Example failure: https://app.circleci.com/pipelines/github/mozilla-services/mozilla-pipeline-schemas/2547/workflows/d47fc074-80bc-4c9c-bb08-56e1a7fb6e79/jobs/5054

```
#!/bin/sh -eo pipefail
docker run mps

============================= test session starts ==============================
platform linux -- Python 3.6.8, pytest-6.0.2, py-1.9.0, pluggy-0.13.1 -- /usr/bin/python3.6
cachedir: .pytest_cache
rootdir: /app
plugins: forked-1.3.0, xdist-2.1.0
collecting ... 

Too long with no output (exceeded 10m0s): context deadline exceeded
```

This appears to be happening on all CI runs right now, which will prevent us confidently merging PRs.

wlach:


Dockerfile
```diff
@@ -15,11 +15,15 @@ RUN dnf -y update && \
...
         python36 \
-        java-1.8.0-openjdk-devel \
+        java-11-openjdk-devel \
         maven \
         cargo \
     && dnf clean all
 
+# various hacks to make sure we use openjdk 11
+ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk-11.0.12.0.7-0.el8_4.x86_64
+RUN alternatives --set java /usr/lib/jvm/java-11-openjdk-11.0.12.0.7-0.el8_4.x86_64/bin/java
```
wlach, line 25: I am a bit concerned this might break if the upstream repositories change, but I don't know of a better way of doing this.

jklukas: Is it possible to specify a more specific package than `java-11-openjdk-devel` to ensure we get stable paths here?

jklukas (approved):
r+wc

Dockerfile
wlach: I'm pretty sure `java-11-openjdk-devel` only has one package available at a time. However, I think I've found a way to use less specific paths. Testing...

Dockerfile
```diff
@@ -15,11 +15,15 @@ RUN dnf -y update && \
...
         python36 \
-        java-1.8.0-openjdk-devel \
+        java-11-openjdk-devel \
         maven \
         cargo \
     && dnf clean all
 
+# ensure we're actually using java 11
+ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk
+RUN alternatives --set java /usr/lib/jvm/java-11-openjdk/bin/java
```
relud, line 25: `/usr/lib/jvm/java-11-openjdk` is already a soft link to alternatives, is this command necessary?

relud (approved):
r+wc

Dockerfile
```diff
@@ -15,11 +15,15 @@ RUN dnf -y update && \
...
+
```
relud, line 26: setting JAVA_HOME and alternatives is unnecessary here, there's only one jdk installed:
```suggestion
```

relud, line 26: setting JAVA_HOME and alternatives is unnecessary here, there's only one jdk installed:
```suggestion
```

wlach: java-1.8.0-openjdk-devel seems to be uninstalled by default, which is why I put this in. Let me try removing it.

relud: it's not installed by default, it's being pulled in by maven

wlach: I see. In any case, yes, it seems like this command *is* necessary (although the invocation of it here was slightly incorrect) since java 1.8.0 gets selected as default otherwise.

relud (approved):
🎉

jklukas:
🙌

PR #699, opened 2021-09-08
Use fastestmirror for dnf packages

relud:
and use `--pull` when building docker image
and use https remote url to avoid ssh host key verification bug (running test locally via `circleci build --job test` stalls on asking for host key verification when using the ssh url)


wlach (approved):
Thank you! I was wondering if there was anything that could be done about my timing out test runs...

PR #705, opened 2021-10-18
Bug 1736117 - Add request_id to QuickSuggest tables

ncloudioj:
Checklist for reviewer:

For glean changes:

Note that although Merino currently uses the UUID/v4 for `request_id`, I didn't specify it in the `pattern` field as it's subject to future changes.

r? @jklukas

### DIFF
--- a/schemas/contextual-services/quicksuggest-impression/quicksuggest-impression.1.schema.json
+++ b/schemas/contextual-services/quicksuggest-impression/quicksuggest-impression.1.schema.json
@@ -82,8 +82,6 @@
   },
   "required": [
     "context_id",
-    "search_query",
-    "matched_keywords",
     "block_id",
     "is_clicked",
     "version"
--- a/templates/contextual-services/quicksuggest-impression/quicksuggest-impression.1.schema.json
+++ b/templates/contextual-services/quicksuggest-impression/quicksuggest-impression.1.schema.json
@@ -66,8 +66,6 @@
   },
   "required": [
     "context_id",
-    "search_query",
-    "matched_keywords",
     "block_id",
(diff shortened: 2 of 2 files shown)


jklukas:
I was just thinking earlier today that this change to Merino flow may be a good opportunity to recreate the schema. In particular, I would propose creating a new schema `suggest-impression v2` which would allow us to do the following:

- Add the `request_id`
- Drop the outdated "quicksuggest" branding from the ping name
- Remove the `search_query` and `matched_keywords` fields, removing uncertainty about whether the ping contains user query data

This would also make it easier for us to monitor the rate of submissions to the online branch vs. the merino branch.

ncloudioj:
Good point. We can certainly do that.

Let's see what else do we need for this migration given that the existing offline&online rollouts will be sticking around for some time.
* Pings from those offline & online rollouts will continue to be sent to the old tables
* Merino-backed impressions will go to the new impression table, shall we do this for clicks as well now, or later?
* Update the derived table (`event_aggregates`)
* We might not be able to know the search queries for impressions unless Merino is allowed to log them and send them to the pipeline, just wanted to point that out

cc: @rebecca-burwei

jklukas:
> for Merino-backed impressions, we will set the search_query and matched_keywords to an empty string (unfortunately, we can leave them out as they both are required fields)

Making these fields optional would be a compatible change to the schema. We can remove them from the `required` list in this PR, and once it flows through the schema deployment process the pipeline will stop rejecting messages that are missing those fields.

ncloudioj:
Oh, I didn't know that was an allowed change to make. I will update my patch for that.

PR #709, opened 2021-11-24
Bug 1742172 Add override_attributes to metaschema

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1742172

The associated pipeline change is in https://github.com/mozilla/gcp-ingestion/pull/1898

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/metadata/metaschema/metaschema.1.schema.json
+++ b/schemas/metadata/metaschema/metaschema.1.schema.json
@@ -47,6 +47,7 @@
         "jwe_mappings": {
           "description": "Mappings of encrypted JWE field paths to destinations where the value decrypted by the pipeline should be placed; initial use case is Account Ecosystem Telemetry; paths must be in [JSON Pointer format](https://tools.ietf.org/html/rfc6901) like '/payload/ecosystemAnonId'",
           "items": {
+            "additionalProperties": false,
             "properties": {
               "decrypted_field_path": {
                 "pattern": "^/.*$",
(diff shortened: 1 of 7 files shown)


schemas/metadata/metaschema/metaschema.1.schema.json
```diff
@@ -47,6 +47,7 @@
         "jwe_mappings": {
           "description": "Mappings of encrypted JWE field paths to destinations where the value decrypted by the pipeline should be placed; initial use case is Account Ecosystem Telemetry; paths must be in [JSON Pointer format](https://tools.ietf.org/html/rfc6901) like '/payload/ecosystemAnonId'",
           "items": {
+            "additionalProperties": false,
```
whd, line 50: :+1:

schemas/metadata/metaschema/metaschema.1.schema.json
```diff
@@ -57,12 +58,45 @@
                 "type": "string"
               }
             },
+            "required": [
```
whd, line 61: This is unrelated cleanup?

schemas/metadata/metaschema/metaschema.1.schema.json
```diff
@@ -57,12 +58,45 @@
...
+              "source_field_path",
+              "decrypted_field_path"
+            ],
+            "type": "object"
+          },
+          "type": "array"
+        },
+        "override_attributes": {
+          "description": "Mappings of Pub/Sub attribute names to static values; these are applied in the Decoder immediately before incorporating metadata into the payload, so can be used to overwrite values calculated in the pipeline; a null value will cause the pipeline to drop the named attribute; some attribute names differ from the nested metadata format in BigQuery, so for example you must use \"geo_city\" here in order to manipulate the value that shows up as metadata.geo.city; implemented for bug 1742172",
```
whd, line 70: 5 semis dang

whd (approved):
LGTM

whd:
Also, I'm guessing this is an array of named key-value pairs instead of an object because (a) it maps more closely to the actual attributes data structure, or (b) is extensible if we need to add some new options.

relud (approved):

schemas/metadata/metaschema/metaschema.1.schema.json
jklukas: Yes, I should have commented on this. While implementing `override_attributes`, I intentionally tried implementing some instances of the field incorrectly and found that the test suite was not picking up the errors, so I realized I need to disallow additionalProperties and specify required fields. I then went ahead and made those same improvements for the jwe_mappings schema.

jklukas:
> Also, I'm guessing this is an array of named key-value pairs instead of an object because (a) it maps more closely to the actual attributes data structure, or (b) is extensible if we need to add some new options.

Mostly (b). I could imagine needing to implement some additional flags here to deal with non-string types or support some basic conditional logic.

PR #730, opened 2022-03-29
Bug 1761790 Change type of Pocket tile_id from int to string 

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1761790

Currently a draft to demonstrate what this solution would look like. Not committed to this path yet.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/templates/activity-stream/impression-stats/impression-stats.2.schema.json
+++ b/templates/activity-stream/impression-stats/impression-stats.2.schema.json
@@ -1,0 +1,78 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "impression-stats",
+  "properties": {
+    @ACTIVITY-STREAM_IMPRESSIONID_1_JSON@,
+    @ACTIVITY-STREAM_PAGE_1_JSON@,
+    @ACTIVITY-STREAM_EXPERIMENTS_1_JSON@,
+    "shield_id": {
+      "description": "[DEPRECATED]: use `experiments` instead. A semicolon separated string to store a list of Shield study IDs",
+      "type": "string"
+    },
+    "addon_version": {
+      "type": "string"
+    },
+    "version": {
+      "type": "string"
+    },
+    "release_channel": {
+      "type": "string"
+    },
+    "locale": {
+      "type": "string"
+    },
+    "source": {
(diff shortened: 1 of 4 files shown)


jklukas:
cc @ncloudioj @kirill-demtchouk

templates/activity-stream/impression-stats/impression-stats.2.schema.json
```diff
@@ -0,0 +1,78 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "impression-stats",
+  "properties": {
+    @ACTIVITY-STREAM_IMPRESSIONID_1_JSON@,
+    @ACTIVITY-STREAM_PAGE_1_JSON@,
+    @ACTIVITY-STREAM_EXPERIMENTS_1_JSON@,
+    "shield_id": {
```
ncloudioj, line 9: I think this field can be dropped now.

ncloudioj:
LGTM, r=nanj.

kirill-demtchouk:
@jklukas Can you remind me: if we wanted to add a new field to an existing schema, would that require incrementing the schema another full version and creating a new destination table in BigQuery? (This is in reference to my earlier comment about potentially using this schema change as an opportunity to start including the main `client_id` with these pings.)

jklukas:
> Can you remind me: if we wanted to add a new field to an existing schema, would that require incrementing the schema another full version and creating a new destination table in BigQuery?

Adding a new field doesn't require a version bump. It's generally only modifying a field's type that requires a version bump. Removing a field would also require a bump, but in practice we usually let deprecated fields stay around (removing them from the required list if they were marked as required) rather than bumping the version to fully remove them.

So, the door is open to adding `client_id` in the future.

jklukas:
> using this schema change as an opportunity to start including the main client_id with these pings

cc @ANich who is working on a reimagining of newtab telemetry. If @kirill-demtchouk isn't aware of that workstream or if @ANich hasn't had a chance to hear Pocket perspectives on newtab telemetry needs, then we should make sure you get a chance to talk.

ncloudioj (approved):
LGTM.

scholtzan (approved):

PR #733, opened 2022-05-03
Bug 1767448 - Add window_inner_height and window_inner_width to impression-stats ping.

kirill-demtchouk:
Scott Downe and I would like to start tracking on the impression-stats ping. I updated both the v1 and v2 schemas to support the new fields. I did not update any sample events in the /validation subfolder since the new fields are not required.

Checklist for reviewer:

For glean changes:


jklukas (commented):
This looks technically correct, but I'll block on leaving an approving review until the bug goes through data review.

It would also be worth adding these new fields in one of the passing pings under the `validation` directory.

PR #735, opened 2022-06-28
Bug 1776797 - Replace scenario with improve_suggest_experience_checke…

ncloudioj:
…d in quicksuggest pings

Checklist for reviewer:

For glean changes:

This fixes [Bug 1776797](https://bugzilla.mozilla.org/show_bug.cgi?id=1776797).

r? @quiiver @jklukas

### DIFF
--- a/schemas/contextual-services/quicksuggest-block/quicksuggest-block.1.schema.json
+++ b/schemas/contextual-services/quicksuggest-block/quicksuggest-block.1.schema.json
@@ -49,6 +49,10 @@
       ],
       "type": "string"
     },
+    "improve_suggest_experience_checked": {
+      "description": "A boolean indicating whether the user has opted in to improving the Firefox Suggest experience",
+      "type": "boolean"
+    },
     "locale": {
       "description": "User's current locale",
       "type": "string"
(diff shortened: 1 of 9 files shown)


quiiver (approved):
i'm not totally sure what the deprecation process is. my assumption is that the absence of fields in the schema is still considered valid but unexpected fields would cause schema invalidation. assuming that is true: lgtm 👍

jklukas (approved):

ncloudioj:
> i'm not totally sure what the deprecation process is. my assumption is that the absence of fields in the schema is still considered valid but unexpected fields would cause schema invalidation.

I think we will keep them both. The old Firefox versions will continue sending `scenario` without the new field. The newer version (103+) will do the opposite. None of them are required, so we should be fine.

jklukas:
> my assumption is that the absence of fields in the schema is still considered valid but unexpected fields would cause schema invalidation

Unless a field is marked in the `required` list of its enclosing `object` in the JSON schema, it is considered optional and a document will not fail validation due to the missing field.

Unexpected fields are generally allowed and will show up in `additional_properties` in the BigQuery tables. In most cases, this means that there's no required order for getting a new field added to the schema here and getting it added in the client.

jklukas:
The `post-artifacts` CI task is failing because it can't provide credentials to PRs from forks.

@quiiver - To get this passing, see the push-to-trigger-ci instructions in https://github.com/mozilla-services/mozilla-pipeline-schemas#packaging-and-integration-tests-optional

quiiver:
triggered build here https://app.circleci.com/pipelines/github/mozilla-services/mozilla-pipeline-schemas?branch=trigger-integration

PR #743, opened 2022-08-23
Update cargo version requirement comment

relud:
incorrect value stems from a typo in https://github.com/matthiaskrgr/cargo-cache/issues/107#issuecomment-945009034, debian bookworm actually installs cargo 1.56

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 FROM --platform=linux/amd64 python:3.8-slim-bullseye
 LABEL maintainer="Mozilla Data Platform"
 
-# copied from library/rust:1.63.0-slim-bullseye to install rust >= 1.46 and cargo >= 0.57 needed to build jsonschema-transpiler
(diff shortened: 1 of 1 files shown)


fbertsch (commented):
Are you sure? https://packages.debian.org/bookworm/cargo shows 0.57.0-7 on bookworm.

relud:
> Are you sure? https://packages.debian.org/bookworm/cargo shows 0.57.0-7 on bookworm.

TL;DR i'm wrong, closing invalid

`cargo --version` and [cargo's changelog](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md) are organized by rust version, and their git repo has branches named `rust-1.56.0` that matches that numbering.

cargo's actual version matches the tags like `0.57.0` on their git repo, which is the version used to name debian packages, and the version displayed in their docs at https://docs.rs/cargo/latest/cargo/.

so when you install cargo 0.57.0 and run `cargo --version` it prints out `cargo 1.56.0`, which I really hate, and is why i opened this PR.

PR #777, opened 2023-06-22
Add geoip skip to metaschema

whd:
/CC @Dexterp37 

There's got to be a better natural language description or name for this.

Another option might be to make this e.g. `geoip_index`, with negative values being from the end (the default would be `-3`). The expected value for trusted proxies configuration would be `-4` or `-5` (I can't remember but it's all written down in the Jira ticket).

I'm trying to be explicit about the fact that it's very possible to receive submissions that don't have the expected number of entries when dealing with namespaces that have trusted proxy metadata configured and that there is fallback behavior as a result.

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/metadata/metaschema/metaschema.1.schema.json
+++ b/schemas/metadata/metaschema/metaschema.1.schema.json
@@ -44,6 +44,10 @@
           },
           "type": "object"
         },
+        "geoip_skip_entries": {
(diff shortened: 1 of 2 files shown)


relud (approved):

Dexterp37:
@whd thanks for this, happy to see it coming along. If @relud is fine with this change, I'm good too!

PR #779, opened 2023-06-30
Bug 1840381 - add  `TargetFrameRate` to `environment.system.gfx`

lelilia:
New environment parameter `TargetFrameRate` in `environment.system.gfx`

https://bugzilla.mozilla.org/show_bug.cgi?id=1840381
https://bugzilla.mozilla.org/show_bug.cgi?id=1840398
https://bugzilla.mozilla.org/show_bug.cgi?id=1840395
https://bugzilla.mozilla.org/show_bug.cgi?id=1840378

Also showed up in the new `main_remainder` table where the schema seems to come from some different place, so not sure how to address that.

### DIFF
--- a/schemas/telemetry/bhr/bhr.4.schema.json
+++ b/schemas/telemetry/bhr/bhr.4.schema.json
@@ -867,6 +867,10 @@
                 "LowEndMachine": {
                   "type": "boolean"
                 },
+                "TargetFrameRate": {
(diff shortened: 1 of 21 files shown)


relud:
> Also showed up in the new `main_remainder` table where the schema seems to come from some different place, so not sure how to address that.

it will be automatically added if it appears in `main`

relud (approved):

PR #786, opened 2023-09-27
Bug 1839487 Add new suggested_index columns to quicksuggest

0c0w3:
https://bugzilla.mozilla.org/show_bug.cgi?id=1855338

Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/contextual-services/quicksuggest-block/quicksuggest-block.1.schema.json
+++ b/schemas/contextual-services/quicksuggest-block/quicksuggest-block.1.schema.json
@@ -90,6 +90,14 @@
       "description": "The source of the suggestion, either “remote-settings” or “merino”. See bug 1839487",
       "type": "string"
     },
+    "suggested_index": {
+      "description": "The intended index of the QuickSuggest suggestion. If `suggested_index_relative_to_group` is true, the index is relative to the Firefox Suggest group; otherwise the index is relative to the entire list of suggestions. Non-negative values (starting at 0) are relative to the start/top of the group/list; negative values are relative to the end/bottom of the group/list.",
+      "type": "integer"
+    },
+    "suggested_index_relative_to_group": {
+      "description": "Whether `suggested_index` is relative to the Firefox Suggest group. If false, it is relative to the entire list of suggestions.",
+      "type": "boolean"
+    },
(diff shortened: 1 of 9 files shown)


0c0w3:
cc @ncloudioj @quiiver, thanks

ncloudioj (approved):
LGTM!

schemas/contextual-services/quicksuggest-block/quicksuggest-block.1.schema.json
```diff
@@ -90,6 +90,14 @@
       "description": "The source of the suggestion, either “remote-settings” or “merino”. See bug 1839487",
       "type": "string"
     },
+    "suggested_index": {
```
quiiver, line 93: What is the reason for having this be non-relative when we already have the global `position` field?

0c0w3: It's helpful to consider a couple different uses of "relative":

1. `suggested_index` can be relative to the Firefox Suggest group (`suggested_index_relative_to_group = true`) or to the whole results list (`suggested_index_relative_to_group = false`), although the latter might also be referred to as "global" or "non-relative"
2. `suggested_index` can be relative the beginning of the group/list or the end. Positive values are relative to the beginning, negative to the end

In practice `suggested_index` is helpful when:

* It's relative to the Firefox Suggest group, or
* It's relative to the end of the results list

In contrast, if it's relative to the start of the results list, then it's practically the same as `position`, modulo the different bases (1-based vs. 0-based). The only time it would be really different is when there are fewer results in the list than the suggested index, e.g. there are only 3 results but the suggested index is 9. But in practice we never use large suggested indexes relative to the whole list. We only ever use a value of 1, for top picks.

All that said, the reason we added these was to support the proposed second treatment branch of higher-placement sponsored suggestions where they're shown at the top of the Suggest section. We would be able to distinguish that case by checking `suggested_index = 0` and `suggested_index_relative_to_group = true`. In general we'd be able to distinguish any other placements we'd want to experiment with. Now that we aren't doing that second branch, at least not right now, we don't actually need these values and so we could just remove them.

PR #788, opened 2023-10-30
DENG-476 - Stop loading main-v4 et al

relud:


whd (approved):
Planning to roll this to stage tomorrow and maybe prod as well

PR #808, opened 2024-03-12
[DENG-3022] Support minimum glean schema

quiiver:
Schema switching in MSG here: https://github.com/mozilla/mozilla-schema-generator/pull/261

Checklist for reviewer:

For glean changes:


templates/include/glean/glean_base.1.schema.json
```diff
@@ -0,0 +1,174 @@
+{
+  "$schema": {
+    "type": "string",
+    "enum": [
+      "moz://mozilla.org/schemas/glean/ping/1"
+    ]
+  },
+  "client_info": @GLEAN_CLIENT_INFO_1_JSON@,
+  "ping_info": {
```
sean-rose, line 9: I'd have expected the `*_info` sections to only be present in the normal schema.  Is there a reason the `*_info` sections are still being included in the min schema?

quiiver: My thinking was that info sections could be included in the min schema but were not a requirement. Happy to change that though. Minimum schema here just relaxed the requirements while keeping the schema the same.

sean-rose: Given the current planned implementation of not sending any `*_info` data for such pings, meaning all of those fields would always be null, I think it would help avoid confusion if the fields were omitted from the min schema (and thus the BQ tables/views).

However there may well be other considerations/trade-offs, like if it'd be a prohibitive amount of work to update all the places existing code assumes certain `*_info` fields exist for all pings (e.g. `client_info.client_id`), or perhaps planning ahead for when we might send some of the `*_info` data in the future and cherry picking individual fields for the schema would be too complicated.

schemas/glean/glean/glean-min.1.schema.json
```diff
@@ -0,0 +1,630 @@
...
+      },
+      "required": [
+        "app_build",
+        "app_display_version",
+        "architecture",
+        "first_run_date",
+        "os",
+        "os_version",
+        "telemetry_sdk_build"
+      ],
```
badboy, line 99: Should these all still be required in the minimal scheme?
I _think_ right now this schema would enforce that either it has no `client_info` or if it exists it has the required fields at least. Is that what we want? Or would we allow a smaller `client_info` too?

templates/include/glean/glean_base.1.schema.json
sean-rose: > if it'd be a prohibitive amount of work to update all the places existing code assumes certain `*_info` fields exist for all pings (e.g. `client_info.client_id`)

@scholtzan I think you'd know best about this.  What do you think?

quiiver: I do agree that a simpler bq schema would be good and way less confusing. I could really go either way on this, however given that the option for turning this off from the SDK is a binary on/off it probably makes the most sense to just remove the info sections from the minimum glean schema.

scholtzan: Yeah, if we are not planning to collect that data and it's not possible from the SDK to even send it for these minimal pings, then it's best to leave it away

quiiver: Updated.

schemas/glean/glean/glean-min.1.schema.json
quiiver: Given the conversation above, I've removed these from the `glean-min` schema. If we want to progressively add info properties we will be able to do that in an adhoc manner directly to the minimum schema.

scholtzan (approved):

whd:
One potential side effect of this change is that wildcard queries that previously worked on stable tables may no longer for datasets with mixed metadata schemas. These appears to have affected GLAM /CC @edugfilho. One way to mitigate this might be to require that minimum glean pings be sequestered in their own namespace(s), but I have a feeling that's a "simple on the pipeline side complicated on the client side" kind of solution.

PR #812, opened 2024-05-03
Ensure `version` property is required when present in legacy telemetry

sean-rose:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/anonymous/anonymous.4.schema.json
+++ b/schemas/telemetry/anonymous/anonymous.4.schema.json
@@ -82,7 +82,8 @@
     }
   },
   "required": [
-    "payload"
+    "payload",
+    "version"
   ],
   "title": "anonymous",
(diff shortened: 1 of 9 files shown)


BenWu (commented):
I don't think this is necessary because the decoder drops these pings before they get to schema validation but it also doesn't hurt to be more explicit in the schema.  I don't have a problem with this especially if it's needed for the other PR but it'd be good to have @akkomar to review

sean-rose:
> especially if it's needed for the other PR

It isn't needed for #811 as I've already made adjustments there to more gracefully handle `version` properties that might be configured to allow nulls (though that should never be the case in the actual data due to how ingestion works as you've pointed out).

So this is more about making the affected schemas more reflective of reality, consistent with the other legacy schemas, and enforcing that consistency for legacy schemas going forward.

PR #822, opened 2024-07-18
Bug 1901263: Add `profileGroupId` to schemas.

Mossop:
This adds the `profileGroupId` being added to the client side in bug 1901263 to the schemas wherever the `clientId` is present. I believe that every ping including `clientId` will include `profileGroupId` however since the new property is not marked as required in any schema (as it cannot be to avoid breaking older versions of Firefox) if new pings are introduced that do not include the property then it should not break anything.

Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title)
- [ ] If adding a new field, the field should have a description (see #576 for an example)
- [ ] If coming from a fork, run integration tests: `./.github/push-to-trigger-integration <username>:<branchname>`

For glean changes:
- [ ] Update `templates/include/glean/CHANGELOG.md`

For modifications to schemas in restricted namespaces (see [`CODEOWNERS`](/CODEOWNERS)):
- [ ] Follow the [change control procedure](https://docs.google.com/document/d/1TTJi4ht7NuzX6BPG_KTr6omaZg70cEpxe9xlpfnHj9k/edit#heading=h.ttegrcfy18ck)

PR #823, opened 2024-07-24
Pin setuptools version to 70.3.0

akkomar:
Recent releases of setuptools cause the build to fail. This fixes it and will unblock https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/822

Checklist for reviewer:

For glean changes:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,7 +69,7 @@ COPY --from=mozilla/ingestion-sink:latest /app/ingestion-sink/target /app/target
 
 # Install python dependencies
 COPY requirements.txt requirements-dev.txt ./
-RUN pip3 install --upgrade pip setuptools && \
+RUN pip3 install --upgrade pip setuptools==70.3.0 && \
     pip3 install -r requirements.txt -r requirements-dev.txt
 
 # Install Java dependencies


scholtzan (approved):

Dockerfile
```diff
@@ -69,7 +69,7 @@ COPY --from=mozilla/ingestion-sink:latest /app/ingestion-sink/target /app/target
 
 # Install python dependencies
 COPY requirements.txt requirements-dev.txt ./
-RUN pip3 install --upgrade pip setuptools && \
+RUN pip3 install --upgrade pip setuptools==70.3.0 && \
```
sean-rose, line 72: Does the newer setuptools cause problems running things locally?  If so, then I'd think you'd also want to pin setuptools in `requirements.txt`.

akkomar: I'm not sure we should add setuptools to requirements: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/36abca46ba9a311628ef8cde510e9f74becba16a/requirements-dev.txt#L28C3-L29

sean-rose: Under normal circumstances I agree, but if the setuptools version incompatibility was going to consistently cause problems for devs that would take precedence IMO.

Luckily this is a moot point at the moment because it turns out the latest production versions of Python only come with setuptools 65 installed by default (and Python 3.8, which is the version used in the Dockerfile, comes with setuptools 56 by default).

PR #824, opened 2024-07-25
Bug 1907663 - Add scale factor fields to gfx

saschanaz:
Checklist for reviewer:

For glean changes:

### DIFF
--- a/schemas/telemetry/bhr/bhr.4.schema.json
+++ b/schemas/telemetry/bhr/bhr.4.schema.json
@@ -898,6 +898,10 @@
                   "description": "Frame rate in Hz, typically 60 or more, see bug 1840381",
                   "type": "integer"
                 },
+                "textScaleFactor": {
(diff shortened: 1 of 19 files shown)


BenWu (commented):
You can do this is by updating the environment schema template [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/main/templates/include/telemetry/environment.1.schema.json) and then running the [build instructions in the readme](https://github.com/mozilla-services/mozilla-pipeline-schemas?tab=readme-ov-file#cmake-build-instructions), which. should give the same output.  CI is failing because the template doesn't match the output

saschanaz:
Thanks, I should have suspected that there should be a better option than regex 😅 for now I tried manual fix.

BenWu (approved):

BenWu:
The column should show up tomorrow.  We seem to be getting significant volume already (in the additional_properties column)

saschanaz:
Thank you!

PR #829, opened 2024-09-27
Bug 1859597 - added stub_build_id to stub installer ping schema

erchen03:


akkomar (changes requested):
There's a test failure due to wrong ordering in the generated file. Can you rebuild schemas following https://github.com/mozilla-services/mozilla-pipeline-schemas?tab=readme-ov-file#adding-a-new-schema and update this PR?

akkomar (approved):

PR #836, opened 2025-01-10
changing windows_ubr to be type string

erchen03:
Checklist for reviewer:

For glean changes:


BenWu:
We can't trivially change the data type of a field after it's added to the bigquery schema.  We would need to do something like `windows_ubr2` which would require a client-side release or some special case in ingestion.  I'll see if we have some precedent for doing this

PR #875, opened 2026-06-29
[DENG-11329] Migrate CI to github actions

BenWu:
https://mozilla-hub.atlassian.net/browse/DENG-11329

No image to push so it's just tests and posting a diff comment for non-forks


.github/push-to-trigger-integration
BenWu, line 1: this doesn't do anything anymore so I just settled for putting the diff in the run summary for forks

scholtzan (approved):

